PR #22, opened 2019-01-10
Multiple roots and federation

### LINKED ISSUE #19: Add basic federation where you can add multiple roots to your trust bundle
Also a useful feature for root certs rotation.

maraino:
### Description
This PR adds support for federated roots, and multiple roots for root rotation. Fixes #19

Federated roots are a list of public certificates that a client can include and trust, a new endpoint (`/federation`) has been added that returns the server roots as well as the federated ones.

The CA now supports one or multiple roots in the `root` property of the JSON, that property is able to unmarshal both strings and arrays, the latter must be always used for multiple roots. The SDK is able to update the root certificates on each renew, and this feature can be used for root rotation following this process:
1. CA is configured with rootA.
2. BootstrapServer/BootstrapClient are created, accepting only rootA.
3. CA is updated and adds rootB, rootA and rootB are accepted.
4. BootstrapServer/BootstrapClient do a renew and include rootB into the trusted pools
5. CA can now create new intermediates and remove rootA.
6. BootstrapServer/BootstrapClient will able to renew with rootB.

A known issue is that at this moment rootA won't be removed, but a new issue with a strategy to support that has been created.


api/api_test.go
```diff
@@ -443,6 +445,20 @@ func (m *mockAuthority) GetEncryptedKey(kid string) (string, error) {
 	return m.ret1.(string), m.err
 }
 
+func (m *mockAuthority) GetRoots() ([]*x509.Certificate, error) {
+	if m.getFederation != nil {
```
dopey: m.getRoots?

maraino: Fixed 8252608

ca/client.go
```diff
@@ -413,6 +399,42 @@ func (c *Client) ProvisionerKey(kid string) (*api.ProvisionerKeyResponse, error)
...
+func (c *Client) Roots() (*api.RootsResponse, error) {
+	u := c.endpoint.ResolveReference(&url.URL{Path: "/roots"})
+	resp, err := c.client.Get(u.String())
+	if err != nil {
+		return nil, errors.Wrapf(err, "client GET %s failed", u)
+	}
+	if resp.StatusCode >= 400 {
+		return nil, readError(resp.Body)
+	}
+	var federation api.RootsResponse
```
dopey: var roots api.RootsResponse?

maraino: Fixed in dbd1bf1

authority/root.go
```diff
@@ -17,13 +17,39 @@ func (a *Authority) Root(sum string) (*x509.Certificate, error) {
...
 	return crt, nil
 }
 
 // GetRootCertificate returns the server root certificate.
 func (a *Authority) GetRootCertificate() *x509.Certificate {
-	return a.rootX509Crt
+	return a.rootX509Certs[0]
+}
+
+// GetRootCertificates returns the server root certificates.
```
dopey: Here we have two methods in the public Authority API that return the same thing. The explanation behind this duplication is that one of these methods is used by the CA API while the other is used by the SDK API. It is possible that the implementations for the methods change (e.g. new conditions are added to one and not the other) in the future.

authority/types.go
```diff
@@ -0,0 +1,104 @@
...
+// First returns the first element of a multiString. It will return an empty
+// string if the multistring is empty.
+func (s multiString) First() string {
+	if len(s) > 0 {
+		return s[0]
+	}
+	return ""
+}
+
+// Empties returns `true` if any string in the array is empty.
```
dopey: Consider changing this to `HasEmpties`? NitPicky and I'm not entirely sure that the name change has any effect on clarity. It's just that `Has...` is a more commonly used name for this sort of thing.

maraino: Fixed in 6e62007

authority/root.go
maraino: Fixed in e8ac3f4

PR #25, opened 2019-01-18
make Duration wrapper publicly accessible

dopey:
### Description

💔Thank you!


dopey:
`type duration time.Duration` does not allow other packages to use time.Duration in place of `duration`. Going back to making Duration public and adding some unit tests.

PR #26, opened 2019-01-23
Add basic federation example

sourishkrout:
### Description
Example code to illustrate how federation can be leveraged as of v0.8.3


examples/basic-federation/server.go
```diff
@@ -0,0 +1,73 @@
...
+		if err != nil {
+			return err
+		}
+		roots, err := ctx.Client.Roots()
+		if err != nil {
+			return err
+		}
+
+		if len(certs.Certificates) > len(roots.Certificates) {
+			fmt.Println("Server is using federated root certificates")
```
maraino: This is weird, the server is using both, federated and root

sourishkrout: You mean the wording of the log message?

maraino: Yes, but I suppose for the example is ok.
I added the comment because in the requests before the comment (get federation and roots), the certs used are the root ones.

sourishkrout:
@maraino made some minor tweaks to the readme. lemme know if it's good to merge please. thanks!

PR #28, opened 2019-01-28
authority/provisioners: fix overflow on 32-bit systems

pallas:
In Go, len returns signed ints, not unsigned ints; consequently, this code
comparison overflows on 32-bit systems, like ARM.


maraino:
@pallas thanks for noticing this

pallas:
No problem, happy to help

PR #33, opened 2019-02-13
hello-mTLS example using gRPC in go

### LINKED ISSUE #32: Add a hello-mTLS example using gRPC
### Description
Add a hello-mTLS example using gRPC

maraino:
### Description
This PR adds a new hello-mTLS example with a gRPC client and server written in go.

Fixes #32


autocert/examples/hello-mtls/go-grpc/server/server.go
```diff
@@ -0,0 +1,139 @@
...
+
+	return pool, nil
+}
+
+// Greeter is a service that sends greetings.
+type Greeter struct{}
+
+// SayHello sends a greeting
+func (g *Greeter) SayHello(ctx context.Context, in *hello.HelloRequest) (*hello.HelloReply, error) {
+	return &hello.HelloReply{Message: "Hello " + in.Name}, nil
```
mmalone: Looks like `in.Name` is coming from the protobuf request? In the HTTPS example I was intentionally trying to demonstrate pulling the name from a validated certificate because it'd be the next step if you wanted to do authorization... so basically that example gets you to the point where you have authenticated identity and are ready to implement your own access control. Might be worthwhile to do the same thing here. If I understand correctly, gRPC has some abstraction around this...

autocert/examples/hello-mtls/go-grpc/server/server.go
```diff
@@ -0,0 +1,139 @@
...
+			tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
+			tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+		},
+		GetCertificate: r.getCertificate,
+	}
+
+	// Schedule periodic re-load of certificate
+	done := make(chan struct{})
+	go func() {
+		ticker := time.NewTicker(tickFrequency)
```
mmalone: Might want to use inotify or some other file system watch here instead of a timer. I only did a timer in my HTTPS example because I didn't know how to do the file system watch.

mmalone: However, this is fine if you thing a fs watch is too much work or if there's some other issue.

autocert/examples/hello-mtls/go/client/client.go
```diff
@@ -37,34 +64,62 @@ func loadRootCertPool() (*x509.CertPool, error) {
...
 					tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
 					tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
 				},
+				// GetClientCertificate is called when a server requests a
+				// certificate from a client.
+				//
+				// In this example keep alives will cause the certificate to
+				// only be called once, but if we disable them,
+				// GetClientCertificate will be called on every request.
+				GetClientCertificate: r.getClientCertificate,
```
mmalone: Does something break if we don't disable keep alives? What's the significance of not calling `GetClientCertificate` on every request?

autocert/examples/hello-mtls/go-grpc/server/server.go
maraino: It can be done, but we need to import a 3rd party package as the inotify implementation is different depending on the OS.

autocert/examples/hello-mtls/go-grpc/server/server.go
maraino: Fixed with [64cbac4](/smallstep/certificates/pull/33/commits/64cbac4e8173f556c7e66ce1d40bdbdbc8221c5d)

autocert/examples/hello-mtls/go/client/client.go
maraino: The client will be able to talk with the client even if the certificate is expired, as long as in the initial connection the certificate was valid.

In real life, if something in the middle closes the connection or the client expires the keepalive timeout, it will work as expected, and we will get the new certificate

autocert/examples/hello-mtls/go-grpc/server/server.go
maraino: Added a comment about using https://github.com/fsnotify/fsnotify

PR #50, opened 2019-03-05
Add support for passively revoking x509 certificates

dopey:
Passive revocation, in the step ecosystem, is defined as revoking the ability of a certificate to be renewed and forcing it to expire.

NOTE: This will not stop the certificate from being valid for the rest of it's lifetime.

Adding an optional persistence layer to implement this feature.


dopey:
This PR is a work in progress.

db/db.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/smallstep/nosql"
+)
+
+// dbConfig represents the JSON attributes used for configuration. At
+// this moment only fields for NewRelic are supported.
```
maraino: There nothing here related to `NewRelic`

authority/authority.go
```diff
@@ -71,6 +77,22 @@ func New(config *Config) (*Authority, error) {
...
+	}
+	tables := [][]byte{revokedCertsTable}
+	for _, b := range tables {
+		if err = a.db.CreateTable(revokedCertsTable); err != nil {
+			return errors.Wrapf(err, "error creating table %s",
+				string(b))
+		}
+	}
+	return nil
+}
```
maraino: I think is better to move all the db logic to the new package `github.com/smallstep/certificates/db`

authority/authorize.go
```diff
@@ -157,6 +157,30 @@ func (a *Authority) Authorize(ott string) ([]interface{}, error) {
...
+		// If the error is `Not Found` then the certificate has not been revoked.
+		// Any other error should be propagated to the caller.
+		if _, err := a.db.Get(revokedCertsTable, []byte(crt.SerialNumber.String())); err != nil {
+			if err.Error() != "not found" {
+				return &apiError{
+					err:     errors.Wrap(err, "error checking revocation bucket"),
+					code:    http.StatusInternalServerError,
+					context: errContext,
+				}
+			}
```
maraino: Move db logic to `github.com/smallstep/certificates/db`
Something like 
```go
package db
type DB struct {
    db nosql.DB
}
func (d *DB) IsRevoked(serialNumber string) (bool, err) {
    ...
}
```

authority/config.go
```diff
@@ -42,6 +42,7 @@ type Config struct {
 	Address          string              `json:"address"`
 	DNSNames         []string            `json:"dnsNames"`
 	Logger           json.RawMessage     `json:"logger,omitempty"`
+	DB               json.RawMessage     `json:"db,omitempty"`
```
maraino: Creating a new db.Config type in the `github.com/smallstep/certificates/db` might be easier than using `json.RawMessage`, and will also work better in `config.Save` method.

db/db.go
```diff
@@ -0,0 +1,33 @@
...
+type dbConfig struct {
```
maraino: Rename it to `Config`

db/db.go
```diff
@@ -0,0 +1,33 @@
...
+	Type string `json:"type"`
+	Path string `json:"path"`
+}
+
+// New returns a new database client that implements the nosql.DB interface.
+func New(raw json.RawMessage) (nosql.DB, error) {
```
maraino: Change it to something like `func New(config Config) (DB, error)` where Config is the configuration above and DB is a new type that wraps all the functionality used ie: `func (db *DB) IsRevoked(sn string) (bool, error)`

api/revoke.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// ReasonStringToCode tries to convert a reason string to an integer code
+func ReasonStringToCode(reason string) (int, error) {
+	// default to 0
+	if reason == "" {
+		return 0, nil
+	}
+
+	code, found := RevocationReasonCodes[strings.ToLower(reason)]
```
maraino: We can also strip spaces so a user can write the reason in a more natural language, i.e: 
`--reason 'affiliation changed'`

api/revoke.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// RevokeRequest is the request body for a revocation request.
+type RevokeRequest struct {
+	Serial     string `json:"serial"`
+	OTT        string `json:"ott"`
+	Reason     string `json:"reason"`
+	Passive    bool   `json:"passive"`
+	CRL        bool   `json:"passive"`
+	OCSP       bool   `json:"passive"`
```
maraino: Same JSON name for 3 properties

api/revoke.go
```diff
@@ -0,0 +1,137 @@
...
+	ReasonCode int
```
maraino: It might be easier for the API to send the reasonCode, and then create the `Reason` string from it. 
Don't you think?

api/revoke.go
```diff
@@ -0,0 +1,137 @@
...
+
+// Validate checks the fields of the RevokeRequest and returns nil if they are ok
+// or an error if something is wrong.
+func (r *RevokeRequest) Validate() (err error) {
+	if r.Serial == "" {
+		return BadRequest(errors.New("missing serial"))
+	}
+	if r.ReasonCode, err = ReasonStringToCode(r.Reason); err != nil {
+		return BadRequest(err)
+	}
```
maraino: As I said before, we can receive a reason code and translate that to a string.

api/revoke.go
```diff
@@ -0,0 +1,137 @@
...
+		provisionerID string
+	)
+
+	// A token indicates that we are using revoking via a provisioner, otherwise
+	// it is assumed that the certificate is revoking itself over mTLS.
+	if len(body.OTT) > 0 {
+		// If a token is passed then Authorize the token.
+		logOtt(w, body.OTT)
+		if provisionerID, err = h.Authority.AuthorizeRevoke(body.OTT); err != nil {
+			WriteError(w, Unauthorized(err))
```
maraino: Missing `return` statement.

authority/authorize.go
```diff
@@ -58,6 +58,46 @@ func (a *Authority) Authorize(ott string) ([]interface{}, error) {
...
+func (a *Authority) AuthorizeRevoke(ott string) (string, error) {
+	var claims = Claims{}
+
+	p, err := a.authorizeToken(ott, &claims, a.revokeAudiences)
+	if err != nil {
+		return "", err
+	}
+
+	return p.ID(), nil
+}
```
maraino: Note to @maraino we need to move this to the new provisioner

authority/authorize.go
```diff
@@ -157,6 +182,21 @@ func (a *Authority) Authorize(ott string) ([]interface{}, error) {
...
+	}
+	if isRevoked {
+		return &apiError{
+			err:     errors.New("certificate has been revoked"),
+			code:    http.StatusUnauthorized,
+			context: errContext,
+		}
+	}
+
+	// Locate the step provisioner extension.
```
maraino: Note to @maraino we need to move this to the new provisioner

authority/tls.go
```diff
@@ -195,15 +196,15 @@ func (a *Authority) Renew(ocx *x509.Certificate) (*x509.Certificate, *x509.Certi
...
-		URIs:                        oldCert.URIs,
+		IsCA:                  oldCert.IsCA,
+		MaxPathLen:            oldCert.MaxPathLen,
+		MaxPathLenZero:        oldCert.MaxPathLenZero,
+		OCSPServer:            oldCert.OCSPServer,
+		IssuingCertificateURL: oldCert.IssuingCertificateURL,
+		DNSNames:              oldCert.DNSNames,
+		EmailAddresses:        oldCert.EmailAddresses,
+		IPAddresses:           oldCert.IPAddresses,
+		URIs:                  oldCert.URIs,
```
maraino: Format is weird

authority/tls.go
```diff
@@ -251,6 +252,52 @@ func (a *Authority) Renew(ocx *x509.Certificate) (*x509.Certificate, *x509.Certi
...
+			return &apiError{errors.New("no persistence layer configured"),
+				http.StatusUnauthorized, errContext}
+		}
+
+		rci := &db.RevokedCertificateInfo{
+			Serial:        serial,
+			ProvisionerID: provisionerID,
+			Reason:        reason,
+			RevokedAt:     time.Now().UTC(),
+		}
```
maraino: We should store the reason code too, it's "easier" to process than a string.

authority/tls.go
```diff
@@ -251,6 +252,52 @@ func (a *Authority) Renew(ocx *x509.Certificate) (*x509.Certificate, *x509.Certi
...
+	}
+	if rts.All || rts.CRL {
+		return &apiError{errors.New("revoke CRL unimplemented"), http.StatusUnauthorized,
+			errContext}
+	}
+	if rts.All || rts.OCSP {
+		return &apiError{errors.New("revoke OCSP unimplemented"), http.StatusUnauthorized,
+			errContext}
+	}
```
maraino: http.StatusBadRequest or http.StatusNotImplemented instead of http.StatusUnauthorized.

db/db.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// RevokedCertificateInfo contains information regarding the certificate
+// revocation action.
+type RevokedCertificateInfo struct {
+	Serial        string
+	ProvisionerID string
+	Reason        int
+	RevokedAt     time.Time
+}
```
maraino: Store the ReasonCode.

authority/tls.go
dopey: this is already the code. you also want to store the string? I'll make it more clear with the variable name that this is a code.

db/db.go
dopey: this is already the reason code. So are you asking for it to be stored as `ReasonCode` instead of `Reason`?

authority/tls.go
maraino: No, i thought it was the string.

db/db.go
maraino: I thought it was the string

api/revoke.go
dopey: Alright, so let's just receive the code and not do any conversion to string. I'll just verify that the code is within bounds. From the cli I'll allow either a `--reasonCode` or a `--reason` and do the conversion on the CLI side, sound good?

api/revoke.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	code, found := RevocationReasonCodes[strings.ToLower(reason)]
+	if !found {
+		return 0, errors.Errorf("unrecognized revocation reason '%s'", reason)
+	}
+
+	return code, nil
+}
+*/
```
maraino: Remove deprecated code.

api/revoke.go
```diff
@@ -0,0 +1,149 @@
...
+
+// RevokeRequest is the request body for a revocation request.
+type RevokeRequest struct {
+	Serial     string `json:"serial"`
+	OTT        string `json:"ott"`
+	ReasonCode int    `json:"reasonCode"`
+	Passive    bool   `json:"passive"`
+	CRL        bool   `json:"crl"`
+	OCSP       bool   `json:"ocsp"`
+	RTS        authority.RevocationTypeSelector
```
maraino: You should add `json:"-"` without it you can unmarshal a JSON like `{"RTS": {"CRL":true}}` into that variable.

In fact, as you are passing that to `Revoke` you can just build it later in the HTTP handler, and remove it from the struct.

api/revoke.go
maraino: I think it makes sense also to support an optional human reason too and store in the database.
But the only data that should be checked it would be the reasonCode.

api/revoke.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	logRevoke(w, body.Serial, provisionerID, body.ReasonCode)
+
+	if err := h.Authority.Revoke(body.RTS, body.Serial, provisionerID, body.ReasonCode); err != nil {
+		WriteError(w, Forbidden(err))
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
```
maraino: http.StatusCreated ?

api/revoke_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/smallstep/certificates/logging"
+)
+
+func TestRevokeRequestValidate(t *testing.T) {
+	type test struct {
+		rr  *RevokeRequest
+		err *Error
+		rts authority.RevocationTypeSelector
+	}
+	tests := map[string]test{
```
maraino: Not for now, but you can look into integrating [gotests](https://github.com/cweill/gotests) in your editor.

authority/authority.go
```diff
@@ -52,7 +55,16 @@ func New(config *Config) (*Authority, error) {
...
+		audiences = append(audiences, fmt.Sprintf("https://%s/sign", name),
+			fmt.Sprintf("https://%s/1.0/sign", name))
+	}
+
+	// Define revokeAudiences: legacy + possible urls without the ports.
+	// The CA might have proxies in front so we cannot rely on the port.
+	revokeAudiences := []string{legacyAuthority}
+	for _, name := range config.DNSNames {
+		revokeAudiences = append(revokeAudiences, fmt.Sprintf("https://%s/revoke", name),
+			fmt.Sprintf("https://%s/1.0/revoke", name))
```
maraino: Note to @maraino integrate this into OIDC branch

authority/authority.go
```diff
@@ -80,6 +93,18 @@ func (a *Authority) init() error {
...
+	// Initialize step-ca Database if defined in configuration.
+	if a.config.DB != nil {
+		if a.db, err = a.db.Init(a.config.DB); err != nil {
+			return err
+		}
+	} else {
+		// Set the authority DB to a nil type db.DB so that we can take advantage
+		// of methods that know when the calling object is nil and respond intelligently.
+		var db *db.DB
+		a.db = db
```
maraino: Right now only `IsRevoked` responds properly if DB is nil, it can be quite dangerous to rely on this for future changes. If you want to do something like that you probably want a noop implementation of the interface that returns whatever is expected.

authority/authorize.go
```diff
@@ -89,13 +130,15 @@ func (a *Authority) Authorize(ott string) ([]interface{}, error) {
 	}
 
 	if err = token.Claims(p.Key, &claims); err != nil {
-		return nil, &apiError{err, http.StatusUnauthorized, errContext}
+		return nil, &apiError{errors.Wrap(err, "provisioner public key cannot parse claims"),
```
maraino: Prepend error wrapping with `authorize: ` so it looks like the rest of errors.

authority/tls.go
```diff
@@ -251,6 +252,52 @@ func (a *Authority) Renew(ocx *x509.Certificate) (*x509.Certificate, *x509.Certi
...
+		}
+	}
+	if rts.All || rts.CRL {
+		return &apiError{errors.New("revoke CRL unimplemented"), http.StatusNotImplemented,
+			errContext}
+	}
+	if rts.All || rts.OCSP {
+		return &apiError{errors.New("revoke OCSP unimplemented"), http.StatusNotImplemented,
+			errContext}
+	}
```
maraino: If `rts.All` is `true` it will create a passive revocation and always return an error.

api/revoke.go
dopey: https://www.restapitutorial.com/httpstatuscodes.html

201 seems to imply that resources were created and those URIs are being returned in the response body. That's not currently the case, although this may change when we support CRL and OCSP. maybe 204 is a more logical response right now?

authority/authority.go
dopey: So the reason behind this was that `Revoke` should fail if DB is nil (you shouldn't be calling that method), but I can see that being confusing and not something we want to rely on.

api/revoke.go
maraino: We're creating a returning a JSON so 204 NoContent, and we're creating an entry in the revoke table, in any case 200 is also ok.

dopey: the JSON just says {status: ok}. I just copied it from a health request I think. Probably no good reason to return it.

maraino: Leave it there, we can add more things later.

authority/tls.go
dopey: lol, good call.

db/db.go
```diff
@@ -0,0 +1,106 @@
...
+
+// Init database with all necessary tables.
+func (db *DB) Init(c *Config) (AuthDB, error) {
+	var err error
+	if db == nil {
+		if db, err = New(c); err != nil {
+			return nil, err
+		}
+	}
```
maraino: I don't really like this, it's kind of confusing

db/db.go
```diff
@@ -0,0 +1,106 @@
...
+	Revoke(rci *RevokedCertificateInfo) error
+}
+
+// DB is a wrapper over the nosql.DB interface.
+type DB struct {
+	nosql.DB
+}
+
+// New returns a new database client that implements the nosql.DB interface.
+func New(c *Config) (*DB, error) {
```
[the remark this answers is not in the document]
maraino: You might want to return an `AuthDB` interface

db/noop.go
```diff
@@ -0,0 +1,19 @@
...
+}
+
+// IsRevoked noop
+func (n *NoopDB) IsRevoked(sn string) (bool, error) {
+	return false, nil
+}
+
+// Revoke noop
+func (n *NoopDB) Revoke(rci *RevokedCertificateInfo) error {
+	return nil
```
maraino: we can return an ErrNotImplemented error

api/revoke.go
```diff
@@ -0,0 +1,117 @@
...
+			return
+		}
+		clientCrt := r.TLS.PeerCertificates[0]
+		if clientCrt.SerialNumber.String() != body.Serial {
+			WriteError(w, Unauthorized(errors.New("client certificate serial number does not match request body")))
+			return
+		}
+	}
+
+	logRevoke(w, body.Serial, provisionerID, body.ReasonCode, body.Reason)
```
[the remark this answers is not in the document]
dopey: I definitely can grab it from there. Another reason (in my mind) for leaving it empty was as an indicator that the cert had been passively revoked over mTLS. If you think having the provisionerID is valuable regardless, then we can set another attribute for mTLS or provisioner.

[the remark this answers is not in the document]
maraino: Let's put the information if we have it. And yes, an attribute for mTLS would be good too.

authority/authorize.go
```diff
@@ -78,33 +80,92 @@ func (a *Authority) Authorize(ott string) ([]provisioner.SignOption, error) {
...
+			code:    http.StatusUnauthorized,
+			context: errContext,
+		}
+	}
+
 	p, ok := a.provisioners.LoadByCertificate(crt)
 	if !ok {
 		return &apiError{
-			err:     errors.New("provisioner not found"),
+			err:     errors.New("authorize: provisioner not found"),
```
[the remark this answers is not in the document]
dopey: do you want me to change all the prepended 'authorize' in this function to renew?

ca/client.go
```diff
@@ -346,6 +346,29 @@ func (c *Client) Renew(tr http.RoundTripper) (*api.SignResponse, error) {
 	return &sign, nil
 }
 
+// Revoke performs the revoke request to the CA and returns the api.RevokeResponse
+// struct.
+func (c *Client) Revoke(req *api.RevokeRequest) (*api.RevokeResponse, error) {
```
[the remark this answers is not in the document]
dopey: can you explain why you think another method would be better? I feel like the api is cleaner this way (while the implementation is lightly more complicated.

authority/authorize.go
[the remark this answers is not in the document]
maraino: Before we were prepending `authorize`to the `Authorize` method, this is only called from the `Renew` method, so I would say, all the errors from Renew should have `renew:` and all errors from `Sign`, `sign:`

ca/client.go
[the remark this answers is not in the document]
maraino: Client does not enforce you to use a client certificate, so if you initialize this with just the root certificate, you won't be able to reuse that client, and you have to create a new one for each one for revoke. And it might be normal to initialize the client at the start of the server and have it in the context of your server so you don't have to look for the initialization parameters again.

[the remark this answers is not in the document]
maraino: Basically, a simple method for self revoking would be good, you can even simplify that so you can automatically create the api.RevokeRequest

db/noop.go
dopey: if we're going to return that for Revoke then shouldn't we returning that for IsRevoked as well? Sort of stops being a NoopDB then though.

maraino: ... well, the cert is not revoked if NoopDB is used, but it cannot revoke.

ca/client.go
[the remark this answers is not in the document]
dopey: I think it might make more sense to just add another parameter (rather than having another method like RevokeSelf) because the other method will still need to take a request body as an argument ({Passive: true}). So I'm thinking if the transport is non-nil then create a new client, otherwise, use the existing one. What do you think? In the mean time I've implemented it as a new method, but, it's just copy-pasta.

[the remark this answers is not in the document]
maraino: I was proposing a self-revocation method, with just the transport, because I don't see a use case for self-revocating itself using only the `passive: true` option, right now it's the only thing that we support, but we would update that one in the future. 
If you don't like that option, I would prefer the API as you have it right now instead of adding an extra transport parameter to revoke, as you can always create a new client. It would be just a known issue.

api/revoke.go
dopey: The api supports two reason args; `reason` (a user supplied string that can be anything), a `reasonCode` which must be one of the OCSP.Reason integers 0-10.

authority/authority.go
```diff
@@ -111,3 +119,8 @@ func (a *Authority) init() error {
 
 	return nil
 }
+
+// Shutdown safely shuts down any clients, databases, etc. held by the Authority.
+func (a *Authority) Shutdown() error {
+	return a.db.Shutdown()
+}
```
maraino: We are not calling this method, so we are not shutting down properly the db.

dopey: Yea, I just added this last night when I was trying to figure out why my DB wasn't syncing offline updates (turned out I was being stupid and using different db file locations). Anyway, I still need to call this from other places.

api/api.go
```diff
@@ -18,19 +18,27 @@ import (
...
+	// context specific Authoirize[Sign|Revoke|etc.] methods.
 	Authorize(ott string) ([]provisioner.SignOption, error)
+	AuthorizeSign(ott string) ([]provisioner.SignOption, error)
+	AuthorizeRevoke(ott string) (provisioner.Interface, error)
 	GetTLSOptions() *tlsutil.TLSOptions
 	Root(shasum string) (*x509.Certificate, error)
 	Sign(cr *x509.CertificateRequest, opts provisioner.Options, signOpts ...provisioner.SignOption) (*x509.Certificate, *x509.Certificate, error)
 	Renew(peer *x509.Certificate) (*x509.Certificate, *x509.Certificate, error)
+	LoadProvisionerByCertificate(*x509.Certificate) (provisioner.Interface, error)
+	LoadProvisionerByToken(string) (provisioner.Interface, error)
```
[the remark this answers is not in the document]
dopey: I am using it from the cli. (offline)

authority/authority.go
maraino: Just add the call [here](https://github.com/smallstep/certificates/blob/master/ca/ca.go#L123)

authority/provisioners.go
```diff
@@ -23,3 +25,39 @@ func (a *Authority) GetProvisioners(cursor string, limit int) (provisioner.List,
...
+	if !ok {
+		return nil, &apiError{errors.Errorf("provisioner not found"),
+			http.StatusNotFound, context{}}
+	}
+	return p, nil
+}
+
+// LoadProvisionerByToken returns an interface to the provisioner that
+// provisioned the certificate.
+func (a *Authority) LoadProvisionerByToken(ott string) (provisioner.Interface, error) {
```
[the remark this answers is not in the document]
dopey: yes

api/revoke.go
```diff
@@ -0,0 +1,139 @@
...
+
+		// Load the Certificate provisioner if one exists.
+		p, err = h.Authority.LoadProvisionerByCertificate(clientCrt)
+		if err != nil {
+			fmt.Printf("err = %+v\n", err)
+			WriteError(w, err)
+			return
+		}
+		isMTLS = true
+	}
```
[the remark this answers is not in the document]
dopey: The else flow assumes that if you have a client certificate then you must have connected over mTLS (The server must have validated your certificate bundle) and therefore you are authorized to revoke your own certificate.

authority/authorize.go
```diff
@@ -24,15 +24,16 @@ type Claims struct {
 	Nonce string   `json:"nonce,omitempty"`
 }
 
-// Authorize authorizes a signature request by validating and authenticating
-// a OTT that must be sent w/ the request.
-func (a *Authority) Authorize(ott string) ([]provisioner.SignOption, error) {
+// loadProvisionerByToken parses the token and returns the provisioner used to generate
```
maraino: `authorizeToken parses the token ...`

authority/provisioner/utils_test.go
```diff
@@ -12,9 +12,9 @@ import (
 	"github.com/smallstep/cli/jose"
 )
 
-var testAudiences = []string{
-	"https://ca.smallstep.com/sign",
-	"https://ca.smallsteomcom/1.0/sign",
+var testAudiences = Audiences{
+	Sign:   []string{"https://ca.smallstep.com/sign", "https://ca.smallsteomcom/1.0/sign"},
+	Revoke: []string{"https://ca.smallstep.com/revoke", "https://ca.smallsteomcom/1.0/revoke"},
```
maraino: I suppose this was a typo `ca.smallsteomcom`

PR #54, opened 2019-03-20
Do not allow pods in one namespace to create certificates for hostnames from another namespace.

justinbarrick:
I would like to prevent pods in one namespace from minting certificates for another namespace.

This seems like a decent approach, but thought I'd open the PR to get feedback.

### DIFF
--- a/autocert/controller/main.go
+++ b/autocert/controller/main.go
@@ -55,6 +55,8 @@ type Config struct {
 	ClusterDomain                   string           `yaml:"clusterDomain"`
 }
 
+// GetClusterDomain returns the Kubernetes cluster domain, defaults to
+// "cluster.local" if not specified in the configuration.
 func (c Config) GetClusterDomain() string {
 	if c.ClusterDomain != "" {
 		return c.ClusterDomain
@@ -409,7 +411,7 @@ func shouldMutate(metadata *metav1.ObjectMeta, namespace string, clusterDomain s
 
 	subject := strings.Trim(annotations[admissionWebhookAnnotationKey], ".")
 
-	err := fmt.Errorf("Subject \"%s\" matches a namespace other than \"%s\" and is not permitted. This check can be disabled by setting restrictCertificatesToNamespace to false in the autocert-config ConfigMap.", subject, namespace)
+	err := fmt.Errorf("subject \"%s\" matches a namespace other than \"%s\" and is not permitted. This check can be disabled by setting restrictCertificatesToNamespace to false in the autocert-config ConfigMap", subject, namespace)
 
 	if strings.HasSuffix(subject, ".svc") && !strings.HasSuffix(subject, fmt.Sprintf(".%s.svc", namespace)) {
 		return false, err
--- a/autocert/controller/main.go
+++ b/autocert/controller/main.go
@@ -35,12 +35,13 @@ var (
 )
 
 const (
-	admissionWebhookAnnotationKey = "autocert.step.sm/name"
-	admissionWebhookStatusKey     = "autocert.step.sm/status"
-	provisionerPasswordFile       = "/home/step/password/password"
-	volumeMountPath               = "/var/run/autocert.step.sm"
-	tokenSecretKey                = "token"
-	tokenSecretLabel              = "autocert.step.sm/token"
+	admissionWebhookAnnotationKey         = "autocert.step.sm/name"
+	admissionWebhookStatusKey             = "autocert.step.sm/status"
+	provisionerPasswordFile               = "/home/step/password/password"
+	volumeMountPath                       = "/var/run/autocert.step.sm"
+	tokenSecretKey                        = "token"
+	tokenSecretLabel                      = "autocert.step.sm/token"
+	clusterDomain                         = "cluster.local"
 )
 
 // Config options for the autocert admission controller.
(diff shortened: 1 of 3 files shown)


maraino:
@justinbarrick, can you please sign the CLA and fix the gofmt errors:
```
autocert/controller/main.go:1::warning: file is not gofmted with -s (gofmt)
autocert/controller/main_test.go:1::warning: file is not gofmted with -s (gofmt)
```

autocert/controller/main.go
```diff
@@ -388,6 +392,34 @@ func shouldMutate(metadata *metav1.ObjectMeta) bool {
...
+	// Check if the parts of the domain match the cluster domain, if they don't
+	// it's not a cluster hostname and we can safely sign it.
+	for index, part := range clusterDomainParts {
+		if len(subjectParts) - 2 > index && subjectParts[index + 2] != part {
+			return true
+		}
+	}
+
+	// If the namespace is not the same as our pod's namespace, do not mutate the pod.
+	if subjectParts[1] != namespace {
```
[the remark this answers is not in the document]
justinbarrick: Geez, I did overcomplicate that, didn't I? I'll clean it up!

justinbarrick:
> This seems like something we should make configurable. Could put it somewhere [around here](https://github.com/smallstep/certificates/blob/master/autocert/install/02-autocert.yaml#L23).

Done!
 
> While we're at it `clusterDomain` should probably be configurable so people don't have to recompile to change it. Better yet, is there a way to discover the `clusterDomain` using the k8s API?

Done! Unfortunately, I haven't been able to find a way to discover it in the API. AFAIK, it can only be found in the kubelet flags or your resolv.conf, which seems like not the best way to do it.

> Should we have a validating webhook that runs and rejects attempts to run pods instead of just refusing to mutate the pod here? Failing silently might confuse people.

Done!

autocert/controller/main_test.go
```diff
@@ -0,0 +1,75 @@
...
+		{"right dots get stripped", "test.default.svc.cluster.local.", "default", true},
+		{"right dots get stripped wrong ns", "test.default.svc.cluster.local.", "kube-system", false},
+		{"dots get stripped", ".test.default.svc.cluster.local.", "default", true},
+		{"dots get stripped wrong ns", ".test.default.svc.cluster.local.", "kube-system", false},
+		{"partial cluster domain", "test.default.svc.cluster", "default", true},
+		{"partial cluster domain wrong ns is still allowed because not valid hostname", "test.default.svc.cluster", "kube-system", true},
+		{"service domain", "test.default.svc", "default", true},
+		{"service domain wrong ns", "test.default.svc", "kube-system", false},
+		{"two part domain", "test.default", "default", true},
+		{"two part domain different ns", "test.default", "kube-system", true},
```
justinbarrick: @mmalone I wasn't sure how to handle this case. A user requests `test.default` but they are in `kube-system`. I'm not sure if there's an unambiguous way to handle it - for example, how would we handle `example.com`? Even if we want to not allow namespaces that do not exist, there could be a race for future namespace names. I opted to go ahead and allow it and users should use full hostnames to ensure security.

autocert/controller/main_test.go
```diff
@@ -0,0 +1,75 @@
...
+		{"full cluster domain", "test.default.svc.cluster.local", "default", true},
+		{"full cluster domain wrong ns", "test.default.svc.cluster.local", "kube-system", false},
+		{"left dots get stripped", ".test.default.svc.cluster.local", "default", true},
+		{"left dots get stripped wrong ns", ".test.default.svc.cluster.local", "kube-system", false},
+		{"right dots get stripped", "test.default.svc.cluster.local.", "default", true},
+		{"right dots get stripped wrong ns", "test.default.svc.cluster.local.", "kube-system", false},
+		{"dots get stripped", ".test.default.svc.cluster.local.", "default", true},
+		{"dots get stripped wrong ns", ".test.default.svc.cluster.local.", "kube-system", false},
+		{"partial cluster domain", "test.default.svc.cluster", "default", true},
+		{"partial cluster domain wrong ns is still allowed because not valid hostname", "test.default.svc.cluster", "kube-system", true},
```
justinbarrick: @mmalone When I switched to the matching with `HasSuffix` I went ahead and dropped the check for blocking hostnames that partially match a cluster domain, because they don't actually resolve in a standard configuration so it wasn't worth the added complexity. In theory, if a user altered their search domain partial cluster domains could resolve, but this seems unlikely.

autocert/controller/main.go
```diff
@@ -385,10 +402,26 @@ func shouldMutate(metadata *metav1.ObjectMeta) bool {
...
+
+	if !restrictToNamespace {
+		return true, nil
+	}
+
+	subject := strings.Trim(annotations[admissionWebhookAnnotationKey], ".")
+
+	err := fmt.Errorf("subject \"%s\" matches a namespace other than \"%s\" and is not permitted. This check can be disabled by setting restrictCertificatesToNamespace to false in the autocert-config ConfigMap", subject, namespace)
+
+	if strings.HasSuffix(subject, ".svc") && !strings.HasSuffix(subject, fmt.Sprintf(".%s.svc", namespace)) {
```
maraino: @justinbarrick I'm not an expert on this @mmalone should now better, but is it common to have a DNS entry that finishes in `.namespace.svc`? My understanding was they finish in `.namespace.svc.clusterDoman`

If this is the case, you will never get to the last `return true, nil`

justinbarrick: In kubernetes the search domain gets set to:

```
namespace.svc.cluster.local svc.cluster.local cluster.local
```

This means that for service `hello` in namespace `world` all of the following are valid, resolvable hostnames:

* If you are in the `world` namespace, `hello`.
* `hello.world`
* `hello.world.svc`
* `hello.world.svc.cluster.local`

I’m not sure what your last sentence means, but I don’t think that is accurate and the test cases confirm the logic.

maraino: Got it didn't read the condition properly, I skipped the first part. I will wait for @mmalone to take a look at this before merging. Thanks

sourishkrout:
Taking a look at this now so we can hopefully merge this soon.

PR #55, opened 2019-03-21
Use standard x509 creating signed certificates

maraino:
### Description
This PR uses the new x509util library that uses standard x509 for creating certificates. It also puts in the right place the validation of certificates.

### DIFF
--- a/authority/tls.go
+++ b/authority/tls.go
@@ -14,7 +14,6 @@ import (
 	"github.com/smallstep/cli/crypto/pemutil"
 	"github.com/smallstep/cli/crypto/tlsutil"
 	"github.com/smallstep/cli/crypto/x509util"
-	stepx509 "github.com/smallstep/cli/pkg/x509"
 )
 
 // GetTLSOptions returns the tls options configured.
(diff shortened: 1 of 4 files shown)


authority/tls.go
```diff
@@ -77,19 +77,22 @@ func (a *Authority) Sign(csr *x509.CertificateRequest, signOpts provisioner.Opti
 		}
 	}
 
-	stepCSR, err := stepx509.ParseCertificateRequest(csr.Raw)
-	if err != nil {
-		return nil, nil, &apiError{errors.Wrap(err, "sign: error converting x509 csr to stepx509 csr"),
-			http.StatusInternalServerError, errContext}
+	if err := csr.CheckSignature(); err != nil {
```
dopey: aren't we already doing this here? https://github.com/smallstep/certificates/blob/master/api/api.go#L180-L182

maraino: Yes, but I thought it makes sense to do it internally. For example, if we add a gRPC frontend and we forget to do it in the gRPC api.

PR #56, opened 2019-03-23
TimeDuration support in sign requests.

maraino:
### Description
This PR adds support for the new type `TimeDuration` that allows the use of RFC3339 time formatted timestamps or time.Duration strings.

Fixes smallstep/cli#60

### DIFF
--- a/authority/provisioner/timeduration.go
+++ b/authority/provisioner/timeduration.go
@@ -7,6 +7,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+var now = func() time.Time {
+	return time.Now().UTC()
+}
+
 // TimeDuration is a type that represents a time but the JSON unmarshaling can
 // use a time using the RFC 3339 format or a time.Duration string. If a duration
 // is used, the time will be set on the first call to TimeDuration.Time.
@@ -25,7 +29,7 @@ func ParseTimeDuration(s string) (TimeDuration, error) {
 	// Try to use the unquoted RFC 3339 format
 	var t time.Time
 	if err := t.UnmarshalText([]byte(s)); err == nil {
-		return TimeDuration{t: t}, nil
+		return TimeDuration{t: t.UTC()}, nil
 	}
 
 	// Try to use the time.Duration string format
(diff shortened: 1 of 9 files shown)


authority/provisioner/timeduration.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface. The time is expected
+// to be a quoted string in RFC 3339 format or a quoted time.Duration string.
+func (t *TimeDuration) UnmarshalJSON(data []byte) error {
+	var s string
+	if err := json.Unmarshal(data, &s); err != nil {
+		return errors.Wrapf(err, "error unmarshaling %s", data)
+	}
+
```
dopey: Why not use ParseTimeDuration() from above here?

authority/provisioner/timeduration.go
```diff
@@ -0,0 +1,124 @@
...
+	// Try to use the time.Duration string format
+	if d, err := time.ParseDuration(s); err == nil {
+		*t = TimeDuration{d: d}
+		return nil
+	}
+
+	return errors.Errorf("failed to parse %s", data)
+}
+
+// Time set once the embedded time and returns it.
```
dopey: Time sets the embedded time and returns it.

dopey: Time calculates the embedded time.Time, sets it, and returns it.

authority/provisioner/timeduration.go
maraino: Because ParseTimeDuration receives a string like `2020-03-14T15:09:26Z` and UnmarshalJSON a string with double quotes `"2020-03-14T15:09:26Z"`

authority/provisioner/timeduration.go
```diff
@@ -0,0 +1,124 @@
...
+func (t *TimeDuration) Time() time.Time {
+	switch {
+	case t == nil:
+		return time.Time{}
+	case t.t.IsZero():
+		if t.d == 0 {
+			return time.Time{}
+		}
+		t.t = now().Add(t.d)
```
dopey: It seems odd to me that I could create a TimeDuration object with a duration, but then not call the Time() method for a while and then have the time.Time calculated way off from what I intended. Is this an implementation you've seen elsewhere?

authority/provisioner/timeduration.go
maraino: Fixed with 7b9e08b

authority/provisioner/timeduration.go
maraino: No, my first implementation was setting the time at the unmarshaling time, but then I've decided to set it (if a duration is used) only when it's going to be used, so as close as possible to the code that actually sets the validity of the certs.

dopey: I'm fine with leaving this implementation as it is. Just think it could be confusing if we ever tried to use it for something that doesn't immediately call time.Time(). Another way to fix the timing bug would have been to use a leeway (e.g. token validation). 🤷‍♀️

PR #64, opened 2019-04-12
Add token generator to ca package

maraino:
### Description
This PR adds an adaptation of the `autocert` provisioner to the `ca` package.

It also adds an stopper signal handler, and a method to update the client transport.


ca/provisioner.go
```diff
@@ -0,0 +1,194 @@
...
+	caURL         string
+	caRoot        string
+	jwk           *jose.JSONWebKey
+	tokenLifetime time.Duration
+}
+
+// NewProvisioner loads and decrypts key material from the CA for the named
+// provisioner. The key identified by `kid` will be used if specified. If `kid`
+// is the empty string we'll use the first key for the named provisioner that
+// decrypts using `passFile`.
```
dopey: passFile?

ca/provisioner.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	jwk := new(jose.JSONWebKey)
+	if err := json.Unmarshal(data, jwk); err != nil {
+		return nil, errors.Wrap(err, "error unmarshaling provisioning key")
+	}
+	return jwk, nil
+}
+
+// loadProvisionerJWKByKid retrieves a provisioner key from the CA by key ID and
+// decrypts it using the specified password file.
```
dopey: not a password file

ca/provisioner.go
```diff
@@ -0,0 +1,194 @@
...
+func loadProvisionerJWKByKid(kid, caURL, caRoot string, password []byte) (*jose.JSONWebKey, error) {
```
dopey: just a note: a kid does not uniquely identify a provisioner. The password will still probably decrypt it, but it may not be the provisioner that the user intended (e.g. could have different claims).

dopey: In this case all you're getting is the JWK so it seems like the provisioner claims don't really matter?

ca/provisioner.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		provisioners = append(provisioners, resp.Provisioners...)
+		if resp.NextCursor == "" {
+			return provisioners, nil
+		}
+		cursor = resp.NextCursor
+	}
+}
+
+// getProvisionerKey returns the encrypted provisioner key with the for the
```
dopey: getProvisionerKey returns the encrypted provisioner key for the ... (grammar)

ca/provisioner.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// getRootCAPath returns the path where the root CA is stored based on the
+// STEPPATH environment variable.
+func getRootCAPath() string {
+	return filepath.Join(config.StepPath(), "certs", "root_ca.crt")
+}
+
+// getProvisioners returns the map of provisioners on the given CA.
+func getProvisioners(caURL, rootFile string) (provisioner.List, error) {
```
dopey: I'm surprised the compiler is smart enough to understand that the function will never continue executing outside of your for loop (you don't need a separate return afterwards). Took me a minute to understand.

ca/provisioner.go
maraino: This file is almost a copy of https://github.com/smallstep/certificates/blob/master/autocert/controller/provisioner.go but instead of a `passFile` it uses a []byte password, but I didn't change the comments, I will.

ca/provisioner.go
maraino: Same reason (copy of autocert)

ca/provisioner.go
maraino: I got the autocert current behavior, but I will change it to use the name if it's not blank.

ca/provisioner.go
maraino: If there's a return or a break in the for loop, the compiler assumes that you'll get to that at some point, but you can be in an infinite loop if that does not happen.

ca/provisioner.go
maraino: if fact, no, this method does a `GET /provisioners/<kid>/encrypted-key` that returns only one encrypted key, even if multiple provisioners are defined. Authority assumes they are the same key, here the name is not required.
But it will be required to build the token. So, we need to make sure the name is always passed.

PR #65, opened 2019-04-18
Cloud identities

maraino:
### Description
This PR will add new provisioners for the following cloud identities:
* GCP instance identities
* AWS instance identities
* Azure instances using OAuth tokens

Related to https://github.com/smallstep/cli/pull/111


authority/provisioner/aws.go
```diff
@@ -0,0 +1,397 @@
...
+// will be used on certificate creation.
+func (p *AWS) AuthorizeSign(token string) ([]SignOption, error) {
+	payload, err := p.authorizeToken(token)
+	if err != nil {
+		return nil, err
+	}
+	doc := payload.document
+
+	// Enforce default DNS and IP if configured.
+	// By default we we'll accept the SANs in the CSR.
```
dopey: 'we' or 'we'll', not both.

authority/provisioner/azure.go
```diff
@@ -0,0 +1,299 @@
...
+func (p *Azure) GetType() Type {
+	return TypeAzure
+}
+
+// GetEncryptedKey is not available in an Azure provisioner.
+func (p *Azure) GetEncryptedKey() (kid string, key string, ok bool) {
+	return "", "", false
+}
+
+// GetIdentityToken retrieves the identity document and it's signature and
```
dopey: no identity doc / signature involved here. Just getting the token, as far as I can tell.

authority/provisioner/azure.go
```diff
@@ -0,0 +1,299 @@
...
+	// Validate TenantID
+	if claims.TenantID != p.TenantID {
+		return nil, errors.New("validation failed: invalid tenant id claim (tid)")
+	}
+
+	re := azureXMSMirIDRegExp.FindStringSubmatch(claims.XMSMirID)
+	if len(re) == 0 {
+		return nil, errors.Errorf("error parsing xms_mirid claim: %s", claims.XMSMirID)
+	}
+	group, name := re[2], re[3]
```
dopey: making assumption about the length of `re` here. Maybe check if it is at least 3 in the above length check.

authority/provisioner/azure.go
```diff
@@ -0,0 +1,299 @@
...
+				break
+			}
+		}
+		if !found {
+			return nil, errors.New("validation failed: invalid resource group")
+		}
+	}
+
+	// Enforce default DNS if configured.
+	// By default we we'll accept the SANs in the CSR.
```
dopey: we we'll

authority/provisioner/gcp.go
```diff
@@ -0,0 +1,299 @@
...
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/smallstep/cli/jose"
+)
+
+// gcpCertsURL is the url that servers Google OAuth2 public keys.
```
dopey: `... the url that serves ...` not servers.

authority/provisioner/gcp.go
```diff
@@ -0,0 +1,299 @@
...
+// Cloud Platform metadata API.
+//
+// If DisableCustomSANs is true, only the internal DNS and IP will be added as a
+// SAN. By default it will accept any SAN in the CSR.
+//
+// If DisableTrustOnFirstUse is true, multiple sign request for this provisioner
+// with the same instance will be accepted. By default only the first request
+// will be accepted.
+type GCP struct {
+	Type                   string   `json:"type"`
```
dopey: Why even have `Type` for any of these new provisioners if `GetType` is going to return the `GCP/AWS/AZURE type` anyway?

authority/provisioner/azure.go
maraino: If the RE succeeds it will always contain the same number of elements, if not it will contain 0.

authority/provisioner/gcp.go
maraino: So we can marshal/unmarshal that information if we need to.

authority/provisioner/aws.go
maraino: Fixed with cf07c8f

authority/provisioner/azure.go
maraino: Fixed with cf07c8f

authority/provisioner/azure.go
maraino: Fixed with cf07c8f

authority/provisioner/azure.go
dopey: ok. I feel like checking that it has the exact # of elements you expect would future proof it, but up to you.

authority/provisioner/gcp.go
dopey: Makes sense.

authority/provisioner/azure.go
maraino: sure I'll change it

authority/provisioner/aws.go
```diff
@@ -115,20 +115,28 @@ type awsInstanceIdentityDocument struct {
 // If DisableTrustOnFirstUse is true, multiple sign request for this provisioner
 // with the same instance will be accepted. By default only the first request
 // will be accepted.
+//
+// If InstanceAge is set, only the instances with an pendingTime within the
```
dopey: "... with a pendingTime ..."

authority/provisioner/collection.go
```diff
@@ -66,8 +66,21 @@ func (c *Collection) Load(id string) (Interface, bool) {
...
+	if fragment == "" {
+		audiences = c.audiences.All()
+	} else {
+		audiences = c.audiences.WithFragment(fragment).All()
+	}
+
 	// match with server audiences
-	if matchesAudience(claims.Audience, c.audiences.All()) {
+	if matchesAudience(claims.Audience, audiences) {
+		// Use fragment to get audiences (GCP, AWS)
```
dopey: Use fragment to get provisioners?

authority/provisioner/collection.go
```diff
@@ -234,3 +247,13 @@ func stripPort(rawurl string) string {
 	u.Host = u.Hostname()
 	return u.String()
 }
+
+// extractFragment extracts the
```
dopey: ... fragment?

authority/provisioner/duration.go
```diff
@@ -12,6 +12,16 @@ type Duration struct {
 	time.Duration
 }
 
+// NewDuration parses a duration string and returns a Duration type or and error
```
dopey: ... or an error ...

authority/provisioner/provisioner.go
```diff
@@ -28,10 +29,44 @@ type Audiences struct {
...
+			ret.Revoke[i] = u.ResolveReference(&url.URL{Fragment: fragment}).String()
+		} else {
+			ret.Revoke[i] = s
+		}
+	}
+	return ret
+}
+
+// generateSignAudience generates a sign audience with the format
+// https://<ca-url>/1.0/sign#provisionerID
```
dopey: this comment is misleading because ca-url has to be of the form "https://..." for it to work. If you leave off the "https://" prefix then it won't work.

authority/provisioner/collection.go
maraino: get provisioner name, I'll fix it.

authority/provisioner/provisioner.go
maraino: I'll replace it with `https://<host>/1.0/sign#provisionerID`

PR #70, opened 2019-04-25
Update nosql integration

dopey:
* shutdown and reload database on SIGHUP

### Description

💔Thank you!

### DIFF
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -3,6 +3,7 @@ package ca
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"log"
 	"net/http"
 
 	"github.com/go-chi/chi"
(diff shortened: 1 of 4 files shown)


ca/ca.go
```diff
@@ -131,6 +131,9 @@ func (ca *CA) Stop() error {
 // Reload reloads the configuration of the CA and calls to the server Reload
 // method.
 func (ca *CA) Reload() error {
+	if err := ca.auth.Shutdown(); err != nil {
+		return err
+	}
```
maraino: This is tricky as this is supposed to do a graceful reload. If something in the next lines fails, the server will still be running with the old configuration but with the database down. 

I don't see an easy way to solve this problem, a partial solution would be to pass the old db as a `With*` parameter in the New method, and inject it later in the authority. We won't be able to change the db configuration without restarting, but that can be ok.

We can decide to not to solve this, and hope for the best, if we do this, we should move the shutdown before the New, to limit the error cases to just the last 2 (New and Reload), add a comment explaining that this is not safe and perhaps some extra context to the errors.

PR #72, opened 2019-05-02
Add used OTT to DB during authToken step

dopey:


authority/authorize.go
```diff
@@ -72,11 +73,23 @@ func (a *Authority) authorizeToken(ott string) (provisioner.Interface, error) {
...
-		}); ok {
-			return nil, &apiError{errors.Errorf("authorizeToken: token already used"), http.StatusUnauthorized, errContext}
+		switch a.db.(type) {
+		case *db.NoopDB:
+			if _, ok := a.ottMap.LoadOrStore(reuseKey, &idUsed{
+				UsedAt:  time.Now().Unix(),
+				Subject: claims.Subject,
+			}); ok {
+				return nil, &apiError{errors.Errorf("authorizeToken: token already used"), http.StatusUnauthorized, errContext}
+			}
```
maraino: To avoid this type switch we could implement the ott map in the db.NoopDB, yes it will look a little weird, but it will be nicer here.

db/db.go
```diff
@@ -43,7 +45,8 @@ type DB struct {
 // New returns a new database client that implements the AuthDB interface.
 func New(c *Config) (AuthDB, error) {
 	if c == nil {
-		return new(NoopDB), nil
+		_db := new(SimpleDB)
+		return _db.Init(c)
```
maraino: Instead of doing this it will be cleaner to do just `newSimpleDB()` or `NewSimpleDB()`

authority/authorize_test.go
```diff
@@ -117,7 +117,7 @@ func TestAuthority_authorizeToken(t *testing.T) {
 					http.StatusUnauthorized, context{"ott": raw}},
 			}
 		},
-		"ok": func(t *testing.T) *authorizeTest {
+		"ok/noopdb": func(t *testing.T) *authorizeTest {
```
maraino: simpledb?

authority/authorize_test.go
```diff
@@ -133,7 +133,7 @@ func TestAuthority_authorizeToken(t *testing.T) {
 				ott:  raw,
 			}
 		},
-		"fail/token-already-used": func(t *testing.T) *authorizeTest {
+		"fail/noopdb/token-already-used": func(t *testing.T) *authorizeTest {
```
maraino: simpledb?

PR #75, opened 2019-05-10
Do not allow reload with database configuration changes.

maraino:
### Description
This PR only allows reloading the CA configuration if the database configuration hasn't changed, or is new. This avoids the stop of the server if something goes wrong.

Fixes #smallstep/ca-component#170

### DIFF
--- a/authority/authority.go
+++ b/authority/authority.go
@@ -32,8 +32,8 @@ type Authority struct {
 // Option sets options to the Authority.
 type Option func(*Authority)
 
-// WithDatabase sets an already initialize authority database to a new
-// authority. This options is intended to be use on graceful reloads.
+// WithDatabase sets an already initialized authority database to a new
+// authority. This option is intended to be use on graceful reloads.
 func WithDatabase(db db.AuthDB) Option {
 	return func(a *Authority) {
 		a.db = db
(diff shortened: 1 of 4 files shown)


authority/authority.go
```diff
@@ -55,11 +69,12 @@ func (a *Authority) init() error {
 	}
 
 	var err error
-
-	// Initialize step-ca Database.
+	// Initialize step-ca Database if it's not already initialize with WithDB.
```
dopey: initialize -> initialized

maraino: Fixed with fb6a1af

authority/authority.go
```diff
@@ -118,6 +133,12 @@ func (a *Authority) init() error {
 	return nil
 }
 
+// GetDatabase returns the authority database. It will return null if it's not
```
dopey: The comment is "sort of" correct. Generally speaking this method will return the SimpleDB (since that is the default if nothing is configured). But if you create an authority w/out running `.Init`, then you'll get nil.

maraino: @dopey Does it makes sense to allow the switch from SimpleDB to Badger or any other type?

authority/authority.go
```diff
@@ -29,8 +29,19 @@ type Authority struct {
 	initOnce bool
 }
 
+// Option sets options to the Authority.
+type Option func(*Authority)
+
+// WithDatabase sets an already initialize authority database to a new
```
dopey: already initialized

authority/authority.go
```diff
@@ -29,8 +29,19 @@ type Authority struct {
...
+// authority. This options is intended to be use on graceful reloads.
```
dopey: this option

authority/authority.go
dopey: I believe that we've decided that we won't support switching databases for reload.

dopey: I think you should just say "GetDatabase returns the authority database.", without the rest. It won't return nil, if the DB is not configured.

PR #76, opened 2019-06-04
Acme Support

dopey:


api/api.go
```diff
@@ -249,6 +255,28 @@ func (h *caHandler) Route(r Router) {
...
+
+	r.MethodFunc("POST", ah.Dir.NewAccount, extractPayloadByJWK(ah.NewAccount))
+	r.MethodFunc("POST", ah.Dir.GetAccount("{accID}", false), extractPayloadByKid(ah.UpdateAccount))
+	r.MethodFunc("POST", ah.Dir.NewOrder, extractPayloadByKid(ah.NewOrder))
+	r.MethodFunc("POST", ah.Dir.GetOrder("{ordID}", false), extractPayloadByKid(ah.GetOrder))
+	r.MethodFunc("POST", ah.Dir.GetFinalize("{ordID}", false), extractPayloadByKid(ah.FinalizeOrder))
+	//r.MethodFunc("POST", ah.Dir.GetFinalize("{ordID}"), extractPayloadByKid(ah.Finalize))
+	r.MethodFunc("POST", ah.Dir.GetAuthz("{authzID}", false), extractPayloadByKid(ah.GetAuthz))
+	r.MethodFunc("POST", ah.Dir.GetChallenge("{chID}", false), extractPayloadByKid(ah.GetChallenge))
+	r.MethodFunc("POST", ah.Dir.GetCertificate("{certID}", false), extractPayloadByKid(ah.GetCertificate))
```
maraino: To split the code better we can Add to the RouterHandler interface the method `Route(pattern string, fn func(r Router)) Router` and use have a new Route method in `api/acme.go`, more or less in the same way that we add the `/1.0` prefix in https://github.com/smallstep/certificates/blob/master/ca/ca.go#L106-L108

So the routes will become for example `/acme/directory` or `/1.0/acme/directory`

api/api.go
```diff
@@ -227,13 +230,16 @@ type FederationResponse struct {
...
+	Authority   Authority
+	acmeHandler *acmeHandler
 }
 
 // New creates a new RouterHandler with the CA endpoints.
 func New(authority Authority) RouterHandler {
 	return &caHandler{
 		Authority: authority,
+		acmeHandler: newACMEHandler(authority, authority.GetDatabase().(nosql.DB),
+			"ca.smallstep.com:8080", "acme", nil),
```
maraino: I suppose this is just for testing right?

api/api.go
```diff
@@ -40,6 +42,7 @@ type Authority interface {
 	GetEncryptedKey(kid string) (string, error)
 	GetRoots() (federation []*x509.Certificate, err error)
 	GetFederation() ([]*x509.Certificate, error)
+	GetDatabase() db.AuthDB
```
maraino: I don't like to have access to the DB at the `/api` level, if we need to access to the DB we should probably go through a layer like the authority package.

api/acme/order.go
```diff
@@ -0,0 +1,308 @@
...
+	NotAfter       string       `json:"notAfter,omitempty"`
+	Error          interface{}  `json:"error,omitempty"`
+	Authorizations []string     `json:"authorizations"`
+	Finalize       string       `json:"finalize"`
+	Certificate    string       `json:"certificate,omitempty"`
+}
+
+// ToACME converts the internal Order type into the public acmeOrder type for
+// presentation in the ACME protocol.
+func (o *Order) ToACME(db nosql.DB, dir *Directory) (interface{}, error) {
```
maraino: I think the ToACME methods are used to print JSON, we should be able to rename without a lot of effort this to MarshalJSON, but we might need to inject the arguments to the order (an app context). 

Basically if we want to return an Order as json we should be able to do:
```go
o := NewOrder(...)
// populate o somehow 
JSON(w, o)
```

api/acme/directory.go
```diff
@@ -0,0 +1,155 @@
...
+	NewNonce   string `json:"newNonce,omitempty"`
+	NewAccount string `json:"newAccount,omitempty"`
+	NewOrder   string `json:"newOrder,omitempty"`
+	NewAuthz   string `json:"newAuthz,omitempty"`
+	RevokeCert string `json:"revokeCert,omitempty"`
+	KeyChange  string `json:"keyChange,omitempty"`
+}
+
+// ToACME returns an ACME representation for the internal Directory type.
+func (d *Directory) ToACME() interface{} {
```
maraino: See my comment in Order.ToACME

api/acme/directory.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+}
+
+// GetAccount returns the location for retrieving Account info by ID.
+func (d *Directory) GetAccount(id string, abs bool) string {
+	p := fmt.Sprintf("/%s/%s/%s", d.prefix, d.ops.Account, id)
+	if abs {
+		p = fmt.Sprintf("https://%s%s", d.dns, p)
+	}
+	return p
```
maraino: It looks like this code is repeated, we can unify it using a new method with the necessary arguments.

api/acme/directory.go
```diff
@@ -0,0 +1,155 @@
...
+	RevokeCert:  "revoke-cert",
+	KeyChange:   "key-change",
+	Certificate: "cert",
+	Directory:   "directory",
+}
+
+// NewDirectory returns a new Direcotyr type.
+func NewDirectory(dns, prefix string, ops *DirectoryOptions) *Directory {
+	if ops == nil {
+		ops = &defaultDirOptions
```
maraino: If we don't need to change this, just get rid of `ops`.

api/acme/authz.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// GetExpiry returns the expiration time of the authz.
+func (ba *BaseAuthz) GetExpiry() time.Time {
+	return ba.Expires
+}
+
+// ToACME converts the internal Authz type into the public acmeAuthz type for
+// presentation in the ACME protocol.
+func (ba *BaseAuthz) ToACME(db nosql.DB, dir *Directory) (interface{}, error) {
```
maraino: See my comment in Order.ToACME

api/acme/account.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// IsValid returns true is the Account is valid.
+func (a *Account) IsValid() bool {
+	return a.Status == statusValid
+}
+
+// ToACME converts the internal Account type into the public acmeAccount
+// type for presentation in the ACME protocol.
+func (a *Account) ToACME(db nosql.DB, dir *Directory) (interface{}, error) {
```
maraino: See my comment in Order.ToACME

api/acme.go
```diff
@@ -0,0 +1,849 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/go-chi/chi"
+	"github.com/pkg/errors"
+	"github.com/smallstep/certificates/api/acme"
```
maraino: This should be moved to a different path `certificate/acme` can be a good place.

api/acme.go
```diff
@@ -0,0 +1,849 @@
...
+	"github.com/smallstep/cli/jose"
+	"github.com/smallstep/nosql"
+)
+
+func link(url, typ string) string {
+	return fmt.Sprintf("<%s>;\"%s\"", url, typ)
+}
+
+type acmeHandler struct {
```
maraino: acmeHandler should have a Route method, see my comment in api.go

api/acme.go
```diff
@@ -0,0 +1,849 @@
...
+	auth   Authority
+	db     nosql.DB
+	Dir    *acme.Directory
+	domain string
+	prefix string
+}
+
+// newACMEHandler returns a new acmeHandler type.
+func newACMEHandler(auth Authority, db nosql.DB, domain, prefix string, dirOpts *acme.DirectoryOptions) *acmeHandler {
```
maraino: `api/acme.go` should be "dumb" it doesn't have to have access to the DB, it's function must be:
1. Read data from the request
2. Build the necessary objects
3. Pass those objects to an `Authority` like interface
4. Get the response of that interface
5. Send the the response to the user.

api/acme.go
```diff
@@ -0,0 +1,849 @@
...
+	}
+
+	if !(payload.isPostAsGet) {
+		WriteError(w, BadRequest(errors.Errorf("unexpected payload")))
+		return
+	}
+
+	// Load the challenge.
+	certID := chi.URLParam(r, "certID")
+	crtBytes, err := a.db.Get([]byte("x509_certs"), []byte(certID))
```
maraino: Who is writing here?

api/api.go
dopey: yea

api/acme/order.go
dopey: They are used to output JSON, but I'm already using the default json marshaler when writing the objects to the DB. The difference here is that a number of fields are hidden when output as an acme response (e.g. ID, Created, etc.)

maraino: It will be easy to clean this once the logic is split on a different layer, the return object of ToACME should be the return object of the method doing the logic.

ca/ca.go
```diff
@@ -103,8 +103,15 @@ func (ca *CA) Init(config *authority.Config) (*CA, error) {
 	// Add api endpoints in / and /1.0
 	routerHandler := api.New(auth)
 	routerHandler.Route(mux)
+	mux.Route("/acme", func(r chi.Router) {
+		routerHandler.RouteACME(r)
+	})
 	mux.Route("/1.0", func(r chi.Router) {
-		routerHandler.Route(r)
```
maraino: You've removed the old handler

api/acmeMiddleware.go
```diff
@@ -0,0 +1,282 @@
...
+		jwk := jws.Signatures[0].Protected.JSONWebKey
+		if jwk == nil {
+			WriteError(w, acme.MalformedErr(errors.Errorf("expected jwk in protected header")))
+			return
+		}
+		ctx = context.WithValue(ctx, jwkContextKey, jwk)
+
+		acc, err := h.Authority.GetAccountByKeyID(jwk.KeyID)
+		switch {
+		case nosql.IsErrNotFound(err):
```
maraino: I would check this in GetAccountByKeyID and return an appropriate error.

api/api.go
maraino: We don't need it.

acme/authority.go
```diff
@@ -0,0 +1,236 @@
...
+import (
+	"crypto/x509"
+
+	"github.com/pkg/errors"
+	"github.com/smallstep/cli/jose"
+	"github.com/smallstep/nosql"
+)
+
+// ACME is the acme authority interface.
+type ACME interface {
```
maraino: Rename it to Interface (it will be acme.Interface)

acme/authority.go
```diff
@@ -0,0 +1,236 @@
...
+	GetOrder(string, string) (*Order, error)
+	FinalizeOrder(string, string, *x509.CertificateRequest) (*Order, error)
+	GetAuthz(string, string) (*Authz, error)
+	GetChallenge(string, string) (*Challenge, error)
+	ValidateChallenge(string, string, *jose.JSONWebKey) (*Challenge, error)
+	GetCertificate(string, string) ([]byte, error)
+}
+
+// Authority is the layer that handles all ACME interactions.
+type Authority struct {
```
maraino: Rename it to ACME

maraino: Well, authority can be ok as it will be `acme.Authority`, but if you can find a different would be better, to not confuse it with the regular authority.

acme/api/middleware.go
```diff
@@ -0,0 +1,308 @@
...
+			if err != nil {
+				api.WriteError(w, acme.ServerInternalErr(errors.Wrap(err, "error generating jwk thumbprint")))
+				return
+			}
+			jwk.KeyID = base64.RawURLEncoding.EncodeToString(kid)
+		}
+		acc, err := h.Auth.GetAccountByKeyID(jwk.KeyID)
+		switch {
+		case nosql.IsErrNotFound(err):
+			break
```
maraino: Is this expected?

acme/api/order.go
```diff
@@ -0,0 +1,186 @@
...
+		if id.Type != "dns" {
+			return acme.MalformedErr(errors.Errorf("identifier type unsupported: %s", id.Type))
+		}
+	}
+	now := time.Now().UTC()
+	if n.NotBefore.IsZero() {
+		n.NotBefore = now
+	}
+	if n.NotAfter.IsZero() {
+		n.NotAfter = now.Add(time.Hour * 24)
```
maraino: Is 24h in the spec? Is this a good enough value or a configuration value like `defaultTLSCertDuration`?

acme/api/order.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// Validate validates a finalize request body.
+func (f *FinalizeRequest) Validate() error {
+	var err error
+	csrBytes, err := base64.RawURLEncoding.DecodeString(f.CSR)
+	if err != nil {
+		return acme.MalformedErr(errors.Wrap(err, "error base64url decoding csr"))
+	}
+	f.csr, err = x509.ParseCertificateRequest(csrBytes)
```
maraino: are we calling `CheckSignature()` at some point?

acme/authz.go
```diff
@@ -0,0 +1,327 @@
...
+		return nil, Wrap(err, "error creating http challenge")
+	}
+	ch2, err := newDNS01Challenge(db, ChallengeOptions{
+		AccountID:  accID,
+		AuthzID:    ba.ID,
+		Identifier: identifier})
+	if err != nil {
+		return nil, Wrap(err, "error creating dns challenge")
+	}
+	ba.Challenges = []string{ch1.getID(), ch2.getID()}
```
maraino: It might be ok, but looks weird that `newDNSAuthz` returns an http-01 and a dns-01 challenge.

authority/config.go
```diff
@@ -46,6 +47,7 @@ type Config struct {
 	DNSNames         []string            `json:"dnsNames"`
 	Logger           json.RawMessage     `json:"logger,omitempty"`
 	DB               *db.Config          `json:"db,omitempty"`
+	ACME             *acme.Config        `json:"acme,omitempty"`
```
maraino: It should be a provisioner.

ca/ca.go
```diff
@@ -100,13 +106,47 @@ func (ca *CA) Init(config *authority.Config) (*CA, error) {
...
+		prefix := config.ACME.Prefix
+		acmeAuth := acme.NewAuthority(auth.GetDatabase().(nosql.DB), dns, prefix, auth)
+		acmeRouterHandler := acmeAPI.New(acmeAuth)
+		mux.Route("/"+prefix, func(r chi.Router) {
+			acmeRouterHandler.Route(r)
+		})
+		mux.Route("/1.0/"+prefix, func(r chi.Router) {
+			acmeRouterHandler.Route(r)
+		})
+	}
```
maraino: This block will change to support different acme provisioners.

ca/ca.go
```diff
@@ -100,13 +106,47 @@ func (ca *CA) Init(config *authority.Config) (*CA, error) {
 	mux := chi.NewRouter()
 	handler := http.Handler(mux)
 
-	// Add api endpoints in / and /1.0
+	// Add regular CA api endpoints in / and /1.0
 	routerHandler := api.New(auth)
 	routerHandler.Route(mux)
+	mux.Route("/", func(r chi.Router) {
+		routerHandler.Route(r)
+	})
```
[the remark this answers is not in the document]
dopey: true

Gopkg.toml
```diff
@@ -49,7 +49,7 @@ required = [
   name = "github.com/smallstep/cli"
 
 [[constraint]]
-  branch = "master"
+  branch = "cmpandswap"
```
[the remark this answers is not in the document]
dopey: yep, master is already compatible. Forgot to change it here. good catch

acme/account.go
```diff
@@ -0,0 +1,197 @@
...
+			"value has changed since last read"))
+	default:
+		return nil
+	}
+}
+
+// update updates the acme account object stored in the database if,
+// and only if, the account has not changed since the last read.
+func (a *account) update(db nosql.DB, contact []string) (*account, error) {
+	b := *a
```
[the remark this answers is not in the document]
dopey: Yep, good to note. I think in this situation that's ok. Because I'm not changing `key`, and I'm just replacing contact.

acme/api/account.go
```diff
@@ -0,0 +1,207 @@
...
+
+// NewAccountRequest represents the payload for a new account request.
+type NewAccountRequest struct {
+	Contact            []string
+	OnlyReturnExisting bool
+}
+
+func validateContacts(cs []string) error {
+	for _, c := range cs {
+		if len(c) == 0 {
```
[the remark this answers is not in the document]
dopey: personally, I don't really care (as you can tell since I'm barely doing any validation). I don't really see us using the contact info for anything. Maybe one of those things where we let users tell us if they would appreciate us being more strict?

acme/api/order.go
dopey: The spec doesn't mention a default duration. This is just to match what we're already doing. I can open up the question to the group to see if we think acme provisioners should have a longer default duration. I think this will be a bit cleaner once the acme provisioner is implemented, because then the defaults will be in the provisioner.

acme/api/order.go
dopey: we are not. because we forgot.

acme/authz.go
dopey: I know ... Just that it's `type` is 'dns`. It goes along with the spec but I'm happy to rename it if you, or other folks, would prefer.

authority/config.go
dopey: Let's chat about this when you feel better. I was under the impression that acme provisioners would go into the existing provisioner list. As I'm thinking about them right now, we should be able to have many acme provisioners, each associated with it's own acme account.

ca/ca.go
dopey: We can hash this out when we talk in more detail about how to implement acme provisioners. If there's a 1:1 relationship between an acme provisioner and an acme account then I don't see a need to offer a different api, unless you're planning on serving different resources.

acme/api/middleware.go
dopey: yea. When you're making a NewAccount request there will be a jwk in the jws header which you will later use when creating the new account (so you won't be able to look the account up by kid b/c it doesn't exist yet). This is the only case in which `context.acc == nil` is not an error.

I'll leave a comment in the code.

acme/api/middleware.go
```diff
@@ -176,9 +188,18 @@ func (h *Handler) extractJWK(next nextHTTP) nextHTTP {
...
 		ctx = context.WithValue(ctx, jwkContextKey, jwk)
 
+		if len(jwk.KeyID) == 0 {
+			kid, err := jwk.Thumbprint(crypto.SHA256)
+			if err != nil {
+				api.WriteError(w, acme.ServerInternalErr(errors.Wrap(err, "error generating jwk thumbprint")))
+				return
+			}
+			jwk.KeyID = base64.RawURLEncoding.EncodeToString(kid)
+		}
```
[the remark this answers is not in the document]
dopey: Is this a question, or a comment, or an observation?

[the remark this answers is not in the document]
maraino: An observation, I know that some `kid` values are URLs, so just making sure you're taking this into account if it can have any effect here.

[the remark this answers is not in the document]
dopey: Ah, yes so you're correct in that some kid values are urls. However context is king. The 'kid' protected header of the JWS (body of the request) must always be the account url if it is defined. If it is not defined, then the 'jwk' protected header must contain the public JWK, and we use the kid of the jwk to look up the account. That is the case in this method. We are looking up the account using a JWK, specifically the 'kid' of the JWK.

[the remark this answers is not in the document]
dopey: The 'jwk' and 'kid' protected headers are mutually exclusive and one of them must be defined.

acme/api/order.go
```diff
@@ -65,9 +66,29 @@ func (f *FinalizeRequest) Validate() error {
...
+			"ordExpires":        o.Expires,
+			"ordIdentifiers":    o.Identifiers,
+			"ordNotBefore":      o.NotBefore,
+			"ordNotAfter":       o.NotAfter,
+			"ordError":          o.Error,
+			"ordAuthorizations": o.Authorizations,
+			"ordFinalize":       o.Finalize,
+			"ordCertificate":    o.Certificate,
+		}
+		rl.WithFields(m)
```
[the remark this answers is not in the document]
dopey: implemented with an interface as you recommended. Unfortunately logrus looks to be escaping whatever we pass so the result is a marshaled object with lots of escapes e.g. `response="{\"newNonce\":\"https://ca.smallstep.com:8080/acme/new-nonce\",\"newAccount\":\"https://ca.smallstep.com:8080/acme/new-account\",\"newOrder\":\"https://ca.smallstep.com:8080/acme/new-order\",\"revokeCert\":\"https://ca.smallstep.com:8080/acme/revoke-cert\",\"keyChange\":\"https://ca.smallstep.com:8080/acme/key-change\"}"`

acme/api/middleware_test.go
```diff
@@ -904,8 +904,8 @@ func TestHandlerValidateJWS(t *testing.T) {
 		"fail/more-than-one-signature": func(t *testing.T) test {
 			jws := &jose.JSONWebSignature{
 				Signatures: []actualjose.Signature{
-					actualjose.Signature{},
-					actualjose.Signature{},
+					{},
+					{},
```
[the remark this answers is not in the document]
dopey: fixed

acme/api/middleware.go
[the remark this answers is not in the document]
dopey: New question as to whether we should always re-calculate the keyid. otherwise a bad actor could pass in a JWK with incorrect key-id to load account information for an account they don't own. They wouldn't be able to make any other requests though as the signature in the JWS would not pass the sig check.

acme/nonce.go
```diff
@@ -26,7 +26,7 @@ func newNonce(db nosql.DB) (*nonce, error) {
 	id := base64.RawURLEncoding.EncodeToString([]byte(_id))
 	n := &nonce{
 		ID:      id,
-		Created: time.Now().UTC().Round(time.Second),
+		Created: round(time.Now().UTC()),
```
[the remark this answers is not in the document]
dopey: fixed

acme/errors.go
```diff
@@ -400,31 +398,18 @@ func (e *Error) Cause() error {
...
-	Type        string        `json:"type"`
-	Detail      string        `json:"detail"`
-	Identifier  interface{}   `json:"identifier,omitempty"`
-	Subproblems []interface{} `json:"subproblems,omitempty"`
-}
-
-// toACME returns an acme representation of the problem type.
-func (e *Error) toACME() *AError {
+// ToACME returns an acme representation of the problem type.
+func (e *Error) ToACME() *AError {
```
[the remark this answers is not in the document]
dopey: The problem with using a marshaller is that I also need to unmarshal the objects (they are stored on authzs and challenges when they are written to disk after a validation error). They are kinda tricky to unmarshal. Maybe I'm just tired and therefore not thinking well, but I felt it was easier to have this intermediate object.

acme/api/middleware.go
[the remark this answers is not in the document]
dopey: I replaced using the jwk.KeyID by always recalculating the keyID when looking up and storing the account information.

acme/api/account.go
```diff
@@ -0,0 +1,207 @@
...
+func (u *UpdateAccountRequest) IsDeactivateRequest() bool {
+	return u.Status == acme.StatusDeactivated
+}
+
+// Validate validates a update-account request body.
+func (u *UpdateAccountRequest) Validate() error {
+	if len(u.Status) > 0 && len(u.Contact) > 0 {
+		return acme.MalformedErr(errors.New("incompatible input; contact and " +
+			"status updates are mutually exclusive"))
+	}
```
maraino: This can be the first case in the switch statement too.

acme/api/account.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil
+	default:
+		return acme.MalformedErr(errors.Errorf("empty update request"))
+	}
+}
+
+// NewAccount is the handler resource for creating new ACME accounts.
+func (h *Handler) NewAccount(w http.ResponseWriter, r *http.Request) {
+	prov, ok := provisionerFromContext(r)
+	if !ok || prov == nil {
```
maraino: Is `ok == true && prov == nil` an actual option, if it is, does it make sense to check it in provisionerFromContext instead of here? 
You're doing a similar check in other places, but for example, you are not doing it on line 97 for `accountFromContext`, if there's a possibility and you always want to display an error we should return false.

acme/api/middleware.go
```diff
@@ -0,0 +1,366 @@
...
+			return
+		}
+		hdr := sig.Protected
+		switch hdr.Algorithm {
+		case jose.RS256, jose.RS384, jose.RS512, jose.ES256, jose.ES384, jose.ES512, jose.EdDSA:
+			// we good
+		default:
+			api.WriteError(w, acme.MalformedErr(errors.Errorf("unsuitable algorithm: %s", hdr.Algorithm)))
+			return
+		}
```
maraino: Should we enforce a minimum bit length for RSA keys?

authority/provisioner/acme.go
```diff
@@ -0,0 +1,78 @@
...
+		profileDefaultDuration(p.claimer.DefaultTLSCertDuration()),
+		newProvisionerExtensionOption(TypeACME, p.Name, ""),
+		newValidityValidator(p.claimer.MinTLSCertDuration(), p.claimer.MaxTLSCertDuration()),
+	}, nil
+}
+
+// AuthorizeRenewal is not implemented for the ACME provisioner.
+func (p *ACME) AuthorizeRenewal(cert *x509.Certificate) error {
+	return errors.New("acme provisioner does not implement AuthorizeRenewal")
+}
```
maraino: I thought we wanted to allow the use of `/renew` for acme certificates, we should implement this.

dopey: Yea, @mmalone mentioned this as well. I thought we were ok with our existing renew flow. I'll either find time to add it before we release ACME or it will be a quick follow. Will make a new issue for it now.

acme/api/middleware.go
dopey: enforcing 2048. good spot.

authority/provisioner/acme.go
dopey: Shoot! I misunderstood your comment. You're right, we do need this to be allowed. I'll fix this. The API/route @mmalone was talking about is a separate concern.

authority/provisioner/collection.go
```diff
@@ -127,6 +127,8 @@ func (c *Collection) LoadByCertificate(cert *x509.Certificate) (Interface, bool)
 				return c.Load("aws/" + string(provisioner.Name))
 			case TypeGCP:
 				return c.Load("gcp/" + string(provisioner.Name))
+			case TypeACME:
+				return c.Load(string(provisioner.Name))
```
maraino: Here we are loading by id, I don't think we are currently enforcing unique names, only unique ids, so I think we should follow the logic used in aws or gcp and use `acme/name`.

ca/ca.go
```diff
@@ -113,27 +113,29 @@ func (ca *CA) Init(config *authority.Config) (*CA, error) {
...
+	if err != nil {
+		return nil, err
 	}
+	port := u.Port()
+	if port != "" && port != "443" {
+		dns = fmt.Sprintf("%s:%s", dns, port)
+	}
+
+	prefix := "acme"
+	acmeAuth := acme.NewAuthority(auth.GetDatabase().(nosql.DB), dns, prefix, auth)
```
maraino: We should think if it makes sense to add support for multiple DNS entries and always use the one used by the user in the initialization (account creation?)

docs/acme.md
```diff
@@ -0,0 +1,89 @@
...
+certificate](https://smallstep.com/docs/cli/ca/root/) at
+`~/.step/certs/root_ca.crt`).
+
+## Enabling ACME
+
+To enable ACME, simply [add an ACME provisioner](https://smallstep.com/docs/cli/ca/provisioner/add/) to your `step-ca` configuration
+by running:
+
+```
+$ step ca provisioner add acme --type ACME
```
maraino: Clarify that acme is the provisioner name.

docs/acme.md
```diff
@@ -0,0 +1,89 @@
...
+
+> Caution: adding a root certificate to your system’s trust store is a global
+> operation. Certificates issued by your CA will be trusted everywhere,
+> including in web browsers.
+
+`step-ca` should work with any ACMEv2
+([RFC8555](https://tools.ietf.org/html/rfc8555)) compliant client that supports
+the http-01 or dns-01 challenge. If you run into any issues please let us know
+[on gitter](https://gitter.im/smallstep/community) or [in an
+issue](https://github.com/smallstep/certificates/issues/new?template=bug_report.md).
```
maraino: Add example using certbot or the client recommended by let's encrypt.

dopey: I added the certbot example from the blog post.

PR #83, opened 2019-06-25
Add option for checking group membership declared in JWT token

matteo-s:
### Description
This PR adds the ability to check group membership inside OIDC provisioners.

### Feature 
By adding a `groups:['groupname']` declaration to the provisioner configuration, the validation code will check 
* if the received JWT token contains a `groups` claim
* if any of the provided group names is contained in such claim

Do note that `groups` is not a *standard* claim (as declared in https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims), but nevertheless many providers support it (Azure, Auth0, Okta)

### Context
We have a setup where users are organized in groups, and we want to filter access to the CA.
We can not leverage domain filtering, since all users of the various groups have the same email domain. 
Ideally, we'd like to have a configurable mechanism which allows us to define (for each provisioner) the claim name and the values to match. For example, we could match against the `roles` claim.
I do not know if you guys have something similar in mind.
By looking at the code, I've seen some specific checks added to azure and gcp to verify claims like *tenantIds*, *projectIds* etc.

I've taken the same approach and added a check for the *groups* field, but maybe something configurable could be shared between the various providers, consolidating the checks and providing additional filtering capabilities to the OIDC provider.

### DIFF
--- a/authority/provisioner/oidc.go
+++ b/authority/provisioner/oidc.go
@@ -33,12 +33,13 @@ func (c openIDConfiguration) Validate() error {
 // openIDPayload represents the fields on the id_token JWT payload.
 type openIDPayload struct {
 	jose.Claims
-	AtHash          string `json:"at_hash"`
-	AuthorizedParty string `json:"azp"`
-	Email           string `json:"email"`
-	EmailVerified   bool   `json:"email_verified"`
-	Hd              string `json:"hd"`
-	Nonce           string `json:"nonce"`
+	AtHash          string   `json:"at_hash"`
+	AuthorizedParty string   `json:"azp"`
+	Email           string   `json:"email"`
+	EmailVerified   bool     `json:"email_verified"`
+	Hd              string   `json:"hd"`
+	Nonce           string   `json:"nonce"`
+	Groups          []string `json:"groups"`
 }
 
 // OIDC represents an OAuth 2.0 OpenID Connect provider.
@@ -52,6 +53,7 @@ type OIDC struct {
 	ConfigurationEndpoint string   `json:"configurationEndpoint"`
 	Admins                []string `json:"admins,omitempty"`
 	Domains               []string `json:"domains,omitempty"`
+	Groups                []string `json:"groups,omitempty"`
 	Claims                *Claims  `json:"claims,omitempty"`
 	configuration         openIDConfiguration
 	keyStore              *keyStore
@@ -184,6 +186,22 @@ func (o *OIDC) ValidatePayload(p openIDPayload) error {
 		}
 		if !found {
 			return errors.New("failed to validate payload: email is not allowed")
+		}
+	}
+
+	// Filter by oidc group claim
+	if len(o.Groups) > 0 {
+		var found bool
+		for _, group := range o.Groups {
+			for _, g := range p.Groups {
+				if g == group {
+					found = true
+					break
+				}
+			}
+		}
+		if !found {
+			return errors.New("validation failed: invalid group")
 		}
 	}
 


mmalone:
Hi @matteo-s,

Thanks for the PR. I'll let @mariano or @dopey review the code but conceptually this is great.

> Do note that `groups` is not a standard claim

This is the main reason we didn't include this feature originally. We use GSuite and do most of our testing there, and GSuite doesn't include this claim. But the fact that it's supported by Azure, Auth0, and Okta tells me it's approaching de-facto standardization. It's hard to imagine how a claim called `groups` would ever have a semantics different enough to break this functionality. I'm cautiously onboard with this change, but perhaps we should mark it as beta for a little while (in docs) and see if anyone finds any issues with other providers.

> By looking at the code, I've seen some specific checks added to azure and gcp to verify claims like tenantIds, projectIds etc.

I think what you're seeing is some new code to support *instance identity documents* on [AWS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html) [GCP](https://cloud.google.com/compute/docs/instances/verifying-instance-identity) and [Azure](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview). The code is in master but we haven't officially released / announced it yet.

> Ideally, we'd like to have a configurable mechanism which allows us to define (for each provisioner) the claim name and the values to match.

We're into capturing low-hanging fruit like `groups` in an ad-hoc way, but my concern is that the various *ands* and *ors* and *nots* that you might want (e.g., *group X and role Y* OR *group A*) will be hard to support without something more sophisticated... Before we go off and try to anticipate what people need we've been waiting to collect use cases. This is a good start, and if you have any more it'd be great to hear them!

Thanks again for contributing! This is great!

maraino:
Hi @matteo-s, the change looks good, but you need to [sign the CLA](https://cla-assistant.io/smallstep/certificates?pullRequest=83) before we can merge it, and I will appreciate a unit test so we can verify in the future that we don't break this feature.

matteo-s:
Hi @mmalone 

>  It's hard to imagine how a claim called groups would ever have a semantics different enough to break this functionality.

I do agree, it's indeed difficult to think of a situation where a provider would express something different than *group membership* in a custom claim named *groups* :) 
Nevertheless, it is a custom claim, so it's up to you guys to decide if it's worth including it.

> We're into capturing low-hanging fruit like groups in an ad-hoc way, but my concern is that the various ands and ors and nots that you might want (e.g., group X and role Y OR group A) will be hard to support without something more sophisticated...

I fully agree, I was brainstorming around the idea of a configurable system for some other use cases, but the complexity of building something that would satisfy any set of conditions is probably out of scope for your project.

Maybe you could consider integrating something like Open Policy Agent (https://www.openpolicyagent.org) as a sort of "external agent", where you format and pass the ID Token via POST to the endpoint, and get back a response code which indicates whether the request is allowed or denied. This way, advanced users could write their own policies.
Again, I'm just thinking aloud...

@maraino 
I'm sorry for the delay, but I need to ask my boss for the CLA. I'll get back as soon as possible.

mmalone:
> I do agree, it's indeed difficult to think of a situation where a provider would express something different than group membership in a custom claim named groups :) Nevertheless, it is a custom claim, so it's up to you guys to decide if it's worth including it.

I appreciate your understanding. We've decided this is a good idea though. We're gonna merge. It just makes sense. We weren't aware that the use of `group` was fairly commonplace. Thank you for bringing this to our attention and thank you for contributing a solution!

> Maybe you could consider integrating something like OPA ... This way, advanced users could write their own policies.

Yeeeaaap! OPA integration could be interesting. We have some other ideas, too. Are you using OPA for anything currently? Curious: would OPA be your preference or was it just the first thing that you thought of?

matteo-s:
> Are you using OPA for anything currently? 

We are evaluating OPA and it looks like a great project, but honestly it is not the simplest solution. Their language for policies is powerful but complex, and currently 90% of our use cases for HTTP API are solved via simple sidecar proxies checking JWT claims against mostly static rules.

> Curious: would OPA be your preference or was it just the first thing that you thought of?

For us, the biggest selling point for OPA is the idea of having a single platform for declaring and evaluating (and eventually audit) complex policies for a variety of services and systems.

As of today, it's not my first choice when I deploy something, but looks like it's gaining momentum.
If the requirements demand complex or mutable policies, OPA would be my preference against building a custom system or against learning (and maintain) another policy system.

sourishkrout:
Hi @matteo-s,

I wanted to check in and see if you had a chance to clear/sign the Contributor License Agreements (which is essentially a copy of Apache's https://www.apache.org/licenses/icla.pdf). We really would like to move forward and merge your PR which is waiting for the accepted CLA. Happy to hop on a call to answer questions if that helps. Please let me know.

Thank you!

matteo-s:
Hello @sourishkrout 
I have now signed the CLA. Sorry for the long delay, I had to wait for an approval from my supervisors, which took longer than expected due to holidays.

maraino:
Thanks @matteo-s

PR #85, opened 2019-07-24
SSH CA support

maraino:
### Description
This PR adds support for signing SSH certificates.


api/api_test.go
```diff
@@ -501,6 +502,8 @@ type mockAuthority struct {
 	getTLSOptions                func() *tlsutil.TLSOptions
 	root                         func(shasum string) (*x509.Certificate, error)
 	sign                         func(cr *x509.CertificateRequest, opts provisioner.Options, signOpts ...provisioner.SignOption) (*x509.Certificate, *x509.Certificate, error)
+	singSSH                      func(key ssh.PublicKey, opts provisioner.SSHOptions, signOpts ...provisioner.SignOption) (*ssh.Certificate, error)
```
dopey: sing -> sign for both

maraino: Can you go ahead and fix the issues, I'm right now on a🛩

api/ssh.go
```diff
@@ -0,0 +1,159 @@
...
+	OTT              string       `json:"ott"`
+	CertType         string       `json:"certType,omitempty"`
+	Principals       []string     `json:"principals,omitempty"`
+	ValidAfter       TimeDuration `json:"validAfter,omitempty"`
+	ValidBefore      TimeDuration `json:"validBefore,omitempty"`
+	AddUserPublicKey []byte       `json:"addUserPublicKey,omitempty"`
+}
+
+// SignSSHResponse is the response object that returns the SSH certificate.
+type SignSSHResponse struct {
```
dopey: We use 'crt' and 'cert' and numerous different places (for shortening Certificate). Let's choose one and be consistent. curl and openssl were using `cert` for their flags and docs so I started doing that.

api/ssh.go
```diff
@@ -0,0 +1,159 @@
...
+	Certificate        SSHCertificate  `json:"crt"`
+	AddUserCertificate *SSHCertificate `json:"addUserCrt,omitempty"`
+}
+
+// SSHCertificate represents the response SSH certificate.
+type SSHCertificate struct {
+	*ssh.Certificate `json:"omitempty"`
+}
+
+// MarshalJSON implements the json.Marshaler interface. The certificate is
```
dopey: A marshaled SSH certificate should be PEM encoded quoted string.

dopey: hmm, we're also base64 encoding, so add that to the documentation.

api/ssh.go
```diff
@@ -0,0 +1,159 @@
...
+// quoted string using the PEM encoding.
+func (c SSHCertificate) MarshalJSON() ([]byte, error) {
+	if c.Certificate == nil {
+		return []byte("null"), nil
+	}
+	s := base64.StdEncoding.EncodeToString(c.Certificate.Marshal())
+	return []byte(`"` + s + `"`), nil
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface. The certificate is
```
dopey: Add to the documentation that it is also base64 encoded (need to decode)

api/ssh.go
```diff
@@ -0,0 +1,159 @@
...
+	if s == "" {
+		c.Certificate = nil
+		return nil
+	}
+	certData, err := base64.StdEncoding.DecodeString(s)
+	if err != nil {
+		return errors.Wrap(err, "error decoding certificate")
+	}
+	pub, err := ssh.ParsePublicKey(certData)
+	if err != nil {
```
dopey: error parsing certificate

api/ssh.go
```diff
@@ -0,0 +1,159 @@
...
+	ctx := provisioner.NewContextWithMethod(context.Background(), provisioner.SignSSHMethod)
+	signOpts, err := h.Authority.Authorize(ctx, body.OTT)
+	if err != nil {
+		WriteError(w, Unauthorized(err))
+		return
+	}
+
+	cert, err := h.Authority.SignSSH(publicKey, opts, signOpts...)
+	if err != nil {
+		WriteError(w, Forbidden(err))
```
dopey: Why Forbidden and not InternalServerError? shouldn't that be the catch all? If it's forbidden then the SignSSH should return a forbidden error. If it doesn't return a known API error then it should be an InternalServerError, right?

authority/authority.go
```diff
@@ -117,6 +121,22 @@ func (a *Authority) init() error {
 		}
 	}
 
+	// Decrypt and load SSH keys
+	if a.config.SSH != nil {
+		if a.config.SSH.HostKey != "" {
+			a.sshCAHostCertSignKey, err = parseCryptoSigner(a.config.SSH.HostKey, a.config.Password)
```
dopey: Should this be the same password as everything else? I assume this is the same password as the intermediate private key uses.

authority/authorize.go
```diff
@@ -72,33 +73,48 @@ func (a *Authority) authorizeToken(ott string) (provisioner.Interface, error) {
...
+	}
 }
 
-// AuthorizeSign authorizes a signature request by validating and authenticating
-// a OTT that must be sent w/ the request.
-func (a *Authority) AuthorizeSign(ott string) ([]provisioner.SignOption, error) {
-	var errContext = context{"ott": ott}
-
+// authorizeSign loads the provisioner from the token, checks that it has not
+// been used again and calls the provisioner AuthorizeSign method. returns a
```
dopey: capitalization

authority/authorize.go
```diff
@@ -72,33 +73,48 @@ func (a *Authority) authorizeToken(ott string) (provisioner.Interface, error) {
...
-	opts, err := p.AuthorizeSign(ott)
+	opts, err := p.AuthorizeSign(ctx, ott)
 	if err != nil {
 		return nil, &apiError{errors.Wrap(err, "authorizeSign"), http.StatusUnauthorized, errContext}
 	}
-
 	return opts, nil
 }
 
+// AuthorizeSign authorizes a signature request by validating and authenticating
```
dopey: I assume this method is planned for deprecation? We should make note of that in the documentation, so that we (and all our users) make sure to use `Authorize` instead.

authority/provisioner/sign_ssh_options.go
```diff
@@ -0,0 +1,306 @@
...
+// to modify the SSH certificate.
+type SSHCertificateOptionModifier interface {
+	SignOption
+	Option(o SSHOptions) SSHCertificateModifier
+}
+
+// SSHCertificateValidator is the interface used to validate an SSH certificate.
+type SSHCertificateValidator interface {
+	SignOption
+	Valid(crt *ssh.Certificate) error
```
dopey: consistency w/ cert

api/ssh.go
maraino: @dopey SSH certificates are not in PEM format

authority/provisioner/sign_ssh_options.go
```diff
@@ -0,0 +1,306 @@
...
+	return nil
+}
+
+// sshDefaultExtensionModifier implements an SSHCertificateModifier that sets
+// the default extensions in an SSH certificate.
+type sshDefaultExtensionModifier struct{}
+
+func (m *sshDefaultExtensionModifier) Modify(cert *ssh.Certificate) error {
+	switch cert.CertType {
+	// Default to no extensions to HostCert
```
dopey: what?

api/ssh.go
maraino: I used the same logic we are using for X509 certificates

authority/authority.go
maraino: That's the current functionality for MVP, but we can change it if we need to.

authority/provisioner/sign_ssh_options.go
dopey: oh, "for" HostCert

authority/ssh.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	return cert, nil
+}
+
+// SignSSHAddUser signs a certificate that provisions a new user in a server.
+func (a *Authority) SignSSHAddUser(key ssh.PublicKey, subject *ssh.Certificate) (*ssh.Certificate, error) {
+	if a.sshCAUserCertSignKey == nil {
+		return nil, &apiError{
+			err:  errors.New("signSSHProxy: user certificate signing is not enabled"),
```
dopey: signSSHAddUser in error msg?

authority/ssh.go
```diff
@@ -0,0 +1,239 @@
...
+			code: http.StatusForbidden,
+		}
+	}
+
+	nonce, err := randutil.ASCII(32)
+	if err != nil {
+		return nil, &apiError{err: err, code: http.StatusInternalServerError}
+	}
+
+	var serial uint64
```
dopey: pull this out into a function. maybe in cli, if we plan on having some ssh cert utils there.

maraino: We can add that to randutil, but there's no need to do it right now

api/ssh.go
dopey: ok, but do you think it should be InternalServerError? I think that should be the catch all.

CMCDragonkai:
Why does using ssh-ca result in creating 2 keypairs one called host and one called user? Don't we only need 1 keypair to represent the CA? What's the 2 keypairs used for?

maraino:
@CMCDragonkai: one key is used to sign host certificates, the other one is used to sign user certificates. And for you information, the upcoming version of step certificates will come with a more complete support for SSH certificates.

dopey:
Note that in the `ssh` branch (soon to be merged to master) we've actually renamed these files to `ssh_user_ca_key` and `ssh_host_ca_key`. @CMCDragonkai do you feel these names help to clarify the purpose of these files?

CMCDragonkai:
Is it not possible to use 1 key to sign both? Similar to PKI? I may be misunderstanding something?

On 14 January 2020 05:11:47 GMT+11:00, Mariano Cano <notifications@github.com> wrote:
>@CMCDragonkai: one key is used to sign host certificates, the other one
>is used to sign user certificates. And for you information, the
>upcoming version of step certificates will come with a more complete
>support for SSH certificates.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/smallstep/certificates/pull/85#issuecomment-573795907

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

dopey:
You can. But it's good hygiene to use different keys for different processes. The process of obtaining a user SSH certificate is likely different than the one for getting a host SSH certificate. Hence, we thought the keys should be different as well. 

We're hoping to help with some of the confusion here in the upcoming release. Ideally, as a user, you would never have to deal with these keys directly.

PR #96, opened 2019-08-27
Enforce >= 2048 bit rsa keys in CSRs

### LINKED ISSUE #94: Enforce that CSR RSA public keys have >= 2048 bits

dopey:
Fixes #94


api/api.go
```diff
@@ -200,6 +200,12 @@ func (s *SignRequest) Validate() error {
 	if err := s.CsrPEM.CertificateRequest.CheckSignature(); err != nil {
 		return BadRequest(errors.Wrap(err, "invalid csr"))
 	}
+	switch k := s.CsrPEM.CertificateRequest.PublicKey.(type) {
+	case *rsa.PublicKey:
+		if k.Size() < 256 {
+			return BadRequest(errors.New("rsa key in CSR must be at least 2048 bits (256 bytes)"))
+		}
+	}
```
maraino: This should be enforced in the authority, not in the api, in case at some point we want to make it configurable.

PR #97, opened 2019-08-28
Skip unsupported provisioners.

maraino:
### Description
Compiled clients like `step` will fail if we add a new provisioner type and the version `step` is compiled does not supports it. This PR changes the behavior and skips unknown provisioners instead of failing.

The main 'problem' of this change is that step-ca is not able to detect invalid provisioners, this can also be a feature as the same configuration can be used by different versions.

### DIFF
--- a/authority/provisioner/provisioner.go
+++ b/authority/provisioner/provisioner.go
@@ -150,7 +150,11 @@ func (l *List) UnmarshalJSON(data []byte) error {
 		case "azure":
 			p = &Azure{}
 		default:
-			return errors.Errorf("provisioner type %s not supported", typ.Type)
+			// Skip unsupported files. When this method is specially used on
+			// clients, these might be compiled with a version that does not
+			// support a specific provisioner type, if we don't skip it we will
+			// force to re-compile the client to make it compatible.
+			continue
 		}
 		if err := json.Unmarshal(data, p); err != nil {
 			return errors.Errorf("error unmarshaling provisioner")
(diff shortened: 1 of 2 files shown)


dopey:
I'm not sure if we should be more prescriptive about how to fix the issue in the WARN statement.

authority/provisioner/provisioner.go
```diff
@@ -150,7 +151,16 @@ func (l *List) UnmarshalJSON(data []byte) error {
...
-			return errors.Errorf("provisioner type %s not supported", typ.Type)
+			// Skip unsupported provisioners. A client using this method may be
+			// compiled with a version of smallstep/certificates that does not
+			// support a specific provisioner type. If we don't skip unknown
+			// provisioners, a client encountering an unknown provisioner will
+			// break. Rather than break the client, we skip the provisioner but
+			// warn the user that an unknown provisioner was found and suggest
+			// that the user update their client's dependency on
+			// step/certificates and recompile.
+			log.Printf("[WARN] Unknown provisioner of type '%s' found. Please update your client.", typ.Type)
```
maraino: In step-sds we do a provisioner request to get a JWK provisioner, and we don't care about the others, so, an error like the following one doesn't make much sense:
```
2019/08/29 12:55:23 [WARN] Unknown provisioner of type 'GCP' found. Please update your client.
```
And it's also a server, but is acting as a client at this point. It makes sense for `step-ca` and might make sense for `step`, so having some kind of pluggable logger might be better.

PR #114, opened 2019-09-25
Add x5c provisioner capabilities

dopey:
CLI PR: https://github.com/smallstep/cli/pull/162


authority/provisioner/x5c.go
```diff
@@ -0,0 +1,264 @@
...
+
+// AuthorizeSign validates the given token.
+func (p *X5C) AuthorizeSign(ctx context.Context, token string) ([]SignOption, error) {
+	claims, err := p.authorizeToken(token, p.audiences.Sign)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check for SSH token
+	if claims.Step != nil && claims.Step.SSH != nil {
```
maraino: I probably this on JWK because at that point I didn't have `MethodFromContext(context.Context)` implemented. Can you change this here and in JWK?

maraino:
```go
if m := MethodFromContext(ctx); m == SignSSHMethod {
   ...
}
```

maraino: We might need to check that those fields are not nil later.

authority/provisioner/x5c.go
```diff
@@ -0,0 +1,264 @@
...
+		Roots: p.rootPool,
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "error verifying x5c certificate chain")
+	}
+	leaf := verifiedChains[0][0]
+
+	if leaf.KeyUsage&x509.KeyUsageDigitalSignature == 0 {
+		return nil, errors.New("certificate used to sign x5c token cannot be used for digital signature")
+	}
```
maraino: Defining `KeyUsages` in `x509.VerifyOptions` does this for you, and I think it checks KeyUsage in the full chain not just in the leaf.

dopey: Just checked and DigitalSignature is not an extended key usage. The [`KeyUsage`](https://godoc.org/crypto/x509#VerifyOptions) param appears to only accept [ExtKeyUsage](https://godoc.org/crypto/x509#ExtKeyUsage).

maraino: Sounds good. But should the full chain support this key usage or just the leaf?

dopey: Just as a point of reference, our root and intermediate certificates do not support digital signature extension. 

I think probably just the leaf needs to support it. @mmalone thoughts?

authority/provisioner/sign_ssh_options.go
```diff
@@ -215,6 +225,29 @@ func (m *sshCertificateValidityModifier) Modify(cert *ssh.Certificate) error {
...
+			max = rem
+		}
+		// If the remaining duration from the provisioning credential is less than
+		// the default duration for the requested type of SSH certificate then we
+		// reset our default duration.
+		if rem < d {
+			d = rem
+		}
+	}
+
```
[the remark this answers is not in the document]
dopey: Are you suggesting here to have one SignOption that modifies the NofAfter & NotBefore, and another one that validates them?

authority/provisioner/sign_ssh_options.go
```diff
@@ -223,15 +221,49 @@ func (m *sshCertificateValidityModifier) Modify(cert *ssh.Certificate) error {
...
+	}
+	dur, err := time.ParseDuration(fmt.Sprintf("%ds", cert.ValidBefore-cert.ValidAfter))
+	if err != nil {
+		return errors.Wrapf(err, "error converting ssh certificate duration; "+
+			"vaf: %d, vbf: %d, d: %d", cert.ValidAfter, cert.ValidBefore, cert.ValidBefore-cert.ValidAfter)
+	}
+	if m.rem < dur {
+		t := time.Unix(int64(cert.ValidAfter), 0)
+		cert.ValidBefore = uint64(t.Add(m.rem).Unix())
+	}
```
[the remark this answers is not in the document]
maraino: I think we can do like in the `profileProvCredDuration` where you apply the `profileDefaultDuration` of the lower value. But later you need to check that the duration is less than m.rem too in case that was the user who set it.

authority/provisioner/sign_options.go
```diff
@@ -197,7 +184,34 @@ func (v emailAddressesValidator) Valid(req *x509.CertificateRequest) error {
...
+type profileProvCredDuration struct {
+	def time.Duration
+	rem time.Duration
+}
+
+func (v profileProvCredDuration) Option(so Options) x509util.WithOption {
+	if v.rem < v.def {
+		v.def = v.rem
+	}
+	return profileDefaultDuration(v.def).Option(so)
```
maraino: If the user was the one who set the duration you're not checking that is lower than v.rem and returning an error.

authority/provisioner/x5c.go
maraino: sounds good

PR #124, opened 2019-10-08
Create ACME database tables when initializing ACME autority.

dopey:
@maraino Thoughts on making this change? Didn't realize that `-`s were special / illegal characters in table names. This change won't break existing operation, but people using the badger DB will have a bunch of "tables" (really just keys) that won't be used anymore.

### DIFF
--- a/acme/authority.go
+++ b/acme/authority.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/smallstep/certificates/authority/provisioner"
+	database "github.com/smallstep/certificates/db"
 	"github.com/smallstep/cli/jose"
 	"github.com/smallstep/nosql"
 )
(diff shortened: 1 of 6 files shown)


acme/authority.go
```diff
@@ -43,11 +44,35 @@ type Authority struct {
 	signAuth SignAuthority
 }
 
+var (
+	accountTable           = []byte("acme_accounts")
+	accountByKeyIDTable    = []byte("acme_keyID_accountID_index")
+	authzTable             = []byte("acme_authzs")
+	challengeTable         = []byte("acme_challenges")
+	nonceTable             = []byte("nonces")
```
maraino: Why not acme_nonces?

acme/common.go
```diff
@@ -22,17 +22,6 @@ type Identifier struct {
 	Value string `json:"value"`
 }
 
-var (
-	accountTable           = []byte("acme-accounts")
-	accountByKeyIDTable    = []byte("acme-keyID-accountID-index")
-	authzTable             = []byte("acme-authzs")
-	challengeTable         = []byte("acme-challenges")
-	nonceTable             = []byte("nonce-table")
```
maraino: You've renamed it to nonces.

maraino: And you're using `_` instead of `-`

acme/authority.go
dopey: I figured we might use `nonces` in other places as well ...

acme/common.go
dopey: Yeah, so using `_` instead of `-` because that's turns out to be a sql special character. All the original tables I created (for certs, revocations, otts) used `_`s. I updated nosql to use backquote all table names anyway (so illegal chars shouldn't matter), but I'd rather not use them when possible, and in this case it's more consistent with the original table names anyway.

acme/authority.go
maraino: ok

acme/common.go
maraino: ok

PR #132, opened 2019-10-24
Fix Debian package link and version in README

rmedaer:
Following the README procedure to install the Debian packages, I figured out that the links were broken.


dopey:
Hey @rmedaer, great catch! Thanks for fixing. 🙏

PR #133, opened 2019-10-29
kubernetes service account provisioner

dopey:


authority/provisioner/k8sSA.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/pkg/errors"
+	"github.com/smallstep/cli/crypto/pemutil"
+	"github.com/smallstep/cli/jose"
+	"golang.org/x/crypto/ed25519"
+)
+
+var (
+	// K8sSAName is the default name used for kubernetes service account tokens,
+	// because there can only be one per instance of step-ca.
+	K8sSAName   = "k8sSA-default"
```
maraino: I don't see anything restricting this, one k8sSA provisioner per step-ca. Is this a restriction because there's no way to distinguish tokens from different?

If we can only have one, we should do some kind of validation.

authority/provisioner/collection.go
```diff
@@ -91,10 +92,16 @@ func (c *Collection) LoadByToken(token *jose.JSONWebToken, claims *jose.Claims)
 	if err := token.UnsafeClaimsWithoutVerification(&payload); err != nil {
 		return nil, false
 	}
-	// Audience is required
-	if len(payload.Audience) == 0 {
+	if len(payload.ServiceAccountName) > 0 {
+		if p, ok := c.Load("k8ssa/" + K8sSAName); ok {
```
maraino: Looks like we are loading the provisioner `k8ssa/k8sSA-default` but storing `k8ssa/<provisioner.Name>`.

authority/provisioner/k8sSA.go
```diff
@@ -0,0 +1,245 @@
...
+type K8sSA struct {
+	Type      string  `json:"type"`
+	Name      string  `json:"name"`
+	Claims    *Claims `json:"claims,omitempty"`
+	PubKeys   []byte  `json:"pubKeys,omitempty"`
+	claimer   *Claimer
+	audiences Audiences
+	//kauthn    kauthn.AuthenticationV1Interface
+	pubKeys []interface{}
+}
```
maraino: Does it make sense to optionally restrict the use of the provisioner to specific service accounts?

authority/provisioner/k8sSA.go
```diff
@@ -0,0 +1,245 @@
...
+func (p *K8sSA) AuthorizeSign(ctx context.Context, token string) ([]SignOption, error) {
+	_, err := p.authorizeToken(token, p.audiences.Sign)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check for SSH sign-ing request.
+	if MethodFromContext(ctx) == SignSSHMethod {
+		return nil, errors.New("ssh certificates not enabled for k8s ServiceAccount provisioners")
+	}
```
maraino: I suppose there's no point in using SSH because you should use `kubectl exec`, right?

authority/provisioner/k8sSA.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		return err
+	}
+	if !rvw.Status.Allowed {
+		return fmt.Errorf("Unable to create kubernetes token reviews: %s", rvw.Status.Reason)
+	}
+
+	return nil
+}
+*/
```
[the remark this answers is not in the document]
dopey: yea, so this code is sort of a guide for how we would integrate against the kubernetes API server for V2. Instead of asking users to supply the pubKeys when the provisioner is added, you would have configuration details for connection to the kubernetes API server. Then you would submit the token token to the API and it would return valid/invalid. But for now we're doing something more basic. I can leave a comment to this effect in the code.

authority/provisioner/k8sSA.go
dopey: yep, we can only have one (for now), because the k8s Service Account tokens don't have enough info in them to easily differentiate between kubernetes clusters. The cli restricts the name of the provisioner by always using this name. I can do some kind of validation in the init.

authority/provisioner/collection.go
dopey: Yea, so the provisioner.name get's hardcoded from the cli. I can leave some documentation in `provisioner add --type k8ssa` to that effect. And also do some validation in the provisioner.init to prevent the CA from starting up if the name doesn't match the default.

authority/provisioner/k8sSA.go
dopey: Might be something for @mmalone  to weigh in on.

authority/provisioner/collection.go
dopey: I've changed this so we are overwriting the provisioner name with the hardcoded one. And added validation in the CA to fail startup if a K8sSA provisioner has the wrong name (or if there's more than one such provisioner).

authority/provisioner/k8sSA.go
dopey: That was my thinking, but might be something to have @mmalone weigh in on as well.

authority/provisioner/k8sSA.go
mmalone: Sure, if it's _really_ easy. Otherwise open a ticket for it and let's do it later when we've got a little less schedule pressure.

authority/provisioner/k8sSA.go
mmalone: I think this is fine for now. I can't think of a k8s/SSH use case. If someone comes along with one we can reconsider.

authority/provisioner/k8sSA.go
dopey: I'll open a ticket.  We're gonna need to revisit to implement the TokenReview API anyway.

authority/provisioner/k8sSA.go
```diff
@@ -0,0 +1,263 @@
...
+func (p *K8sSA) Init(config Config) (err error) {
+	switch {
+	case p.Type == "":
+		return errors.New("provisioner type cannot be empty")
+	case p.Name == "":
+		return errors.New("provisioner name cannot be empty")
+	// Verify that any kubernetes service account provisioner has the standard,
+	// default name.
+	case p.Name != K8sSAName:
+		return errors.Errorf("k8sSA provisioner name must be %s, but got %s", K8sSAName, p.Name)
```
maraino: I would let the user let to use any name, and just use the name on `GetID()` used in `Collection.Store(p Interface)`

authority/provisioner/collection.go
maraino: I would prefer to let the user set a custom name, and just use the fixed `GetID()`, this ID is used in `Collection.Store()`.

authority/provisioner/k8sSA.go
```diff
@@ -0,0 +1,245 @@
...
+	ServiceAccountUID  string `json:"kubernetes.io/serviceaccount/service-account.uid,omitempty"`
+}
+
+// K8sSA represents a Kubernetes ServiceAccount provisioner; an
+// entity trusted to make signature requests.
+type K8sSA struct {
+	Type      string  `json:"type"`
+	Name      string  `json:"name"`
+	Claims    *Claims `json:"claims,omitempty"`
+	PubKeys   []byte  `json:"pubKeys,omitempty"`
```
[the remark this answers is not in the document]
maraino: Are you going to change this?

authority/provisioner/k8sSA.go
maraino: Sounds good.

authority/provisioner/k8sSA.go
maraino: Sounds good.

authority/provisioner/collection.go
maraino: And by the way, `LoadByCertificate` does not have support for K8sSA.

authority/provisioner/k8sSA.go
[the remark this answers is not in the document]
dopey: yea, my bad. it was staged  and not pushed.

PR #145, opened 2019-12-17
wip

dopey:


errs/error.go
```diff
@@ -0,0 +1,197 @@
...
+	return json.Marshal(&ErrorResponse{Status: e.Status, Message: http.StatusText(e.Status)})
+}
+
+// Format implements the fmt.Formatter interface.
+func (e *Error) Format(f fmt.State, c rune) {
+	if err, ok := e.Err.(fmt.Formatter); ok {
+		err.Format(f, c)
+		return
+	}
+	fmt.Println(e.Err.Error())
```
maraino: You need to write in the `fmt.State` no in the standard output.

errs/error.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// WithMessage sets the user friendly message for a given error.
+func WithMessage(e error, m string) error {
+	if err, ok := e.(*Error); ok {
+		err.Msg = m
+		return err
+	}
+	return &Error{
+		Status: http.StatusInternalServerError,
```
maraino: We should change if it implements the `StatusCoder` interface first

errs/error.go
```diff
@@ -0,0 +1,197 @@
...
+		return err
+	}
+	e.Status = er.Status
+	e.Err = fmt.Errorf(er.Message)
+	return nil
+}
+
+// NewError returns a new Error. If the given error implements the StatusCoder
+// interface we will ignore the given status.
+func NewError(status int, err error) error {
```
maraino: Perhaps we can add a variadic parameter that allows setting the Message, or create a new method that allows that, something like `NewErrorMsg(status, err, msg)`

errs/error.go
```diff
@@ -0,0 +1,197 @@
...
+// StackTracer must be by those errors that return an stack trace.
+type StackTracer interface {
+	StackTrace() errors.StackTrace
+}
+
+// Error represents the CA API errors.
+type Error struct {
+	Status int
+	Err    error
+	Msg    string
```
maraino: Message instead of Msg

errs/error.go
maraino: Adding here proposal discussed over slack:
```go
type Option func(e *Error) // or type Option func(e *Error) error

func New(status int, err error, opts ...Option) {}
func Wrap(status int, err error, msg string, opts ...Options) {}
// The args before the first option will be the arguments for the format string.
func Wrapf(status int, err error, format string, args ...interface{}) {}
```

errs/error.go
dopey: If I change this to Message then we'd need to rename the method `Message`, implements the friendlyerror type.

maraino: 👍

api/ssh.go
```diff
@@ -424,17 +425,17 @@ func (h *caHandler) SSHConfig(w http.ResponseWriter, r *http.Request) {
 func (h *caHandler) SSHCheckHost(w http.ResponseWriter, r *http.Request) {
 	var body SSHCheckPrincipalRequest
 	if err := ReadJSON(r.Body, &body); err != nil {
-		WriteError(w, BadRequest(errors.Wrap(err, "error reading request body")))
+		WriteError(w, errs.Wrap(http.StatusBadRequest, err, "error reading request body"))
```
maraino: Looks ok, but most of the similar errors are written like:
```go
WriteError(w, errs.BadRequest(errors.Wrap(err, "error reading request body")))
```

errs/error.go
```diff
@@ -0,0 +1,249 @@
...
+// ErrorResponse represents an error in JSON format.
+type ErrorResponse struct {
+	Status  int    `json:"status"`
+	Message string `json:"message"`
+}
+
+// Cause implements the errors.Causer interface and returns the original error.
+func (e *Error) Cause() error {
+	return e.Err
+}
```
maraino: Not required right now, but from Go 1.13, the new way to do it is using an `Unwrap() error`, this is not supported by `pkg/errors`, we could just have both `Cause` and `Unwrap`.

errs/error.go
```diff
@@ -0,0 +1,249 @@
...
+	if e == nil {
+		return nil
+	}
+	if err, ok := e.(*Error); ok {
+		err.Err = errors.Wrap(err.Err, m)
+		e = err
+	} else {
+		e = errors.Wrap(e, m)
+	}
+	return StatusCodeError(status, e, opts...)
```
maraino: Passing the options to `StatusCodeError` that does not support all the status codes looks weird, especially if we want to add a new one. I think you should just create the Error here.

errs/error.go
```diff
@@ -0,0 +1,249 @@
...
+			break
+		}
+	}
+	if err, ok := e.(*Error); ok {
+		err.Err = errors.Wrapf(err.Err, format, args...)
+		e = err
+	} else {
+		e = errors.Wrapf(e, format, args...)
+	}
+	return StatusCodeError(status, e, opts...)
```
maraino: See the previous comment

api/ssh.go
dopey: So, the trouble here is that we probably want to use our Wrap method which has a func type of `func(status int, err error, ....)` so we'd be writing `BadRequest(errs.Wrap(http.StatusBadRequest, err, ...`), repeating the bad request part.

PR #146, opened 2019-12-20
Perform domain normalization for wildcard domains

anxolerd:
### Description

Perform domain normalization for wildcard domains, so we do query
TXT records for _acme-challenge.example.domain instead of
_acme-challenge.*.example.domain when performing DNS-01 challenge. In
this way the behavior is consistent with letsencrypt and records queried
are in sync with the ones that are shown in certbot manual mode.

### DIFF
--- a/acme/challenge.go
+++ b/acme/challenge.go
@@ -385,11 +385,21 @@ func (dc *dns01Challenge) validate(db nosql.DB, jwk *jose.JSONWebKey, vo validat
 		return dc, nil
 	}
 
-	txtRecords, err := vo.lookupTxt("_acme-challenge." + dc.Value)
+	// Normalize domain for wildcard DNS names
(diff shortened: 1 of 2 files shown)


anxolerd:
@dopey, added the testcase as requested, however, I am not sure it is as small and simple as should/can be. If there is anything I can remove to make the purpose of testcase clearer, please, let me know.

dopey:
Looks great! I forgot how complicated I made those tests 😬. I'm going to go ahead and merge and then make one small change. 

```
domain := strings.TrimPrefix(dc.Value, "*.")
```

`TrimPrefix` already checks for `HasPrefix`: 

```func TrimPrefix(s, prefix string) string {
  	if HasPrefix(s, prefix) {
  		return s[len(prefix):]
  	}
  	return s
}
```

Thanks for the contribution!

PR #150, opened 2020-01-04
Backdate

maraino:
### Description

This PR changes the x509 and SSH certificates to the NotBefore or ValidAfter are one minute before the current time.

### DIFF
--- a/authority/config.go
+++ b/authority/config.go
@@ -28,6 +28,7 @@ var (
 		MaxVersion:    1.2,
 		Renegotiation: false,
 	}
+	defaultBackdate         = time.Minute
 	defaultDisableRenewal   = false
 	defaultEnableSSHCA      = false
 	globalProvisionerClaims = provisioner.Claims{
@@ -65,10 +66,11 @@ type Config struct {
 
 // AuthConfig represents the configuration options for the authority.
 type AuthConfig struct {
-	Provisioners         provisioner.List    `json:"provisioners"`
-	Template             *x509util.ASN1DN    `json:"template,omitempty"`
-	Claims               *provisioner.Claims `json:"claims,omitempty"`
-	DisableIssuedAtCheck bool                `json:"disableIssuedAtCheck,omitempty"`
+	Provisioners         provisioner.List      `json:"provisioners"`
+	Template             *x509util.ASN1DN      `json:"template,omitempty"`
+	Claims               *provisioner.Claims   `json:"claims,omitempty"`
+	DisableIssuedAtCheck bool                  `json:"disableIssuedAtCheck,omitempty"`
+	Backdate             *provisioner.Duration `json:"backdate,omitempty"`
 }
 
 // Validate validates the authority configuration.
@@ -91,6 +93,17 @@ func (c *AuthConfig) Validate(audiences provisioner.Audiences) error {
 	if c.Template == nil {
 		c.Template = &x509util.ASN1DN{}
 	}
+
+	if c.Backdate != nil {
+		if c.Backdate.Duration < 0 {
+			return errors.New("authority.backdate cannot be less than 0")
+		}
+	} else {
+		c.Backdate = &provisioner.Duration{
+			Duration: defaultBackdate,
+		}
+	}
+
 	return nil
 }
 
--- a/authority/provisioner/azure.go
+++ b/authority/provisioner/azure.go
@@ -209,7 +209,7 @@ func (p *Azure) Init(config Config) (err error) {
 	return nil
 }
 
-// parseToken returuns the claims, name, group, error.
+// parseToken returns the claims, name, group, error.
 func (p *Azure) parseToken(token string) (*azurePayload, string, string, error) {
 	jwt, err := jose.ParseSigned(token)
 	if err != nil {
@@ -335,7 +335,7 @@ func (p *Azure) AuthorizeSSHSign(ctx context.Context, token string) ([]SignOptio
 		// Set the default extensions.
 		&sshDefaultExtensionModifier{},
 		// Set the validity bounds if not set.
-		sshDefaultValidityModifier(p.claimer),
+		&sshDefaultDuration{p.claimer},
 		// Validate public key
 		&sshDefaultPublicKeyValidator{},
 		// Validate the validity period.
--- a/authority/provisioner/sign_options.go
+++ b/authority/provisioner/sign_options.go
@@ -15,10 +15,12 @@ import (
 	"golang.org/x/crypto/ed25519"
 )
 
-// Options contains the options that can be passed to the Sign method.
+// Options contains the options that can be passed to the Sign method. Backdate
+// is automatically filled and can only be configured in the CA.
 type Options struct {
-	NotAfter  TimeDuration `json:"notAfter"`
-	NotBefore TimeDuration `json:"notBefore"`
+	NotAfter  TimeDuration  `json:"notAfter"`
+	NotBefore TimeDuration  `json:"notBefore"`
+	Backdate  time.Duration `json:"-"`
 }
 
 // SignOption is the interface used to collect all extra options used in the
@@ -189,12 +191,22 @@ func (v emailAddressesValidator) Valid(req *x509.CertificateRequest) error {
 type profileDefaultDuration time.Duration
 
 func (v profileDefaultDuration) Option(so Options) x509util.WithOption {
+	var backdate time.Duration
 	notBefore := so.NotBefore.Time()
 	if notBefore.IsZero() {
 		notBefore = time.Now()
+		backdate = -1 * so.Backdate
 	}
 	notAfter := so.NotAfter.RelativeTime(notBefore)
-	return x509util.WithNotBeforeAfterDuration(notBefore, notAfter, time.Duration(v))
+	return func(p x509util.Profile) error {
+		fn := x509util.WithNotBeforeAfterDuration(notBefore, notAfter, time.Duration(v))
+		if err := fn(p); err != nil {
+			return err
+		}
+		crt := p.Subject()
+		crt.NotBefore = crt.NotBefore.Add(backdate)
+		return nil
+	}
 }
 
 // profileLimitDuration is an x509 profile option that modifies an x509 validity
@@ -208,10 +220,12 @@ type profileLimitDuration struct {
 // certificate to one that is superficially imposed.
 func (v profileLimitDuration) Option(so Options) x509util.WithOption {
 	return func(p x509util.Profile) error {
+		var backdate time.Duration
 		n := now()
 		notBefore := so.NotBefore.Time()
 		if notBefore.IsZero() {
 			notBefore = n
+			backdate = -1 * so.Backdate
 		}
 		if notBefore.After(v.notAfter) {
 			return errors.Errorf("provisioning credential expiration (%s) is before "+
@@ -232,7 +246,7 @@ func (v profileLimitDuration) Option(so Options) x509util.WithOption {
 			}
 		}
 		crt := p.Subject()
-		crt.NotBefore = notBefore
+		crt.NotBefore = notBefore.Add(backdate)
 		crt.NotAfter = notAfter
 		return nil
 	}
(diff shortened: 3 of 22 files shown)


authority/provisioner/sign_ssh_options.go
```diff
@@ -301,8 +335,15 @@ func (v *sshCertificateValidityValidator) Valid(cert *ssh.Certificate) error {
 		return errors.Errorf("unknown ssh certificate type %d", cert.CertType)
 	}
 
-	// seconds
-	dur := time.Duration(cert.ValidBefore-cert.ValidAfter) * time.Second
+	// To not take into account the backdate, time.Now() will be used to
+	// calculate the duration if ValidAfter is in the past.
```
dcow: Is this code path ever used in cases where the cert was not immediately previously generated for signing?

authority/provisioner/sign_ssh_options.go
```diff
@@ -46,13 +46,22 @@ type SSHCertificateOptionsValidator interface {
 	Valid(got SSHOptions) error
 }
 
+// sshModifierFunc is an adapter to allow the use of ordinary functions as SSH
+// certificate modifiers.
+type sshModifierFunc func(cert *ssh.Certificate) error
+
+func (f sshModifierFunc) Modify(cert *ssh.Certificate) error {
```
dcow: Is this just for readability or is there a more formal reason why adding an alternate way to invoke `f` using its type alias is beneficial?

authority/provisioner/sign_ssh_options.go
```diff
@@ -199,67 +208,92 @@ func (m *sshDefaultExtensionModifier) Modify(cert *ssh.Certificate) error {
...
+		}
 
-func sshDefaultValidityModifier(c *Claimer) SSHCertificateModifier {
-	return &sshValidityModifier{c, time.Time{}}
-}
+		if cert.ValidBefore == 0 {
+			certValidBefore := certValidAfter.Add(d)
+			if m.NotAfter.Before(certValidBefore) {
+				certValidBefore = m.NotAfter
+				println(2, certValidBefore.String())
```
dcow: Do we want this println?

authority/provisioner/sign_ssh_options.go
```diff
@@ -199,67 +208,92 @@ func (m *sshDefaultExtensionModifier) Modify(cert *ssh.Certificate) error {
...
 
-// sshLimitValidityModifier adjusts the duration to
-// min(default, remaining provisioning credential duration).
-// E.g. if the default is 12hrs but the remaining validity of the provisioning
-// credential is only 4hrs, this option will set the value to 4hrs (the min of the two values).
-func sshLimitValidityModifier(c *Claimer, validBefore time.Time) SSHCertificateModifier {
-	return &sshValidityModifier{c, validBefore}
+		// Apply backdate safely
+		if cert.ValidAfter > backdate {
+			cert.ValidAfter -= backdate
```
dcow: Won't this result in a very small cert.ValidAfter since the times are likely to be close together?

authority/provisioner/sign_ssh_options_test.go
```diff
@@ -384,3 +440,120 @@ func Test_sshValidityModifier(t *testing.T) {
...
+	defer fn()
+
+	newClaimer := func(claims *Claims) *Claimer {
+		c, err := NewClaimer(claims, globalProvisionerClaims)
+		if err != nil {
+			t.Fatal(err)
+		}
+		return c
+	}
+	unix := func(d time.Duration) uint64 {
```
dcow: Why did you decide to use `unix` time for the ssh stuff but the golang Time elsewhere?

authority/config.go
```diff
@@ -65,10 +66,11 @@ type Config struct {
...
 type AuthConfig struct {
-	Provisioners         provisioner.List    `json:"provisioners"`
-	Template             *x509util.ASN1DN    `json:"template,omitempty"`
-	Claims               *provisioner.Claims `json:"claims,omitempty"`
-	DisableIssuedAtCheck bool                `json:"disableIssuedAtCheck,omitempty"`
+	Provisioners         provisioner.List      `json:"provisioners"`
+	Template             *x509util.ASN1DN      `json:"template,omitempty"`
+	Claims               *provisioner.Claims   `json:"claims,omitempty"`
+	DisableIssuedAtCheck bool                  `json:"disableIssuedAtCheck,omitempty"`
+	Backdate             *provisioner.Duration `json:"backdate,omitempty"`
```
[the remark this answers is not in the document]
maraino: Yes, this is part of `gofmt`, github has also an option to show the diffs without taking into account the spaces.

authority/provisioner/sign_options.go
```diff
@@ -189,12 +191,22 @@ func (v emailAddressesValidator) Valid(req *x509.CertificateRequest) error {
...
+	var backdate time.Duration
 	notBefore := so.NotBefore.Time()
 	if notBefore.IsZero() {
-		notBefore = time.Now()
+		notBefore = now()
+		backdate = -1 * so.Backdate
 	}
 	notAfter := so.NotAfter.RelativeTime(notBefore)
-	return x509util.WithNotBeforeAfterDuration(notBefore, notAfter, time.Duration(v))
+	return func(p x509util.Profile) error {
```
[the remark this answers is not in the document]
maraino: No, you cannot.

authority/provisioner/azure.go
```diff
@@ -209,7 +209,7 @@ func (p *Azure) Init(config Config) (err error) {
 	return nil
 }
 
-// parseToken returuns the claims, name, group, error.
+// parseToken returns the claims, name, group, error.
```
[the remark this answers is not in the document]
maraino: I'm not gonna spend the time split this commit.

authority/provisioner/sign_ssh_options.go
maraino: All the currently supported modifiers/validators are methods on types. With the backdate I needed to change a couple of methods that used to be `SSHCertificateModifier` to be a `SSHCertificateOptionModifier` so I can pass some extra options:

```go
type SSHCertificateModifier interface {
	SignOption
	Modify(cert *ssh.Certificate) error
}

type SSHCertificateOptionModifier interface {
	SignOption
	Option(o SSHOptions) SSHCertificateModifier
}
```

Adding that type I can easily convert just a function to a value with the `SSHCertificateOptionModifier` type, and I don't really need to add two extra types to implement those methods.

This is also common in Go, a great example is `http.HandlerFunc`, see:
https://golang.org/pkg/net/http/#HandlerFunc

authority/provisioner/sign_ssh_options.go
maraino: No, fixed with 77af30bfa395af5a025bb045315359c6fad7977c

authority/provisioner/sign_ssh_options.go
maraino: No, ValidAfter is the number of seconds since the Unix "epoch" and backdate will be a few seconds, 60 by default. The condition above is to avoid overflows if the backdate is set to a very long value.

authority/provisioner/sign_ssh_options.go
maraino: It's always used in your case, but the duration can be defined by the user, this change avoids errors when the certificate duration exceeds the maximum duration because of the backdate.

For example, let's say the maximum is 24h and we request a 24h cert, but because of the backdate the validity will be 24h1m, but it will really be ~24h using the time of the CA, as part of the validity is ~1m ago in the past.

authority/provisioner/sign_ssh_options_test.go
maraino: This is how ssh certificates are defined, and I'm using a library that requires them in that format.

authority/provisioner/utils_test.go
```diff
@@ -31,7 +31,7 @@ var (
 		DisableRenewal:    &defaultDisableRenewal,
 		MinUserSSHDur:     &Duration{Duration: 5 * time.Minute}, // User SSH certs
 		MaxUserSSHDur:     &Duration{Duration: 24 * time.Hour},
-		DefaultUserSSHDur: &Duration{Duration: 4 * time.Hour},
+		DefaultUserSSHDur: &Duration{Duration: 16 * time.Hour},
```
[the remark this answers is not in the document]
maraino: No, I'm just using the current defaults that we changed some time ago.

authority/ssh.go
```diff
@@ -365,9 +368,12 @@ func (a *Authority) RenewSSH(oldCert *ssh.Certificate) (*ssh.Certificate, error)
...
 	}
-	dur := time.Duration(oldCert.ValidBefore-oldCert.ValidAfter) * time.Second
-	va := time.Now()
-	vb := va.Add(dur)
+
+	backdate := a.config.AuthorityConfig.Backdate.Duration
+	duration := time.Duration(oldCert.ValidBefore-oldCert.ValidAfter) * time.Second
+	now := time.Now()
+	va := now.Add(-1 * backdate)
+	vb := now.Add(duration - backdate)
```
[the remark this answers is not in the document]
maraino: To keep the cert duration stable. First, we are adding the backdate to (ValidAfter), getting a value in the past, if we don't subtract it, we will get a duration that exceeds the previous one. 

For example, if the duration variable is `16h1m` (it includes the backdate), and we don't substract it, we will get a new duration of `16h2m` on this renew, `16h3m` on the next one, and so on ...

authority/ssh.go
```diff
@@ -490,9 +496,12 @@ func (a *Authority) RekeySSH(oldCert *ssh.Certificate, pub ssh.PublicKey, signOp
...
 		return nil, errors.New("rekeySSH: cannot rekey certificate without validity period")
 	}
-	dur := time.Duration(oldCert.ValidBefore-oldCert.ValidAfter) * time.Second
-	va := time.Now()
-	vb := va.Add(dur)
+
+	backdate := a.config.AuthorityConfig.Backdate.Duration
+	duration := time.Duration(oldCert.ValidBefore-oldCert.ValidAfter) * time.Second
+	now := time.Now()
+	va := now.Add(-1 * backdate)
```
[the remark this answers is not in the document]
maraino: To keep the duration stable, ... this line is actually adding the backdate.

authority/tls.go
```diff
@@ -136,14 +140,17 @@ func (a *Authority) Renew(oldCert *x509.Certificate) ([]*x509.Certificate, error
...
+	now := time.Now().UTC()
+
 	newCert := &x509.Certificate{
 		PublicKey:                   oldCert.PublicKey,
 		Issuer:                      issIdentity.Crt.Subject,
 		Subject:                     oldCert.Subject,
-		NotBefore:                   now,
-		NotAfter:                    now.Add(duration),
+		NotBefore:                   now.Add(-1 * backdate),
+		NotAfter:                    now.Add(duration - backdate),
```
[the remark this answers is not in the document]
maraino: To keep the duration stable, ... See my previous comment.

authority/config.go
```diff
@@ -91,6 +93,17 @@ func (c *AuthConfig) Validate(audiences provisioner.Audiences) error {
 	if c.Template == nil {
 		c.Template = &x509util.ASN1DN{}
 	}
+
+	if c.Backdate != nil {
+		if c.Backdate.Duration < 0 {
```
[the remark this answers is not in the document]
maraino: I don't see any use case of having backdates in the future, you can already get a certificate in the future if you want to. And I prefer to enforce a format rather than using the absolute value.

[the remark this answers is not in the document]
dcow: fair point

authority/ssh.go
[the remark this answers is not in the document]
dcow: I see. Thanks

authority/provisioner/sign_ssh_options.go
dcow: Oh backdate here is a duration not a date. Got it.

authority/provisioner/sign_ssh_options.go
dcow: I think I get it but I'm sure it will become more clear as I see more go code.

PR #162, opened 2020-01-23
Retry ACME challenge more than once

### LINKED ISSUE #149: ACME http01Challenge randomly not works with k8s after pull new docker image
### Subject of the issue
When acme server challenging acme client (lego) it gets error:
- dial tcp: i/o timeout
- dial tcp 10.106.221.133:80: connect: connection refused

It is caused by ingress when during container startup with updated docker image.

### Your environment - kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:54Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:09:08Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"

### Expected behaviour
Use retry-able http client (https://github.com/hashicorp/go-retryablehttp) for acme http challenge. 

### Actual behaviour
Randomly acme client doesn't obtain certificate for updated container.

### Additional context
When I replaced https://github.com/smallstep/certificates/blob/967e86a48bfc0fd0729c366b33b74ed083f4bcfd/acme/authority.go#L265 by
```go
        import retryablehttp "github.com/hashicorp/go-retryablehttp"
        ...
	client := retryablehttp.NewClient()
	client.HTTPClient = &http.Client{
		Timeout: 20 * time.Second,
	}
	client.RetryWaitMin = 100 * time.Millisecond
	client.RetryWaitMax = 1 * time.Second
	client.RetryMax = 10
```
It is works as is expected.

jkralik:
### Description
http.Client doesn't supports retry during connect. For k8s this
can cause issues like:
- dial tcp: i/o timeout
- dial tcp 10.106.221.133:80: connect: connection refused

### Fixes
#149 

💔Thank you!

### DIFF
--- a/acme/authority.go
+++ b/acme/authority.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"time"
 
+	retryablehttp "github.com/hashicorp/go-retryablehttp"
 	"github.com/pkg/errors"
 	"github.com/smallstep/certificates/authority/provisioner"
 	database "github.com/smallstep/certificates/db"
@@ -262,9 +263,14 @@ func (a *Authority) ValidateChallenge(p provisioner.Interface, accID, chID strin
 	if accID != ch.getAccountID() {
 		return nil, UnauthorizedErr(errors.New("account does not own challenge"))
 	}
-	client := http.Client{
-		Timeout: time.Duration(30 * time.Second),
+	client := retryablehttp.NewClient()
+	client.HTTPClient = &http.Client{
+		Timeout: 20 * time.Second,
 	}
+	client.RetryWaitMin = 100 * time.Millisecond
+	client.RetryWaitMax = 1 * time.Second
+	client.RetryMax = 10
+
 	ch, err = ch.validate(a.db, jwk, validateOptions{
 		httpGet:   client.Get,
 		lookupTxt: net.LookupTXT,
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/go-chi/chi v3.3.4-0.20181024101233-0ebf7795c516+incompatible
 	github.com/go-sql-driver/mysql v1.4.1 // indirect
 	github.com/golangci/golangci-lint v1.18.0 // indirect
+	github.com/hashicorp/go-retryablehttp v0.6.4
 	github.com/juju/ansiterm v0.0.0-20180109212912-720a0952cc2a // indirect
 	github.com/konsorten/go-windows-terminal-sequences v1.0.1 // indirect
 	github.com/lunixbochs/vtclean v0.0.0-20180621232353-2d01aacdc34a // indirect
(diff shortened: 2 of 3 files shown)


maraino:
@dopey: can you take a look to this?

dopey:
Yep. I'd like to get to the bottom of the database issue first since it seems these are related. According to @jkralik the size of the database has stopped endlessly expanding since they made this change, so I definitely want to pull it in.

My only qualm is that we'd be pulling in an external dependency. I haven't had any time to check if that's definitely necessary, or if this behavior is something we can replicate using the standard golang clients.

jkralik:
This fix doesn't fix endlessly expanding DB, but it just try challenge more times. It's fix #149. I think the issue just trigger the bug with endlessly expanding DB.

dcow:
@jkralik I reviewed the [ACME spec](https://tools.ietf.org/html/rfc8555). [Section 8.2](https://tools.ietf.org/html/rfc8555#section-8.2) discusses some mandatory and considerations when implementing challenge validation retries in both the server and the client. Most notably:

> The server MUST provide information about its retry state to the
> client via the "error" field in the challenge and the Retry-After
> HTTP header field in response to requests to the challenge resource.
> The server MUST add an entry to the "error" field in the challenge
> after each failed validation query.  The server SHOULD set the Retry-
> After header field to a time after the server's next validation
> query, since the status of the challenge will not change until that
> time.
>
> Clients can explicitly request a retry by re-sending their response
> to a challenge in a new POST request (with a new nonce, etc.).  This
> allows clients to request a retry when the state has changed (e.g.,
> after firewall rules have been updated).  Servers SHOULD retry a
> request immediately on receiving such a POST request.  In order to
> avoid denial-of-service attacks via client-initiated retries, servers
> SHOULD rate-limit such requests.

The spec mandates that we tie the retry state into our challenge resource so that clients can discern what is happening during the challenge/validation process. This also helps us keep track of retries in the code in any event (as opposed to "fire and forget").

Additionally, while some amount of server retry is allowable to handle exactly your type of scenario (propagation delay for newly provisioned infrastructure), based on my interpretation of language in the section, the responsibility really falls on the client to continue to bug the server until the challenge can be validated. In other words, I think making sure your client only requests the challenge once it knows the infrastructure is ready, and making sure your client re-requests the challenge until it is happy with the outcome.

Thank you for the patch! Since you've likely got this up and running and it works for you, feel free to keep on doing what you're doing. In terms of pulling this into the official CA, we'd like to hold off until we can fully address the requirements laid out in section 8.2 of the spec. If you're interested in revising your patch we're more than happy to work with you to get things in shape. Just say so and we can reopen or you can propose a new patch.

PR #166, opened 2020-01-28
Allow DNS names in CommonName for ACME

maraino:
### Description
The ACME [RFC8555](https://tools.ietf.org/html/rfc8555) allows the CSR with DNS names only in the CommonName part of the subject:

```
The CSR MUST indicate the exact same set of requested identifiers as the initial
newOrder request. Identifiers of type "dns" MUST appear either in the commonName
portion of the requested subject name or in an extensionRequest attribute
[RFC2985] requesting a subjectAltName extension, or both.
```


acme/order.go
```diff
@@ -253,18 +254,30 @@ func (o *order) finalize(db nosql.DB, csr *x509.CertificateRequest, auth SignAut
...
-	if !reflect.DeepEqual(csrNames, orderNames) {
+	orderNames = uniqueLowerNames(orderNames)
+
+	// Validate identifier names against CSR alternative names.
+	if len(csr.DNSNames) != len(orderNames) {
 		return nil, BadCSRErr(errors.Errorf("CSR names do not match identifiers exactly"))
 	}
+	for i := range csr.DNSNames {
+		if csr.DNSNames[i] != orderNames[i] {
+			return nil, BadCSRErr(errors.Errorf("CSR names do not match identifiers exactly"))
```
dopey: Can we print the csr and order names here e.g.

```
BadCSRErr(errors.Errorf("CSR names do not match order identifiers exactly; CSR names = %s, Order names = %s", csr.DNSNames, orderNames))
```

maraino: done.

maraino:
Replaced by https://github.com/smallstep/certificates/pull/167

PR #175, opened 2020-02-04
Its 2020!

Johannestegner:
### Description

Minor update to the LICENSE to increase the year by +1.  
Last commit to file was to update to 2019, so thought I'd keep the tradition going ;)


dopey:
Hey @Johannestegner would you mind writing some tests for this ;p

Thanks! We like this tradition as well. Let's keep it going.

Issue #170, opened 2020-01-29
[Docker] listen tcp 127.0.0.1:443: bind: permission denied


strarsis:
### Subject of the issue
When `step ca` server should bind to the default ACME port 443, it fails in docker:
````
listen tcp 127.0.0.1:443: bind: permission denied
````

### Your environment
* OS: WSL (Bash on Windows)
* Version: Windows 10 Pro x64

### Steps to reproduce
`docker-compose.yml`:
```yml
version: '3.7'

services:
  step-ca:
    image: smallstep/step-ca:0.14.0-rc.2
    # currently needed for step-ca
    tty: true
    command: "step-ca /step/config/ca.json --password-file=/step/secrets/password"
    volumes:
      - "./step:/step"
    labels:
      traefik.enable: true
      traefik.http.routers.traefik.rule: Host(`ca.localhost`)
      traefik.http.routers.traefik.entrypoints: web
      traefik.http.routers.traefik.service: api@internal
    networks:
      - "web"

networks:
  web:
    external: true
````

### Expected behaviour
The `step ca` container should go up and listen on port 443.

### Actual behaviour
The `step ca` container fails to go up, and fails with error:
````
listen tcp 127.0.0.1:443: bind: permission denied
````

### Additional context
This seems to be an usual issue with `go` applications, I found similar issues with `caddy`.

maraino:
Hi @strarsis: this is probably because the step-ca is being run using the `step` user that doesn't have permissions to bind ports below 1024. 
There a few solutions:
* The simpler one would be to switch the port to one over 1024. And optionally a reverse proxy can be set to forward 443 to let's say port 9000.
* A better one is to grant permissions to bind 443, you'll need to run this as a root `setcap CAP_NET_BIND_SERVICE=+eip /path/to/step-ca`
* And finally you can run the service as root.

I'm not sure if it's a good idea to produce the docker images with the setcap enabled, as it's probably the decision of enabling this belongs to the person that will run it. And it's probably easy to create a new Dockerfile, extending the certificates one (`FROM smallstep/step-ca:0.14.0-rc.2`) and running the setcap command as a root user.

strarsis:
Currently I let step-ca bind to a different port and configure traefik to use that custom port instead.

chfl4gs:
You can just create new Dockerfile and build new image.

FROM smallstep/step-ca:latest

USER root

RUN apk add --no-cache libcap \
[...]

PR #176, opened 2020-02-06
Let step-ca bind to ports < 1024

moqmar:
### Description
This PR allows the `step-ca` binary in the Docker container to bind to ports lower than 1024.

### Explanation
I'm trying to use Smallstep in an environment that blocks many outgoing ports and doesn't let me set Docker's `--port` argument directly, so I'd like to run `step-ca` from the Docker container directly on port 443. Unfortunately, this isn't allowed as the daemon is not being started as root, but rather as the `step` user. This commit resolves that issue by adding the `CAP_NET_BIND_SERVICE` to the `step-ca` binary, allowing ports lower than 1024 to be used (https://superuser.com/a/892391).

As a nice side-effect, this also makes the example value ":443" from `step ca init` work from inside a Docker container.

Oh, and thank you a lot for the awesome work with Smallstep! :heart:

### DIFF
--- a/docker/Dockerfile.step-ca
+++ b/docker/Dockerfile.step-ca
@@ -7,6 +7,10 @@ ENV PWDPATH="/home/step/secrets/password"
 
 COPY $BINPATH "/usr/local/bin/step-ca"
 
+USER root
+RUN apk add --no-cache libcap && setcap CAP_NET_BIND_SERVICE=+eip /usr/local/bin/step-ca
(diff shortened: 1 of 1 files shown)


moqmar:
Oh, just noticed that this also solves #170 😃

dopey:
Hey @moqmar can you sign the CLA? The change looks good and ready to merge in.

Issue #178, opened 2020-02-07
Support for ALPN-01 Challenge


ibrt:
I was wondering if there are any plans to add support for the ALPN-01 challenge type... It's part of the standard, increasing in popularity, and very useful when the client cannot or is not willing to expose port 80.

By looking at your code I think it's going to be fairly straightforward... Adding the new challenge type [here](https://github.com/smallstep/certificates/blob/master/acme/challenge.go), using the [existing implementation](https://github.com/letsencrypt/boulder/blob/master/va/tlsalpn.go) from Boulder as reference. Plus a bit of plumbing here and there.

Are you willing to accept PRs for this? If so I can take a stab at it.

dcow:
@ibrt absolutely! We would love to support `alpn-01` challenges. If you have any questions along the way don't hesitate to ask either here or in our Gitter room (:

dopey:
I'm gonna close this. Thanks again for the contribution @ibrt!

PR #180, opened 2020-02-07
Support for ALPN-01 Challenge

ibrt:
As discussed in https://github.com/smallstep/certificates/issues/178.

This mostly follows the [existing implementation](https://github.com/letsencrypt/boulder/blob/master/va/tlsalpn.go) from Boulder with one exception: to keep it simple, instead of explicitly resolving the domain and trying to connect to all corresponding IP addresses, I let the Go standard library use its default behavior. It is still [RFC-compliant](https://tools.ietf.org/html/draft-ietf-acme-tls-alpn-01) (item 3.2), just maybe a bit less likely to succeed in some edge cases.

Besides updating the unit tests, I've also manually tested this by 1) building step-ca and running locally, 2) adding `test.local` to my hosts file, and 3) running `sudo lego --accept-tos --server="https://localhost:8443/acme/acme/directory" --email="..." --domains="test.local" --tls run`.


acme/challenge.go
```diff
@@ -344,6 +358,148 @@ func (hc *http01Challenge) validate(db nosql.DB, jwk *jose.JSONWebKey, vo valida
...
+		if err = tc.storeError(db,
+			RejectedIdentifierErr(errors.Errorf("incorrect certificate for tls-alpn-01 challenge: "+
+				"leaf certificate must contain a single DNS name, %v", tc.Value))); err != nil {
+			return nil, err
+		}
+		return tc, nil
+	}
+
+	idPeAcmeIdentifier := asn1.ObjectIdentifier{1, 3, 6, 1, 5, 5, 7, 1, 31}
+	idPeAcmeIdentifierV1Obsolete := asn1.ObjectIdentifier{1, 3, 6, 1, 5, 5, 7, 1, 30, 1}
```
cpu: It makes me a little bit sad to see this obsolete OID propagate further. Is backwards compatibility with incorrect TLS-ALPN-01 challenge responses important for Smallstep? The larger ACME ecosystem might benefit from a nudge in the right direction by taking a firmer stance in new code and only expecting `idPeAcmeIdentifier`.

For reference, the Let's Encrypt metrics show Boulder sees the obsolete OID ~25-30% of the time. Not an insubstantial amount, but also not a majority.

acme/challenge.go
```diff
@@ -344,6 +358,148 @@ func (hc *http01Challenge) validate(db nosql.DB, jwk *jose.JSONWebKey, vo valida
...
+			return nil, err
+		}
+		return tc, nil
+	}
+	defer conn.Close()
+
+	cs := conn.ConnectionState()
+	certs := cs.PeerCertificates
+
+	if len(certs) == 0 {
```
ibrt: Ok this is mildly frustrating - I don't want to remove this check but I don't see a way to trigger it in tests using the HTTPS server from the standard library... Any ideas welcome

acme/challenge.go
ibrt: I don't have an opinion - if you are going to be the one merging this PR (hopefully) and prefer me to remove it I'm happy to oblige, just let me know. 

It would seem wise though to still check for this value and return a descriptive error if validation fails because of it, WDYT?

dcow: Talked with @mmalone rather briefly, so apologies if we flip-flop, but we're cool with removing support in favor of a helpful error message, at least initially. If it's too frustrating for people, it's not hard at all to add back. Thanks @cpu for the critical eye.

ibrt: No problem, fixed it.

ibrt: So one interesting thing I just found out working on this is that the [golang.org/x/crypto/acme](https://github.com/golang/crypto/blob/master/acme/acme.go#L770) package is still generating certs with the obsolete identifier... I'm going to see if it can be updated :)

acme/challenge.go
maraino: @ibrt I don't think there's a way to trigger this in tests using the standard server, you might need to go very low level, and the tls client might crash before getting here.
Right now you can configure `tlsDial`, it's very possible that returning just a  `&tls.Conn{}` or perhaps a closed connection might do it. In any case, it's better to be safe always, so I wouldn't remove this check.

acme/challenge.go
maraino: @ibrt golang team welcomes contributions, you will need to create an issue first in golang/go, if it's not there yet, and then submit a PR, using GitHub, but by personal experience, I would recommend you to use directly Gerrit.
https://golang.org/doc/contribute.html

ibrt: They actually had a pending PR for that which has now been merged: https://github.com/golang/crypto/pull/91 :)

acme/challenge.go
ibrt: The latter was a good hint - I got it to work thank you.

acme/challenge_test.go
```diff
@@ -359,6 +446,32 @@ func TestChallengeUnmarshal(t *testing.T) {
...
+			assert.FatalError(t, err)
+			return test{
+				chb: b,
+				err: ServerInternalErr(errors.New("unexpected challenge type foo")),
+			}
+		},
+		"fail/unexpected-type-dns": func(t *testing.T) test {
+			dnsCh, err := newDNSCh()
+			assert.FatalError(t, err)
+			_tlsALPNCh, ok := dnsCh.(*dns01Challenge)
```
dopey: Should the names here be `_dnsCh`, etc? (assuming alpn is here by mistake)

acme/challenge_test.go
```diff
@@ -359,6 +446,32 @@ func TestChallengeUnmarshal(t *testing.T) {
...
+			assert.Fatal(t, ok)
+			_tlsALPNCh.baseChallenge.Type = "foo"
+			b, err := json.Marshal(dnsCh)
+			assert.FatalError(t, err)
+			return test{
+				chb: b,
+				err: ServerInternalErr(errors.New("unexpected challenge type foo")),
+			}
+		},
```
dopey: Can you add a `ok/tls-alpn01` challenge? Should be as easy as copy pasta-ing `ok/http` and just changing some of the var names.

acme/challenge_test.go
ibrt: Yup my bad

acme/challenge_test.go
ibrt: Done

ibrt:
It should be good now, thanks for reviewing!

PR #182, opened 2020-02-08
Fix help command indication

mkontani:
### Description
Thanks good product.
I tried one by one and found miss command indicaton about help.


dopey:
Nice spot again! Appreciate the thorough checks.

Issue #160, opened 2020-01-21
Doc issue - I already have PKI in place.


spencerryan:
In https://github.com/smallstep/certificates/blob/master/docs/questions.md

`I already have PKI in place. Can I use this with my own root certificate?` 

Is either missing a link, or is blank:

`Absolutely. [Details here].`

I'd love to see some documentation on how to do this!

dopey:
Hey, thanks for point this out! I agree that documentation leaves something to be desired ;p

We're actually working on a revamp of all of our documentation (structuring, missing items, etc.) and we'll be pulling this issue in to make sure it doesn't get forgotten.

In short, when you run `step ca init` this  creates a PKI with a root and intermediate certificate.  The root certificate and the intermediate cert + key are then referenced (by file location) in the `step-ca` configuration file - `ca.json`. 

Suppose you have your own root and intermediate certificates already. You can run `step ca init` and just replace the values in `ca.json` with the locations of your existing files and the CA should work.

Suppose you only have a root certificate and you'd like to bootstrap a new PKI using that root (so, create a new intermediate cert signed by that root). You can run `step ca init --root <my-root.pem> --key <my-root-priv-key>` and this will create an intermediate certificate and create a `ca.json` file with the correct root and intermediates already filled in.

Let me know if that answers your question, or if you'd like more details on some of the steps.

mmalone:
Just wrote some more detailed instructions in an email response to a similar question. They were asking about using AD CS as the existing root CA, but instructions will be the same except for the CSR signing step. Just dropping those notes here in case it helps someone else:

First, the easy answer. If you have your root CA signing key available you can run `step ca init --root=<path to root cert> --key=<path to root signing key>` and that should do what you want.That said, CAs are usually pretty locked down and it's bad practice to move the private key around. So I'm gonna assume that's not an option and give you the more complex instructions to do this "the right way", by generating a CSR for `step-ca`, getting it signed by your exiting root, and configuring `step-ca` to use it.

First, some background. When you run `step ca init` we create a couple artifacts under `~/.step/`. The important ones for us are:

  * `~/.step/certs/root_ca.crt` is your root CA certificate
  * `~/.step/secrets/root_ca_key` is your root CA signing key
  * `~/.step/certs/intermediate_ca.crt` is your intermediate CA cert
  * `~/.step/secrets/intermediate_ca_key` is the intermediate signing key used by `step-ca`

[...]

PR #199, opened 2020-02-24
Update FAQ: "I already have PKI"

tashian:
This addresses issue #160 

I extended @mmalone's example a bit and added it to the FAQ.
In addition to Mike's directions for AD CS, I added directions for AWS private CA, CFSSL, and OpenSSL.

Some small lingering questions:

- [x] Which CAs are most important to support in the directions?
- [x] Add CFSSL instructions
- [x] When issuing an intermediate cert for `step-ca`, which signing algo is best?
- [x] How long should the intermediate cert be valid for? 10 years
- [x] Test the AWS ACM PCA directions

The AD CS directions were confirmed by some folks in #160

### DIFF
--- a/docs/questions.md
+++ b/docs/questions.md
@@ -106,7 +106,103 @@ automated method as your system grows.
 
 ## I already have PKI in place. Can I use this with my own root certificate?
 
-Absolutely. [Details here].
+Yes. There's a easy way, and a longer but more secure way to do this.
+
+### Option 1: The easy way
+
+If you have your root CA signing key available, you can run:
+
+```bash
+step ca init --root=[ROOT_CERT_FILE] --key=[ROOT_PRIVATE_KEY_FILE]
+```
+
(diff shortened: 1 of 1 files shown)


docs/questions.md
```diff
@@ -106,7 +106,153 @@ automated method as your system grows.
...
+
+```bash
+step ca init --root=[ROOT_CERT_FILE] --key=[ROOT_PRIVATE_KEY_FILE]
+```
+
+The root certificate can be in PEM or DER format, and the signing key can be a PEM file containing a PKCS#1, PKCS#8, or RFC5915 (for EC) key.
+
+### Option 2: More secure
+
+That said, CAs are usually pretty locked down and it's bad practice to move the private key around. So I'm gonna assume that's not an option and give you the more complex instructions to do this "the right way", by generating a CSR for `step-ca`, getting it signed by your exiting root, and configuring `step-ca` to use it.
```
mmalone: s/exiting/existing

PR #220, opened 2020-03-31
Enforce a duration for identity certificates

maraino:
### Description
This PR enforce the same duration of the SSH certificates in the identity certificates.

### DIFF
--- a/api/ssh.go
+++ b/api/ssh.go
@@ -489,7 +489,7 @@ type identityModifier struct {
 	NotAfter  time.Time
 }
 
-func (m *identityModifier) Constrain(cert *x509.Certificate) error {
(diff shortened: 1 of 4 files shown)


maraino:
@dopey 

> 1. Why are we using this rather than the existing provisioner.SignOptions method of setting the nbf and naf?

We cannot use a regular SignOptions method because there's a validator that will block that option. This specific method is executed after all the validators. For the same reason we cannot use the provisioner.Options.

> 2. If we use this then the verb is contrain but the noun would be constraint. So the method would be constrain but the object/type would be constraintModifier.

Does this looks right ?
```
type CertificateConstraintModifier interface {
	SignOption
	Constrain(cert *x509.Certificate) error
}
```

Issue #235, opened 2020-04-27
ACME directory URL response should use the base host in the request URL, not a hardcoded URL from the ca config.


dopey:
See #193

Issue #193, opened 2020-02-19
step-ca is hard to configure behind a reverse proxy / load balancer


jholloway7:
### What would you like to be added

step-ca hardwires the host name and port in its ACME directory response. The makes it difficult to configure the server to run behind a reverse proxy or load balancer that is using a different name/port than the first one listed in the step-ca configuration.  

The typical solution is to use the Host header in the HTTP request to return a URL in the same form that was requested by the client.  Some frameworks do validate the Host header against a known list of "valid" domains and return 400 Bad Request if a request makes it to the server with an invalid host name.   

Example:
https://acme.secops-live.local/acme/terraform/directory

If step-ca is configure with dnsNames = [ "localhost", "acme.secops-live.local" ], the response from the directory will contain https://localhost/acme/terraform/...  which isn't reachable by a client that only knows how to reach the directory via https://acme.secops-live.local.

NOTE: This pattern may also affect other HTTP APIs within step-ca outside of the ACME protocol. I do recall having some issues with the step CLI that seemed to be related, but I didn't notice the underlying issue until trying to use it as an ACME provider.

https://github.com/smallstep/certificates/blame/8e882faf44f89178e769cf53923b4db8222bea54/ca/ca.go#L116

I think there may also be some relevant bits in the ACME v2 spec, but I'm not that familiar with JWS to know for sure what they're saying here:

https://ietf-wg-acme.github.io/acme/draft-ietf-acme-acme.html#rfc.section.6.4

The workaround is to list the dns name first and to use the standard 443 port.  These options work for my limited use case, but would be fragile in trying to do anything slightly more sophisticated.

### Why this is needed

Most applications these days need the ability to live behind a reverse proxy and/or load balancer to give the deployer options on how to configure their network without directly exposing the application.

Thanks!

cpu:
> The typical solution is to use the Host header in the HTTP request to return a URL in the same form that was requested by the client. 

For what it's worth this is the approach Let's Encrypt uses to [generate directory URLs](https://github.com/letsencrypt/boulder/blob/27e785f3f238f4811a44f485c141ac1b925d47cb/web/relative.go#L8-L10) in Boulder (which is deployed behind a reverse proxy).

t0xicCode:
To add to this, it's also an issue if you configure step-ca to run on a non-privileged port, then expose a different port to the external network. 

[...]

PR #236, opened 2020-04-27
Add new acmeHost config option

clive-jevons:
### Description

In order to be able to override the calculated host name used
by the ACME directory (dns[0] + <port if not 443>), add an extra
config option called acmeHost, which will be used instead. It's
optional, and if not specified, behaviour will be as before.

Related to #193 and #235 also step-ca helm chart issue nr 4

This is my very first foray into golang, hence I'd really appreciate some help vis-a-vis
how best to test this change and generally any feedback :)

### DIFF
--- a/authority/config.go
+++ b/authority/config.go
@@ -55,6 +55,7 @@ type Config struct {
 	IntermediateKey  string               `json:"key"`
 	Address          string               `json:"address"`
 	DNSNames         []string             `json:"dnsNames"`
+	ACMEHost         string               `json:"acmeHost,omitempty"`
 	KMS              *kms.Options         `json:"kms,omitempty"`
 	SSH              *SSHConfig           `json:"ssh,omitempty"`
 	Logger           json.RawMessage      `json:"logger,omitempty"`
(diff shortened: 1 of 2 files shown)


dopey:
Thanks for the PR @clive-jevons! We're finally getting around to tackling a bunch of open ACME issues this week, so this was very timely.

Looking at the [original issue](https://github.com/smallstep/certificates/issues/193) and the comment by @cpu, I think we want to implement this in the more robust / "relative way" that boulder has done. 

If that's something you want to take a pass at, let me know. One of us (smallstep team) will be grabbing this ticket tomorrow - ish so you're welcome to work with us or help with review.

dopey:
With regards to your actual golang changes, they look great (I'm no exert either tbf). Honestly, if I had implemented it your way in the beginning we probably wouldn't need to fix this now.

clive-jevons:
Thanks for the feedback @dopey 😊
Just checked out the linked boulder code, and it does indeed look very sensible. Time allowing I'll try and get involved 👍

clive-jevons:
Hey @dopey ,

I've added a new PR: https://github.com/smallstep/certificates/pull/243 ... is that more what you're looking for?
Currently still contains a TODO because I was unsure about whether to force the HTTPS protocol. The Boulder implementation seems to allow other protocols, if I've understood their code correctly. Should be a minor change though if you require it to always be HTTPS ;) 

Cheers,
C

dopey:
Closing this as it was merged as part of another PR.

PR #242, opened 2020-04-30
8555 § 8.3 Challenge Retries

dcow:
Taking over #181.

### DIFF
--- a/acme/api/handler.go
+++ b/acme/api/handler.go
@@ -180,8 +180,9 @@ func (h *Handler) GetChallenge(w http.ResponseWriter, r *http.Request) {
 	ch, err = h.Auth.ValidateChallenge(prov, acc.GetID(), chID, acc.GetKey())
 	if err != nil {
 		api.WriteError(w, err)
-	} else if ch.Status != acme.StatusValid && ch.Status != acme.StatusInvalid {
-		w.Header().Add("Retry-After", "60")
+	} else if ch.Retry.Active {
+		retryAfter := int(ch.Retry.Backoffs) * (10 - ch.Retry.Called)
+		w.Header().Add("Retry-After", string(retryAfter))
 		api.JSON(w, ch)
 	} else {
 		getLink := h.Auth.GetLink
--- a/acme/api/handler.go
+++ b/acme/api/handler.go
@@ -183,7 +183,7 @@ func (h *Handler) GetChallenge(w http.ResponseWriter, r *http.Request) {
 	} else if ch.Retry.Active {
 		retryAfter := int(ch.Retry.Backoffs) * (10 - ch.Retry.Called)
 		w.Header().Add("Retry-After", string(retryAfter))
-		api.JSON(w, ch)
+		api.WriteProcessing(w, ch)
 	} else {
 		getLink := h.Auth.GetLink
 		w.Header().Add("Link", link(getLink(acme.AuthzLink, acme.URLSafeProvisionerName(prov), true, ch.GetAuthzID()), "up"))
--- a/acme/api/handler.go
+++ b/acme/api/handler.go
@@ -2,14 +2,13 @@ package api
 
 import (
 	"fmt"
-	"net/http"
-
 	"github.com/go-chi/chi"
 	"github.com/pkg/errors"
 	"github.com/smallstep/certificates/acme"
 	"github.com/smallstep/certificates/api"
 	"github.com/smallstep/certificates/authority/provisioner"
 	"github.com/smallstep/cli/jose"
+	"net/http"
 )
 
 func link(url, typ string) string {
@@ -181,13 +180,15 @@ func (h *Handler) GetChallenge(w http.ResponseWriter, r *http.Request) {
 	ch, err = h.Auth.ValidateChallenge(prov, acc.GetID(), chID, acc.GetKey())
 	if err != nil {
 		api.WriteError(w, err)
-		return
+	} else if ch.Status != acme.StatusValid && ch.Status != acme.StatusInvalid {
+		w.Header().Add("Retry-After", "60")
+		api.JSON(w, ch)
+	} else {
+		getLink := h.Auth.GetLink
+		w.Header().Add("Link", link(getLink(acme.AuthzLink, acme.URLSafeProvisionerName(prov), true, ch.GetAuthzID()), "up"))
+		w.Header().Set("Location", getLink(acme.ChallengeLink, acme.URLSafeProvisionerName(prov), true, ch.GetID()))
+		api.JSON(w, ch)
 	}
-
-	getLink := h.Auth.GetLink
-	w.Header().Add("Link", link(getLink(acme.AuthzLink, acme.URLSafeProvisionerName(prov), true, ch.GetAuthzID()), "up"))
-	w.Header().Set("Location", getLink(acme.ChallengeLink, acme.URLSafeProvisionerName(prov), true, ch.GetID()))
-	api.JSON(w, ch)
 }
 
 // GetCertificate ACME api for retrieving a Certificate.
--- a/acme/api/handler.go
+++ b/acme/api/handler.go
@@ -181,9 +181,13 @@ func (h *Handler) GetChallenge(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		api.WriteError(w, err)
 	} else if ch.Retry.Active {
-		retryAfter := int(ch.Retry.Backoffs) * (10 - ch.Retry.Called)
-		w.Header().Add("Retry-After", string(retryAfter))
-		api.WriteProcessing(w, ch)
+		retryAfter, err := h.Auth.BackoffChallenge(prov, acc.GetID(), chID, acc.GetKey())
+		if err != nil {
+			api.WriteError(w, err)
+		} else {
+			w.Header().Add("Retry-After", retryAfter.String())
+			api.WriteProcessing(w, ch)
+		}
 	} else {
 		getLink := h.Auth.GetLink
 		w.Header().Add("Link", link(getLink(acme.AuthzLink, acme.URLSafeProvisionerName(prov), true, ch.GetAuthzID()), "up"))
--- a/acme/api/handler_test.go
+++ b/acme/api/handler_test.go
@@ -41,6 +41,7 @@ type mockAcmeAuthority struct {
 	updateAccount       func(provisioner.Interface, string, []string) (*acme.Account, error)
 	useNonce            func(string) error
 	validateChallenge   func(p provisioner.Interface, accID string, id string, jwk *jose.JSONWebKey) (*acme.Challenge, error)
+	backoffChallenge    func(p provisioner.Interface, accID, chID string, jwk *jose.JSONWebKey) (time.Duration, error)
 	ret1                interface{}
 	err                 error
 }
@@ -200,6 +201,17 @@ func (m *mockAcmeAuthority) ValidateChallenge(p provisioner.Interface, accID str
 		return nil, m.err
 	default:
 		return m.ret1.(*acme.Challenge), m.err
+	}
+}
+
+func (m *mockAcmeAuthority) BackoffChallenge(p provisioner.Interface, accID, chID string, jwk *jose.JSONWebKey) (time.Duration, error) {
+	switch {
+	case m.backoffChallenge != nil:
+		return m.backoffChallenge(p, accID, chID, jwk)
+	case m.err != nil:
+		return -1, m.err
+	default:
+		return m.ret1.(time.Duration), m.err
 	}
 }
 
--- a/acme/api/handler_test.go
+++ b/acme/api/handler_test.go
@@ -589,6 +589,7 @@ func ch() acme.Challenge {
 		URL:     "https://ca.smallstep.com/acme/challenge/chID",
 		ID:      "chID",
 		AuthzID: "authzID",
+		Retry:   &acme.Retry{Called:0, Backoffs:1, Active:false},
 	}
 }
 
@@ -733,6 +734,37 @@ func TestHandlerGetChallenge(t *testing.T) {
 				ch:         ch,
 			}
 		},
+		"ok/retry-after": func(t *testing.T) test {
+			key, err := jose.GenerateJWK("EC", "P-256", "ES256", "sig", "", 0)
+			assert.FatalError(t, err)
+			acc := &acme.Account{ID: "accID", Key: key}
+			ctx := context.WithValue(context.Background(), provisionerContextKey, prov)
+			ctx = context.WithValue(ctx, accContextKey, acc)
+			// TODO: Add correct key such that challenge object is already "active"
+			chiCtxInactive := chi.NewRouteContext()
+			chiCtxInactive.URLParams.Add("chID", "chID")
+			//chiCtxInactive.URLParams.Add("Active", "true")
+			ctx = context.WithValue(ctx, chi.RouteCtxKey, chiCtxInactive)
+			ch := ch()
+			ch.Retry.Active = true
+			chJSON, err := json.Marshal(ch)
+			assert.FatalError(t, err)
+			ctx = context.WithValue(ctx, payloadContextKey, &payloadInfo{value: chJSON})
+			return test{
+				auth: &mockAcmeAuthority{
+					validateChallenge: func(p provisioner.Interface, accID, id string, jwk *jose.JSONWebKey) (*acme.Challenge, error) {
+						assert.Equals(t, p, prov)
+						assert.Equals(t, accID, acc.ID)
+						assert.Equals(t, id, ch.ID)
+						assert.Equals(t, jwk.KeyID, key.KeyID)
+						return &ch, nil
+					},
+				},
+				ctx:        ctx,
+				statusCode: 200,
+				ch:         ch,
+			}
+		},
 	}
 	for name, run := range tests {
 		tc := run(t)
@@ -760,13 +792,17 @@ func TestHandlerGetChallenge(t *testing.T) {
 				assert.Equals(t, ae.Identifier, prob.Identifier)
 				assert.Equals(t, ae.Subproblems, prob.Subproblems)
 				assert.Equals(t, res.Header["Content-Type"], []string{"application/problem+json"})
-			} else {
+			} else if res.StatusCode >= 200 && assert.True(t,res.Header["Retry-After"] == nil){
 				expB, err := json.Marshal(tc.ch)
 				assert.FatalError(t, err)
 				assert.Equals(t, bytes.TrimSpace(body), expB)
 				assert.Equals(t, res.Header["Link"], []string{fmt.Sprintf("<https://ca.smallstep.com/acme/authz/%s>;rel=\"up\"", tc.ch.AuthzID)})
 				assert.Equals(t, res.Header["Location"], []string{url})
 				assert.Equals(t, res.Header["Content-Type"], []string{"application/json"})
+			} else {
+				expB, err := json.Marshal(tc.ch)
+				assert.FatalError(t, err)
+				assert.Equals(t, bytes.TrimSpace(body), expB)
 			}
 		})
 	}
--- a/acme/api/handler_test.go
+++ b/acme/api/handler_test.go
@@ -792,17 +792,19 @@ func TestHandlerGetChallenge(t *testing.T) {
 				assert.Equals(t, ae.Identifier, prob.Identifier)
 				assert.Equals(t, ae.Subproblems, prob.Subproblems)
 				assert.Equals(t, res.Header["Content-Type"], []string{"application/problem+json"})
-			} else if res.StatusCode >= 200 && assert.True(t,res.Header["Retry-After"] == nil){
+			} else if res.StatusCode >= 200 {
 				expB, err := json.Marshal(tc.ch)
 				assert.FatalError(t, err)
 				assert.Equals(t, bytes.TrimSpace(body), expB)
 				assert.Equals(t, res.Header["Link"], []string{fmt.Sprintf("<https://ca.smallstep.com/acme/authz/%s>;rel=\"up\"", tc.ch.AuthzID)})
 				assert.Equals(t, res.Header["Location"], []string{url})
 				assert.Equals(t, res.Header["Content-Type"], []string{"application/json"})
-			} else {
+			} else if res.StatusCode >= 100 {
 				expB, err := json.Marshal(tc.ch)
 				assert.FatalError(t, err)
 				assert.Equals(t, bytes.TrimSpace(body), expB)
+				assert.True(t, res.Header["Retry-After"] != nil)
+				assert.Equals(t, res.Header["Content-Type"], []string{"application/json"})
 			}
 		})
 	}
--- a/acme/api/handler_test.go
+++ b/acme/api/handler_test.go
@@ -589,7 +589,7 @@ func ch() acme.Challenge {
 		URL:     "https://ca.smallstep.com/acme/challenge/chID",
 		ID:      "chID",
 		AuthzID: "authzID",
-		Retry:   &acme.Retry{Called:0, Backoffs:1, Active:false},
+		Retry:   &acme.Retry{Called: 0, Active: false},
 	}
 }
 
--- a/acme/authority.go
+++ b/acme/authority.go
@@ -278,7 +278,7 @@ func (a *Authority) ValidateChallenge(p provisioner.Interface, accID, chID strin
 	dialer := &net.Dialer{
 		Timeout: 30 * time.Second,
 	}
-	for i:=0; i < 10; i++ {
+	for ch.getRetry().Active {
 		ch, err = ch.validate(a.db, jwk, validateOptions{
 			httpGet:   client.Get,
 			lookupTxt: net.LookupTXT,
--- a/acme/authority.go
+++ b/acme/authority.go
@@ -5,6 +5,8 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
+	"math"
+	"math/rand"
 	"net"
 	"net/http"
 	"net/url"
@@ -263,21 +265,38 @@ func (a *Authority) ValidateChallenge(p provisioner.Interface, accID, chID strin
 	if accID != ch.getAccountID() {
 		return nil, UnauthorizedErr(errors.New("account does not own challenge"))
 	}
+	retry := ch.getRetry()
+	if retry.Active {
+		return ch.toACME(a.db, a.dir, p)
+	}
+	retry.Mux.Lock()
+	defer retry.Mux.Unlock()
+
 	client := http.Client{
 		Timeout: time.Duration(30 * time.Second),
 	}
 	dialer := &net.Dialer{
 		Timeout: 30 * time.Second,
 	}
-	ch, err = ch.validate(a.db, jwk, validateOptions{
-		httpGet:   client.Get,
-		lookupTxt: net.LookupTXT,
-		tlsDial: func(network, addr string, config *tls.Config) (*tls.Conn, error) {
-			return tls.DialWithDialer(dialer, network, addr, config)
-		},
-	})
-	if err != nil {
-		return nil, Wrap(err, "error attempting challenge validation")
+	for i:=0; i < 10; i++ {
+		ch, err = ch.validate(a.db, jwk, validateOptions{
+			httpGet:   client.Get,
+			lookupTxt: net.LookupTXT,
+			tlsDial: func(network, addr string, config *tls.Config) (*tls.Conn, error) {
+				return tls.DialWithDialer(dialer, network, addr, config)
+			},
+		})
+		if err != nil {
+			return nil, Wrap(err, "error attempting challenge validation")
+		}
+		if ch.getStatus() == StatusValid {
+			break
+		}
+		if ch.getStatus() == StatusInvalid {
+			return ch.toACME(a.db, a.dir, p)
+		}
+		duration := time.Duration(ch.getRetry().Backoffs + math.Mod(rand.Float64(), 5))
+		time.Sleep(duration*time.Second)
 	}
 	return ch.toACME(a.db, a.dir, p)
 }
--- a/acme/authority.go
+++ b/acme/authority.go
@@ -6,7 +6,6 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"math"
-	"math/rand"
 	"net"
 	"net/http"
 	"net/url"
@@ -21,6 +20,7 @@ import (
 
 // Interface is the acme authority interface.
 type Interface interface {
+	BackoffChallenge(provisioner.Interface, string, string, *jose.JSONWebKey) (time.Duration, error)
 	DeactivateAccount(provisioner.Interface, string) (*Account, error)
 	FinalizeOrder(provisioner.Interface, string, string, *x509.CertificateRequest) (*Order, error)
 	GetAccount(provisioner.Interface, string) (*Account, error)
@@ -295,8 +295,7 @@ func (a *Authority) ValidateChallenge(p provisioner.Interface, accID, chID strin
 		if ch.getStatus() == StatusInvalid {
 			return ch.toACME(a.db, a.dir, p)
 		}
-		duration := time.Duration(ch.getRetry().Backoffs + math.Mod(rand.Float64(), 5))
-		time.Sleep(duration*time.Second)
+		time.Sleep(ch.getBackoff())
 	}
 	return ch.toACME(a.db, a.dir, p)
 }
@@ -311,4 +310,25 @@ func (a *Authority) GetCertificate(accID, certID string) ([]byte, error) {
 		return nil, UnauthorizedErr(errors.New("account does not own certificate"))
 	}
 	return cert.toACME(a.db, a.dir)
+}
+
+// BackoffChallenge returns the total time to wait until the next sequence of validation attempts completes
+func (a *Authority) BackoffChallenge(p provisioner.Interface, accID, chID string, jwk *jose.JSONWebKey) (time.Duration, error) {
+	ch, err := getChallenge(a.db, chID)
+	if err != nil {
+		return -1, err
+	}
+	if accID != ch.getAccountID() {
+		return -1, UnauthorizedErr(errors.New("account does not own challenge"))
+	}
+
+	remCalls := ch.getRetry().MaxAttempts - math.Mod(ch.getRetry().Called, ch.getRetry().MaxAttempts)
+	totBackoff := 0*time.Second
+	for i:=0; i < int(remCalls); i++ {
+		clone := ch.clone()
+		clone.Retry.Called += float64(i)
+		totBackoff += clone.getBackoff()
+	}
+
+	return totBackoff, nil
 }
--- a/acme/authority.go
+++ b/acme/authority.go
@@ -323,8 +323,8 @@ func (a *Authority) BackoffChallenge(p provisioner.Interface, accID, chID string
 	}
 
 	remCalls := ch.getRetry().MaxAttempts - math.Mod(ch.getRetry().Called, ch.getRetry().MaxAttempts)
-	totBackoff := 0*time.Second
-	for i:=0; i < int(remCalls); i++ {
+	totBackoff := 0 * time.Second
+	for i := 0; i < int(remCalls); i++ {
 		clone := ch.clone()
 		clone.Retry.Called += float64(i)
 		totBackoff += clone.getBackoff()
--- a/acme/challenge.go
+++ b/acme/challenge.go
@@ -11,6 +11,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"math"
+	"math/rand"
 	"net"
 	"net/http"
 	"strings"
@@ -38,10 +40,14 @@ type Challenge struct {
 }
 
 type Retry struct {
-	Called    int         `json:"id"`
-	Backoffs  float64     `json:"backoffs"`
-	Active    bool        `json:"active"`
-	Mux	      sync.Mutex  `json:"mux"`
+	Called       float64        `json:"id"`
+	BaseDelay    time.Duration  `json:"basedelay"`
+	MaxDelay     time.Duration  `json:"maxdelay"`
+	Multiplier   float64        `json:"multiplier"`
+	Jitter       float64        `json:"jitter"`
+	MaxAttempts  float64        `json:"maxattempts"`
+	Active       bool           `json:"active"`
+	Mux	         sync.Mutex     `json:"mux"`
 }
 
 // ToLog enables response logging.
@@ -85,6 +91,7 @@ type challenge interface {
 	getAuthzID() string
 	getToken() string
 	getRetry() *Retry
+	getBackoff() time.Duration
 	clone() *baseChallenge
 	getAccountID() string
 	getValidated() time.Time
@@ -131,7 +138,7 @@ func newBaseChallenge(accountID, authzID string) (*baseChallenge, error) {
 		Status:    StatusPending,
 		Token:     token,
 		Created:   clock.Now(),
-		Retry:     &Retry{Called:0, Backoffs:1, Active:false},
+		Retry:     &Retry{Called:0, BaseDelay:1, MaxDelay: 30, Jitter: 1, MaxAttempts:10, Active:false},
 	}, nil
 }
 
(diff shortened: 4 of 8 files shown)


acme/authority.go
```diff
@@ -263,21 +265,37 @@ func (a *Authority) ValidateChallenge(p provisioner.Interface, accID, chID strin
 	if accID != ch.getAccountID() {
 		return nil, UnauthorizedErr(errors.New("account does not own challenge"))
 	}
+	retry := ch.getRetry()
+	if retry.Active {
+		return ch.toACME(a.db, a.dir, p)
+	}
+	retry.Mux.Lock()
```
dcow, line 272: Not sure we need a mutex. I can't think of why multiple routines would be trying to write retry at once. On the read side I assume writes to the db are atomic and reads won't view partially updated data. TODO: verify. cc @dopey

acme/authority.go
```diff
@@ -263,21 +265,37 @@ func (a *Authority) ValidateChallenge(p provisioner.Interface, accID, chID strin
...
+		if err != nil {
+			return nil, Wrap(err, "error attempting challenge validation")
+		}
+		if ch.getStatus() == StatusValid {
+			break
+		}
+		if ch.getStatus() == StatusInvalid {
+			return ch.toACME(a.db, a.dir, p)
+		}
+		time.Sleep(ch.getBackoff())
```
dcow, line 298: I would like to change this to a timer with context so we can cancel pending retries when a new request comes in. Also, a replicated CA needs to be able to communicate not just retry state with other CAs, but share the responsibility of executing retries. I _think_ I have a solution but I need to review the spec more to make sure there's a meaningful difference between a client re-post-ing to the challenge endoing and a client requiesting an entirely new validation.

acme/authority.go
```diff
@@ -293,3 +311,24 @@ func (a *Authority) GetCertificate(accID, certID string) ([]byte, error) {
...
+		return -1, err
+	}
+	if accID != ch.getAccountID() {
+		return -1, UnauthorizedErr(errors.New("account does not own challenge"))
+	}
+
+	remCalls := ch.getRetry().MaxAttempts - math.Mod(ch.getRetry().Called, ch.getRetry().MaxAttempts)
+	totBackoff := 0 * time.Second
+	for i := 0; i < int(remCalls); i++ {
+		clone := ch.clone()
```
dcow, line 328: why clone?

acme/challenge.go
```diff
@@ -101,6 +118,7 @@ type baseChallenge struct {
 	Validated time.Time `json:"validated"`
 	Created   time.Time `json:"created"`
 	Error     *AError   `json:"error"`
+	Retry     *Retry     `json:"retry"`
```
dcow, line 121: I don't know if we need to return this. 8555 says to include retry info in the error field.

acme/challenge.go
```diff
@@ -303,21 +349,32 @@ func newHTTP01Challenge(db nosql.DB, ops ChallengeOptions) (challenge, error) {
 // updated.
 func (hc *http01Challenge) validate(db nosql.DB, jwk *jose.JSONWebKey, vo validateOptions) (challenge, error) {
 	// If already valid or invalid then return without performing validation.
-	if hc.getStatus() == StatusValid || hc.getStatus() == StatusInvalid {
+	if hc.getStatus() == StatusValid {
+		return hc, nil
+	}
+	if hc.getStatus() == StatusInvalid {
+		// TODO: Resolve segfault on upd.save
```
dcow, line 356: This happens not as a result of the new test in this PR, but because things in the invalid state now hit the db, and the "ok/status-already-invalid" test doesn't provide a mock db so the test runner tries to invoke a null function.

acme/challenge.go
```diff
@@ -303,21 +349,32 @@ func newHTTP01Challenge(db nosql.DB, ops ChallengeOptions) (challenge, error) {
...
+		upd := hc.clone()
+		upd.Status = StatusPending
+		if err := upd.save(db, hc); err != nil {
+			fmt.Printf("Error in save: %s\n\n\n", err)
+			return nil, err
+		}
+		fmt.Print("Through Save\n\n")
```
dcow, line 363: remove

acme/challenge.go
```diff
@@ -303,21 +349,32 @@ func newHTTP01Challenge(db nosql.DB, ops ChallengeOptions) (challenge, error) {
...
-	if hc.getStatus() == StatusValid || hc.getStatus() == StatusInvalid {
+	if hc.getStatus() == StatusValid {
+		return hc, nil
+	}
+	if hc.getStatus() == StatusInvalid {
+		// TODO: Resolve segfault on upd.save
+		upd := hc.clone()
+		upd.Status = StatusPending
+		if err := upd.save(db, hc); err != nil {
+			fmt.Printf("Error in save: %s\n\n\n", err)
```
dcow, line 360: remove

acme/challenge.go
```diff
@@ -335,22 +392,28 @@ func (hc *http01Challenge) validate(db nosql.DB, jwk *jose.JSONWebKey, vo valida
...
 	expected, err := KeyAuthorization(hc.Token, jwk)
 	if err != nil {
+		// TODO retry?
 		return nil, err
 	}
 	if keyAuth != expected {
-		if err = hc.storeError(db,
-			RejectedIdentifierErr(errors.Errorf("keyAuthorization does not match; "+
-				"expected %s, but got %s", expected, keyAuth))); err != nil {
+		// TODO should we just bail here, this is a "successful" failure.
```
dcow, line 399: todo

acme/challenge.go
```diff
@@ -335,22 +392,28 @@ func (hc *http01Challenge) validate(db nosql.DB, jwk *jose.JSONWebKey, vo valida
...
+		if err = hc.iterateRetry(db, RejectedIdentifierErr(errors.Errorf("keyAuthorization does not match; "+
+			"expected %s, but got %s", expected, keyAuth))); err != nil {
+			// TODO what's the difference between returning nil with an error, and returning a retry-
```
dcow, line 402: todo

acme/challenge.go
```diff
@@ -335,22 +392,28 @@ func (hc *http01Challenge) validate(db nosql.DB, jwk *jose.JSONWebKey, vo valida
 
 	expected, err := KeyAuthorization(hc.Token, jwk)
 	if err != nil {
+		// TODO retry?
```
dcow, line 395: todo

acme/challenge.go
```diff
@@ -494,15 +557,16 @@ func (tc *tlsALPN01Challenge) validate(db nosql.DB, jwk *jose.JSONWebKey, vo val
 	}
 
 	if foundIDPeAcmeIdentifierV1Obsolete {
-		if err = tc.storeError(db,
+		// TODO this isn't likely going to change without user action, this is probably a good case for a hard failure...
```
dcow, line 560: todo: discuss

api/utils.go
```diff
@@ -52,6 +52,11 @@ func JSON(w http.ResponseWriter, v interface{}) {
 	JSONStatus(w, v, http.StatusOK)
 }
 
+// JSON writes the passed value into the http.ResponseWriter.
+func WriteProcessing(w http.ResponseWriter, v interface{}) {
+	JSONStatus(w, v, http.StatusProcessing)
```
dcow, line 57: I don't think we want to use status processing right here. Specifically, https://tools.ietf.org/html/rfc2518#section-10.1 says:

> This status code SHOULD only be sent
> when the server has a reasonable expectation that the request will
> take significant time to complete. As guidance, if a method is taking
> longer than 20 seconds (a reasonable, but arbitrary value) to process
> the server SHOULD return a 102 (Processing) response. The server MUST
> send a final response after the request has been completed.

Also, in general HTTP is concerned with the exchange of information about resources not some domain-specific notion of resource state.

acme/challenge.go
```diff
@@ -173,6 +197,25 @@ func (bc *baseChallenge) getError() *AError {
...
+	if bc.Retry.Called == 0 {
+		return bc.Retry.BaseDelay
+	}
+
+	backoff, max := float64(bc.Retry.BaseDelay), float64(bc.Retry.MaxDelay)
+	backoff *= math.Pow(bc.Retry.Multiplier, bc.Retry.Called)
+	if backoff > max {
+		backoff = max
+	}
+	// Introduce Jitter to ensure that clustered requests wont operate in unison
```
dcow, line 211: Not sure this matters either because we're not doing anything on a periodic basis. Only scenario where this might apply is if some client does a large batch of challenges all at once. In this case I think we would delegate managing that scenario to rate-limiting code.

dcow:
https://tools.ietf.org/html/rfc8555#section-7.1.6 illustrates the challenge lifecycle. Retries are mentioned:

> Note that within the "processing" state, the server may attempt to validate the challenge multiple times (see Section 8.2).  Likewise, client requests for retries do not cause a state change.  If validation is successful, the challenge moves to the "valid" state; if there is an error, the challenge moves to the "invalid" state.

I read this as _"if the server successfully acquires a challenge result (e.g. http resource, dns record, tls cert) but verification fails (the data is not correct), then move to the "invalid" state"_. This type of "error" is different from a transient network error which should be retried.

So I'm going to sift through all the error logic and bucket things into one of those two definitions. And we'll only retry for transient errors.

Also, once the state is "invalid", I don't think a challenge can be revived. This means the proposed implementation, where a challenge lives in the invalid state between retry attempts, is not correct. The challenge should remain pending until it's either valid or invalid.

cpu:
> Also, once the state is "invalid", I don't think a challenge can be revived. This means the proposed implementation, where a challenge lives in the invalid state between retry attempts, is not correct. The challenge should remain pending until it's either valid or invalid.

@dcow I _think_ you would want it to remain in a "processing" state instead of a "pending" state for failed attempts that support further retries. I think that's the meant of the errata that Rob from Sectico filed against RFC 8555 related to retries, `eid5732`: https://www.rfc-editor.org/errata/eid5732

acme/authority.go
dcow: s/validation/_order_

dcow:
> (_Edit: this part of the spec is a little bit rough. Despite my best efforts it landed in 8555 without anyone ever having implemented it at the time... :sweat: _)

@cpu thanks for the color. The errata is super useful—figuring out when to consider a challenge invalid is exactly where I'm hung up. Do you have any insight on:

> While the server is still trying, the status of the challenge remains "processing"; it is only marked "invalid" once the server has given up.

from section 8.2 with respect to `eid5732`? Is the intention that orders are only allowed to be in the processing state for a short amount of time after the initial client request and the server needs to put the kibosh on future client retries by marking the challenge invalid? I'm inclined to let a challenge remain in a processing state indefinitely until an authorization expires so as to allow clients to request a retry once the infrastructure is ready. I guess one could interpret that behavior as a server that never "gives up" even if it won't automatically retry again unless the client pokes it.

acme/authority.go
dcow: Upon review, there is definitely a difference between a re-requesting a challenge and requesting a new order.

My working strategy will be to have the server check the challenge record prior to executing a retry in order to make sure the current time is _after_ a `RetryAfter` field and the current Count is less than some maximum. If the RetryAfter field is in the past, prior to executing the retry, the server will update the RetryAfter field to be in the future and increment the Count field. 

If a server receives a new request on the challenge endpoint, it will first reset the count field on the challenge record, move RetryAfter into the future, and then immediately perform a validation attempt. Upon transient error, the server will schedule a retry at the time it listed in the RetryAfter field.

I think it also makes sense to add an identifier field so that the instance receiving the latest client challenge request can take over retries by updating the field to its own id.  When instances start, they query for all outstanding challenges with retries < MAX and ordinal == theirOrdinal. This enables graceful restart in a replicated context without one host sucking up all the retry responsibility. In k8s, you would deploy step-ca as a stateful set and pass the ordinal as this identifier.

cpu:
> from section 8.2? Is the intention that orders are only allowed to be in the processing state for a short amount of time after the initial client request and the server needs to put the kibosh on future client retries by marking the challenge invalid? 

@dcow I think an _order_ should only enter into the processing state when the client POSTs the finalization URL to begin issuance. At that point all of the challenges need to be valid for the finalization to start. In practice I suspect most clients expect orders to not remain in the processing state for very long simply because Let's Encrypt's issuance is more or less synchronous with the finalization request.

> I'm inclined to let a challenge remain in a processing state indefinitely until an authorization expires or otherwise becomes invalid. I guess one could interpret that behavior as a server that never "gives up" even if it won't automatically retry again unless the client pokes it.

I think that makes sense for the _challenges_ :+1: I can't think of a reason that would be a problem.

Hope that helps!

dcow:
Remaining TODO:

- [ ] Reschedule retries when the ca starts up (from a crash or otherwise).
- [ ] Test & Tests

acme/api/handler_test.go
```diff
@@ -203,6 +204,17 @@ func (m *mockAcmeAuthority) ValidateChallenge(p provisioner.Interface, accID str
 	}
 }
 
+func (m *mockAcmeAuthority) BackoffChallenge(p provisioner.Interface, accID, chID string, jwk *jose.JSONWebKey) (time.Duration, error) {
```
dcow, line 207: remove

dcow:
Re rescheduling when we come back up. Discussed with @dopey and @maraino this morning. Given our currentl limitation to k/v db, we decided to split this work out into a separate ticket. We may want to wait to address it until we have better support for crafting the type of query we want: select all the challenges where owner = my_ordinal && status == processing && num_retries < max_retries

acme/api/handler.go
```diff
@@ -149,31 +148,63 @@ func (h *Handler) GetAuthz(w http.ResponseWriter, r *http.Request) {
 	api.JSON(w, authz)
 }
 
-// GetChallenge ACME api for retrieving a Challenge.
```
dopey, line 152: Gometalinter expects exported method/function documentation to start with the name of the method -- hence `GetChallenge ...`. We can disable that check but I think it's worthwhile to force us to document exported methods.

acme/api/handler.go
```diff
@@ -184,10 +215,25 @@ func (h *Handler) GetChallenge(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	getLink := h.Auth.GetLink
-	w.Header().Add("Link", link(getLink(acme.AuthzLink, acme.URLSafeProvisionerName(prov), true, ch.GetAuthzID()), "up"))
-	w.Header().Set("Location", getLink(acme.ChallengeLink, acme.URLSafeProvisionerName(prov), true, ch.GetID()))
-	api.JSON(w, ch)
+	switch ch.Status {
+	case acme.StatusPending:
+		panic("validation attempt did not move challenge to the processing state")
```
dopey, line 220: panic? If it's an internal server error then let's return that. I don't think we want to panic.

acme/api/handler.go
```diff
@@ -184,10 +215,25 @@ func (h *Handler) GetChallenge(w http.ResponseWriter, r *http.Request) {
...
+	// Add a Retry-After header to indicate that the client should check again in the future.
+	case acme.StatusProcessing:
+		w.Header().Add("Retry-After", ch.RetryAfter)
+		w.Header().Add("Cache-Control", "no-cache")
+		api.JSON(w, ch)
+	case acme.StatusInvalid:
+		api.JSON(w, ch)
+	case acme.StatusValid:
+		getLink := h.Auth.GetLink
+		w.Header().Add("Link", link(getLink(acme.AuthzLink, acme.URLSafeProvisionerName(prov), true, ch.GetAuthzID()), "up"))
```
dopey, line 231: Why are we only setting `link` and `location` for valid challenges?

acme/api/handler.go
```diff
@@ -184,10 +215,25 @@ func (h *Handler) GetChallenge(w http.ResponseWriter, r *http.Request) {
...
+		w.Header().Set("Location", getLink(acme.ChallengeLink, acme.URLSafeProvisionerName(prov), true, ch.GetID()))
+		api.JSON(w, ch)
+	default:
+		panic("unexpected challenge state" + ch.Status)
```
dopey, line 235: don't panic

acme/api/handler_test.go
```diff
@@ -760,13 +798,21 @@ func TestHandlerGetChallenge(t *testing.T) {
...
-			} else {
+			} else if res.StatusCode >= 200 {
 				expB, err := json.Marshal(tc.ch)
 				assert.FatalError(t, err)
 				assert.Equals(t, bytes.TrimSpace(body), expB)
-				assert.Equals(t, res.Header["Link"], []string{fmt.Sprintf("<https://ca.smallstep.com/acme/authz/%s>;rel=\"up\"", tc.ch.AuthzID)})
-				assert.Equals(t, res.Header["Location"], []string{url})
 				assert.Equals(t, res.Header["Content-Type"], []string{"application/json"})
+				switch tc.ch.Status {
+				case "processing":
```
dopey, line 807: I might be misunderstanding the code, but if the status is `invalid` we still return a 200, right? So shouldn't we have another case here? Not really sure what you'd verify -- maybe that those fields aren't set? But just having the case with a "do nothing" will help to clarify.

acme/authority.go
```diff
@@ -54,8 +57,24 @@ var (
 	orderTable             = []byte("acme_orders")
 	ordersByAccountIDTable = []byte("acme_account_orders_index")
 	certTable              = []byte("acme_certs")
+	ordinal                int
 )
 
+// Ordinal is used during challenge retries to indicate ownership.
```
dopey, line 63: We should document what the init method does (what is it's purpose) above the method. Is it meant to initialize an authority? Is it meant to initialize just the "ordinal"?

Note: we generally don't document unexported methods unless they are doing something worth documenting. However, we do ty to use descriptive names. `init` could refer to a lot of things.

acme/authority.go
```diff
@@ -54,8 +57,24 @@ var (
...
+func init() {
+	ordstr := os.Getenv("STEP_CA_ORDINAL")
```
dopey, line 65: Leaving myself a note to discuss the env var name. Are we using `STEP_CA_xxx` in other places? I know that we use `STEPxxx`.

acme/authority.go
```diff
@@ -54,8 +57,24 @@ var (
...
+	if ordstr == "" {
+		ordinal = 0
+	} else {
+		ord, err := strconv.Atoi(ordstr)
+		if err != nil {
+			log.Fatal("Unrecognized ordinal ingeter value.")
+			panic(nil)
```
dopey, line 72: why are we panicking? This method should return an error and be handled just like any other error.

If we'd like to switch to using more panic's we can have that discussion. That would be a pretty significant change and I think we'd rather handle errors in a uniform way. If anything, that can be part of another PR.

acme/authority.go
```diff
@@ -254,32 +273,187 @@ func (a *Authority) GetAuthz(p provisioner.Interface, accID, authzID string) (*A
 	return az.toACME(a.db, a.dir, p)
 }
 
-// ValidateChallenge attempts to validate the challenge.
```
dopey, line 257: The documentation should start with `ValidateChallenge ...`

acme/authority.go
```diff
@@ -254,32 +273,187 @@ func (a *Authority) GetAuthz(p provisioner.Interface, accID, authzID string) (*A
...
+// 5. If the retry will commence, immediately update Retry.NextAttempt and save the challenge.
+//
+// Finally, if this instance is terminated, retries need to be reschedule when the instance restarts. This
+// is handled in the acme provisioner (authority/provisioner/acme.go) initialization.
+//
+// Note: the default ordinal does not need to be changed unless step-ca is running in a replicated scenario.
+//
 func (a *Authority) ValidateChallenge(p provisioner.Interface, accID, chID string, jwk *jose.JSONWebKey) (*Challenge, error) {
 	ch, err := getChallenge(a.db, chID)
+
```
dopey, line 314: Don't add a space here. The error handling should happen right underneath. For no other reason than "that's how we do it everywhere else, so let's be consistent".

acme/authority.go
```diff
@@ -254,32 +273,187 @@ func (a *Authority) GetAuthz(p provisioner.Interface, accID, authzID string) (*A
...
+	// Validate the challenge belongs to the account owned by the requester.
```
dopey, line 315: Move this documentation below the error handling.

acme/challenge.go
```diff
@@ -175,7 +184,7 @@ func (bc *baseChallenge) getError() *AError {
 
 // toACME converts the internal Challenge type into the public acmeChallenge
 // type for presentation in the ACME protocol.
-func (bc *baseChallenge) toACME(db nosql.DB, dir *directory, p provisioner.Interface) (*Challenge, error) {
```
dopey, line 178: every `toACME` method has the same interface. We can change that, but maybe that should be part of a different PR. I don't really, other than it's nice having all those methods having the same API. Even if one arg isn't used by some of the methods.

dcow: Perhaps this would be useful additional information to include in a comment and or to codify via an explicit interface that all the acme data types adopt. I removed the the parameter because it's unused, as you mentioned.

acme/authority.go
dcow: There's nothing to return and the entire program can't continue execution if this value can't be parsed. 

In general, I'm we should fail loudly and obnoxiously for anything that isn't spurious. For programmer or operator logic/configuration/usage errors, we should know immediately. For something like a read error that can happen during normal execution of the code, then yes internal server errors displayed in the UI (the request response) are appropriate.

acme/authority.go
dcow: I thought `init` was how you run go code at package import time. Or is it just any func starting with `init`?

acme/authority.go
dcow: Yup ran into this elsewhere. While I don't understand this lint rule, it's not hard to conform. Do you have any background on why this lint rule exists/is important?

acme/api/handler_test.go
dcow: Yes, I think I need to add the case for "invalid" here. Good catch. These tests pass though, so I don't think we're actually testing that path, currently.

acme/api/handler_test.go
dcow: done

acme/api/handler_test.go
dcow: But that would happen below. If the challenge is processing up here we return the challenge in the processing state with retryafter information.

acme/authority.go
dcow: Done! This looks odd to me too--dunno what I was thinking...

acme/authority.go
dcow: done

acme/authority.go
```diff
@@ -254,32 +273,187 @@ func (a *Authority) GetAuthz(p provisioner.Interface, accID, authzID string) (*A
...
+//  (a) Set Retry.Owner to this instance's ordinal (STEP_CA_ORDINAL).
+//  (b) Set Retry.NumAttempts to 0 and Retry.MaxAttempts to the desired max.
+//  (c) Set Challenge.Status to "processing"
+//  (d) Set retry_after to a time (retryInterval) in the future.
+// 2. Perform the validation attempt.
+// 3. If the validation attempt results in a challenge that is still processing, schedule a retry.
+//
+// It's possible that another request to re-attempt the challenge comes in while a retry attempt is
+// pending from a previous request. In general, these old attempts will see that Retry.NextAttempt
+// is in the future and drop their task. But this also might have happened on another instance, etc.
```
dcow, line 299: Mention that if it happens on another instance, the ordinal will have changed or just omit.

acme/api/handler.go
dcow: FWIW the previous comment simply had "GetChallenge" prefixed onto an otherwise complete sentence (making it grammatically incorrect) in order to appease the linter:
```
// GetChallenge ACME api for retrieving a Challenge.
```
The coherent sentence is "ACME api for retrieving a Challenge", or perhaps "GetChallenge **is the** ACME api for retrieving a Challenge".  

So while I agree that we should document functions, if people are just going to smash the func name on the front of a sentence because the linter says so, I don't understand the value of the lint rule. Maybe it's so you land on the comment instead of the func name if you search? Or maybe it's to try an enforce a certain comment voicing? Is there a place I can read up on the lint rules?

Also, does the lint rule enforce a comment on all public functions? Or just enforce that if a comment exists, it should begin with the func name?

dcow: > Also, does the lint rule enforce a comment on all public functions? Or just enforce that if a comment exists, it should begin with the func name?

It looks like the linter error message distinguishes between having a comment and the comment not being in the preferred format.

acme/api/handler.go
dcow: I think previously we would have written out the acme error. I'm not familiar with the `Link` and `Location` headers WRT ACME. Should then be returned on all valid resource gets?

acme/api/handler.go
dcow: Does golang have asserts?

acme/api/handler.go
dopey: Yeah, I think we should return these with any 2xx response.

acme/api/handler.go
dopey: Not that I know of. Our normal way of error handling is to let the errors bubble up through the stack. Your function returns an error and then the parent function gets to decide how it wants to handle the error.

acme/api/handler_test.go
dopey: yeah, I was just saying 1) it would be nice to have a case for invalid, and 2) it would be nice to have an invalid test.

acme/authority.go
dopey: No, I think that's right. An init method gets run when the package is loaded. However, it would be nice to document what the method is actually responsible for doing. Especially in cases where the name of the method/func can't be very descriptive.

acme/api/handler.go
maraino: This rule is also not as a custom configuration that we set up, it's the default and it is used all over the standard library and 3rd party libraries in Go, see for example https://golang.org/pkg/crypto/x509/#CreateCertificate

acme/authority.go
dopey: Yep, agree that there's nothing else you could be returning. This was more in reference to `panic()` being used in other places.

That said, https://github.com/smallstep/certificates/blob/master/cmd/step-ca/main.go#L143-L149 is a good example of how we should exit. Choose where the output is going and what status code to show the user.

acme/authority.go
dopey: quoting mariano: 

This rule is also not as a custom configuration that we set up, it's the default and it is used all over the standard library and 3rd party libraries in Go, see for example https://golang.org/pkg/crypto/x509/#CreateCertificate

acme/challenge.go
dopey: Eh, you're probably right. In my mind it was more like "if we're gonna change this we may as well change all the other instances too". We can change them over time though.

acme/api/handler.go
maraino: There are no asserts in the standard library, but you can create them. A C/C++ assert is basically 
```
func assert(expession bool) {
    if (expression) {
        return
    }
    panic("assertion failed")
}
```

On HTTP servers panics are recovered at top level (you can also recover them before), and they cause a 500 error, but I agree with Max and I think is better to return an error, I think our errors will also be formatted as json and might be more meaningfull than just a panic.

dcow: Right, I do that in many places. But there are a few different classes of errors. Sometimes you want to indicate that there's something horribly wrong like a logical bug in the code vs a spurious read/write error. What I'm trying to do here is signal a logical error in the code. This is an error for developer eyes not user eyes. What can a user do with a "500: Internal server error the program is not operating the way it was intended" sort of error?

acme/authority.go
maraino: In the `step` CLI, each flag has a special environment variable that we can set up, for example, if the flag is `--ca-url` you can set up `STEP_CA_URL`, in the special case where the flag has an alias, only the first alias is taken into account `--force, -f` will be only `STEP_FORCE`.

So if we have the flag `--ca-ordinal`, that we don't afaik, they will collapse.

acme/authority.go
dcow: I mention that the ordinal is used during challenge retries to indicate ownership and there's extensive commentary on a few of the methods where the meat of the usage is. Are you just looking for more context here? Or something more like "Load ordinal from the environment and store it in a var" type of comment?

acme/authority.go
maraino: This is an init, so it's ok to panic, in fact, I believe log.Fatal will panic, so it will never reach the panic, we probably want to be more explicit in the error, show, what the value was, and probably the environment variable.

dcow: So you'd prefer `log.Fatal` to `panic` here?

acme/challenge.go
dcow: I can make an interface if we want. Up to you!

acme/authority.go
maraino: Yes, it's to run code at import time.

acme/api/handler.go
dcow: I guess I was wondering why. Does golang have a blog post or anything regarding their preferred comment style, voicing, etc.? Just looking to inform myself and understand why the linter is asking for what it asks for.

acme/authority.go
maraino: I mean, it will never reach panic there, https://golang.org/src/log/log.go?s=6917:6957#L190

acme/authority.go
maraino: Instead of just `ValidateChallenge ...` to pass the linter you can use the previous command, and then write the full docs that you've written.

acme/api/handler.go
dopey: You can create your own error type, if you like, and handle it in any fine-grained way you like. I think it's better to use the existing error handling infrastructure, one that we have firm control over. 

As to what the user sees, errors from the CA are printed in the CA logs with the full stack trace. Whatever message you put in the error will appear as part of that trace. If you have a descriptive and useful message, that's what will appear.

maraino: But on an HTTP server, the user will see a 500 error, you can panic before the server initializes, on `init` for example.

acme/challenge.go
dopey: I don't think it's worth it. I thought there would be some benefit to having those methods all have the same function parameters, but I don't think there was any.

acme/api/handler.go
maraino: https://golang.org/doc/effective_go.html#commentary

acme/challenge.go
```diff
@@ -301,61 +345,61 @@ func newHTTP01Challenge(db nosql.DB, ops ChallengeOptions) (challenge, error) {
...
+	case StatusValid, StatusInvalid:
 		return hc, nil
+	default:
+		panic("unknown challenge state: " + hc.getStatus())
 	}
-	url := fmt.Sprintf("http://%s/.well-known/acme-challenge/%s", hc.Value, hc.Token)
 
+	up := &http01Challenge{hc.baseChallenge.clone()}
+
+	url := fmt.Sprintf("http://%s/.well-known/acme-challenge/%s", hc.Value, hc.Token)
```
maraino, line 363: HTTP?

acme/authority.go
dcow: We could add this all the way into the UI and make it a flag. My gut impression was that this wouldn't be an option that people use frequently but if everything is configurable via a flag then I can add that.

acme/challenge.go
maraino: Are Value and Token url safe? In not you should use `url.URL`

acme/authority.go
dcow: I meant "replace panic with log.fatal".

acme/authority.go
dcow: I mean the reason I changed it in the first place is that restating name of the function in a sentence does not documentation make. `ValidateChallenge attempts to validate the challenge.` is entirely redundant and superfluous and tells the reader nothing. So I wrote a nice explanation of what the func is doing etc. I'll write a more descriptive intro sentence that begins with ValidateChallenge.

acme/api/handler.go
dcow: > Doc comments work best as complete sentences, which allow a wide variety of automated presentations. The first sentence should be a one-sentence summary that starts with the name being declared. 

Thanks!

acme/challenge.go
```diff
@@ -573,32 +596,39 @@ func (dc *dns01Challenge) validate(db nosql.DB, jwk *jose.JSONWebKey, vo validat
 	}
 	h := sha256.Sum256([]byte(expectedKeyAuth))
 	expected := base64.RawURLEncoding.EncodeToString(h[:])
-	var found bool
+
+	if len(txtRecords) == 0 {
+		e := errors.Errorf("no TXT record found at '%s'", record)
+		up.Error = DNSErr(e).ToACME()
+		return up, nil
+	}
```
maraino, line 604: It looks to me that this check should go after we do the lookup.

acme/authority.go
dcow: Also, this would be better in `NewAuthority` so I'll move it there.

acme/authority.go
maraino: My comment was about how `step` works, I was not suggesting to add a flag, yet.

After reviewing everything, I don't like this as an environment variable, I'm going to propose a couple of solutions:
1. Add a flag to `step-ca` and pass it here, we need to be careful to not break Caddy.
2. Read a random string from the db, if it's not there, create it and use that one.

dcow: I'm currently working on adding a flag.

acme/challenge.go
dcow: This isn't new I just moved the line. I'm happy to use a URL if you think it'd be best. I'm pretty sure the token is specified to be url safe and so would be the hostname. But I don't know if we ever sanitize these things.

acme/challenge.go
dcow: I believe the lookup is here: https://github.com/smallstep/certificates/pull/242/files/5e5a76c3b53e0edc1c654f8b334e651e9b9e3b21#diff-96e563cc3503e2e7e76fad23ad6b0a09R586 (line 586). The following code is just where we try to match one of the returned records with the token.

dcow: I added this check so we could distinguish between no dns records and no _matched_ records with the idea being that no records is transient whereas an incorrect record is a failure. We could, however, just consider all mismatches transient, because it's dns.

acme/authority.go
```diff
@@ -254,32 +256,209 @@ func (a *Authority) GetAuthz(p provisioner.Interface, accID, authzID string) (*A
...
+	default:
+		e := errors.Errorf("post-validation challenge in unexpected state, %s", ch.getStatus())
+		return nil, ServerInternalErr(e)
+	}
+	return ch.toACME(a.dir, p)
+}
+
+// The challenge validation process is specific to the type of challenge (dns-01, http-01, tls-alpn-01).
+// But, we still pass generic "options" to the polymorphic validate call.
+func (a *Authority) validate(ch challenge, jwk *jose.JSONWebKey) (challenge, error) {
```
dopey, line 365: I'd like to have the method name here be `validateChallenge` or `_validateChallenge`. Otherwise it seems like it's validation of the authority itself. I appreciate that you can tell from the args, but I was still confused.

acme/authority.go
```diff
@@ -254,32 +256,209 @@ func (a *Authority) GetAuthz(p provisioner.Interface, accID, authzID string) (*A
...
+
+// RetryChallenge behaves similar to ValidateChallenge, but simply attempts to perform a validation and
+// write update the challenge record in the db if the challenge has remaining retry attempts.
+//
+// see: ValidateChallenge
+func (a *Authority) RetryChallenge(chID string) {
+	ch, err := getChallenge(a.db, chID)
 	if err != nil {
-		return nil, Wrap(err, "error attempting challenge validation")
+		return
```
dopey, line 390: Let's at least log the error here, otherwise we're failing silently, right?

acme/authority.go
```diff
@@ -254,32 +256,209 @@ func (a *Authority) GetAuthz(p provisioner.Interface, accID, authzID string) (*A
...
+	}
+	switch ch.getStatus() {
+	case StatusPending:
```
dopey, line 393: Let's add the challenge id (and any other info you think might help debugging, you may even consider printing the entire challenge object in the case of an error) to every trace/log output from this method. Debugging is going to be a pain anyway, given that our db is nosql, but it will be harder still without a challenge id.

acme/authority.go
```diff
@@ -254,32 +256,209 @@ func (a *Authority) GetAuthz(p provisioner.Interface, accID, authzID string) (*A
...
+		return
+	}
+
+	// When retrying, check to make sure the ordinal has not changed.
+	// Make sure there are still retries left.
+	// Then check to make sure Retry.NextAttempt is in the past.
+	retry := ch.getRetry()
+	switch {
+	case retry.Owner != a.ordinal:
+		return
```
dopey, line 413: We should log here, "Retry of ACME Challenge %s has changed ownership to ordinal %d" or something. Let's make it easy to understand what's happened.

acme/authority.go
```diff
@@ -254,32 +256,209 @@ func (a *Authority) GetAuthz(p provisioner.Interface, accID, authzID string) (*A
...
 	})
+}
+
+const retryInterval = 12 * time.Second
+
+// RetryChallenge behaves similar to ValidateChallenge, but simply attempts to perform a validation and
+// write update the challenge record in the db if the challenge has remaining retry attempts.
+//
+// see: ValidateChallenge
+func (a *Authority) RetryChallenge(chID string) {
```
dopey, line 387: More logging.

1) Anywhere we return due to error we should log the error and, at the very least, the challenge id.

2) Anywhere we return for logical reasons, consider logging as well.  (e.g. the ordinal has changed).

acme/authority.go
```diff
@@ -254,32 +256,209 @@ func (a *Authority) GetAuthz(p provisioner.Interface, accID, authzID string) (*A
...
+		return
+	case !retry.Active():
+		return
+	}
+	t, err := time.Parse(time.RFC3339, retry.NextAttempt)
+	now := time.Now().UTC()
+	switch {
+	case err != nil:
+		return
+	case t.Before(now):
```
dopey, line 422: Can we leave a comment about this check. Just reading the code I'm not sure why this matters. We only want to continue with a retry attempt if the retryTime is in the future?

acme/authority.go
```diff
@@ -254,32 +256,209 @@ func (a *Authority) GetAuthz(p provisioner.Interface, accID, authzID string) (*A
...
+	up := ch.clone()
+	up.Retry.NextAttempt = now.Add(retryInterval).UTC().Format(time.RFC3339)
+	up.Retry.NumAttempts++
+	err = up.save(a.db, ch)
+	if err != nil {
+		return
+	}
+	ch = up
+
+	p, err := a.LoadProvisionerByID(retry.ProvisionerID)
```
dopey, line 436: We shouldn't ignore these errors. We should log and try to exit ceremoniously.

acme/authority.go
```diff
@@ -254,32 +256,209 @@ func (a *Authority) GetAuthz(p provisioner.Interface, accID, authzID string) (*A
...
+	acc, err := a.GetAccount(p, ch.getAccountID())
```
dopey, line 437: same as above

acme/challenge.go
```diff
@@ -573,32 +602,39 @@ func (dc *dns01Challenge) validate(db nosql.DB, jwk *jose.JSONWebKey, vo validat
...
+	up.Status = StatusInvalid
+	e := errors.Errorf("keyAuthorization does not match; expected %s, but got %s",
+		expectedKeyAuth, txtRecords)
+	up.Error = IncorrectResponseErr(e).ToACME()
+	return up, nil
+}
 
-	if err := upd.save(db, dc); err != nil {
-		return nil, err
+// KeyAuthorization creates the ACME key authorization value from a token
```
dopey, line 629: This method is only used in this file. Probably doesn't need to be exported.

acme/authority.go
dcow: Looks like I need to flip this. We want to exit if next-attempt is in the future, not the past.

acme/authority.go
dcow: Is it possible for someone to remove a provisioner? If so we may just want a warning but not try to exit the entire process.

acme/challenge.go
dcow: I didn't add this, just moved it down to the bottom. I can make it private if it's not intended to be called by others.

acme/challenge.go
dcow: Based on https://letsencrypt.org/docs/challenge-types/#dns-01-challenge:

> You can have multiple TXT records in place for the same name. For instance, this might happen if you are validating a challenge for a wildcard and a non-wildcard certificate at the same time. However, you should make sure to clean up old TXT records, because if the response size gets too big Let’s Encrypt will start rejecting it.

I think my suggestion to leave the DNS challenge open: 

>  We could, however, just consider all mismatches transient, because it's dns.

might be the best course.

antonok-edm:
fwiw, I rebased this branch onto `master` at [`antonok-edm/certificates#challenge-retry`](https://github.com/antonok-edm/certificates/tree/challenge-retry) - let me know if there's anything I can do to help get something like it merged upstream

PR #243, opened 2020-05-03
Read host and protocol information from request for links

clive-jevons:
When constructing links we want to read the required host and protocol
information in a dynamic manner from the request for constructing ACME
links such as the directory information. This way, if the server is
running behind a proxy, and we don't know what the exposed URL should
be at runtime, we can construct the required information from the
host, tls and X-Forwarded-Proto fields in the HTTP request.
Inspired by the LetsEncrypt Boulder project (web/relative.go).

### Description

💔Thank you!

### DIFF
--- a/acme/api/handler.go
+++ b/acme/api/handler.go
@@ -123,7 +123,7 @@ func (h *Handler) GetDirectory(w http.ResponseWriter, r *http.Request) {
 		api.WriteError(w, err)
 		return
 	}
-	dir := h.Auth.GetDirectory(prov)
+	dir := h.Auth.GetDirectory(prov, baseURLFromRequest(r))
 	api.JSON(w, dir)
 }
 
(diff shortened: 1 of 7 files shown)


dopey:
Just took a preliminary read through and it looks really good. I'm gonna test it out against the popular acme clients today and also give some thought to the `http & https` vs `https only` question.

Will let you know if anything comes up. Thanks again for taking the time - this is awesome.

t0xicCode:
Just as some added information, given that I contributed to the original issue, the RFC specified that https is required for communication between the client and server (see section 6.1 https://tools.ietf.org/html/rfc8555#section-6.1 ).

mmalone:
I wonder if we can get the same information using SNI instead of host headers and keep end-to-end TLS?

dopey:
Hey @clive-jevons I took your PR and built off of it in https://github.com/smallstep/certificates/pull/249. The directory response was working correctly, but most of the other ACME responses also have links that needed to be updated which meant having to pass the baseURL all the way down the stack. Check out the changes and let me know if you have any questions / suggestions.

dopey:
@t0xicCode Thanks for providing the link and the added info. For the context of this PR I'm going to leave the code as mandating/assuming that TLS is being used. If people want to see a version that allows http, we can open up another issue to track that (there's already [this issue](https://github.com/smallstep/certificates/issues/246) which is kinda similar). I'll leave a note with a link to Boulder's implementation in the code, in case we come back and change this.

dopey:
Closing this, because it was merged. Feel free to re-open if necessary. Thanks again for laying down the base @clive-jevons!

PR #248, opened 2020-05-06
Initialize the required config fields on embedded authorities

maraino:
### Description

This change is to make easier the use of embedded authorities. It can be difficult for third parties to know what fields are required. The new init methods will define the minimum usable configuration.


mholt:
FWIW, here's my change downstream: https://github.com/caddyserver/caddy/commit/8dbc5f70a51aa17fa1ec9c2a34361d0b5d3ccae5#diff-2a67ba78cb7a940b54c2435310467dfb

Much nicer :+1:

Issue #122, opened 2019-10-07
Provide a default systemd unit file


kbabioch:
### What would you like to be added

Distributions are starting to pick up this package, but so far there seems to be no template unit file provided by you (= upstream), so it has to be implemented by every distribution, which is bad for packagers and users, since it might diverge.

The one I'm currently using (for Arch Linux and openSUSE) looks like this:

```
[Unit]
Description=Step CA

[Service]
ExecStart=/usr/bin/step-ca /var/lib/stepca/.step/config/ca.json --password-file /var/lib/stepca/password
User=stepca
Group=stepca

[Install]
WantedBy=multi-user.target
```

However this is only a very basic one. There are a few more caveats to consider:

- The path to the configuration file should probably be configurable, i.e. there should be an `EnvironmentFile`
- You probably only want to provide a template. Depending on the build system, it could contain some variables / macros that will be replaced. The best example I can think of right now is `wpa_supplicant` (see [here](https://w1.fi/cgit/hostap/tree/wpa_supplicant/systemd/wpa_supplicant.service.in) (although its a different build environment, i.e. not Go).

### Why this is needed

To make sure that packaging effort is kept small and that user / admin experience is the same across different distributions.

dopey:
Hey @kbabioch this is a great idea. Distribution can be a bit of a struggle but this seems like an easy win for operators.

maraino:
Just noting that the one we've used in some examples looks like:
```
[Unit]
Description=Step Certificates
Wants=basic.target
After=basic.target network.target

[Service]
WorkingDirectory=/usr/local/lib/step
ExecStart=/usr/local/bin/step-ca /usr/local/lib/step/config/ca.json --password-file /usr/local/lib/step/secrets/intermediate_pass
KillMode=process
Restart=on-failure
RestartSec=10s

[Install]
WantedBy=multi-user.target
```
See:
* https://github.com/smallstep/step-aws-emojivoto/blob/master/conf/puppet/code/environments/production/modules/step/templates/service.systemd.erb
* https://github.com/smallstep/step-aws-emojivoto/blob/master/conf/puppet/code/environments/production/modules/step/manifests/certificates.pp

darix:
the interesting part with the service file is how to support password file and interactive mode for the password

maraino:
I didn't know about systemd support for [password agents](https://www.freedesktop.org/wiki/Software/systemd/PasswordAgents/), we can take a look into it.
[...]

PR #251, opened 2020-05-07
document breakout from step to step-cli

nimbius:
### Description
Hi Smallstep
It looks like you're breaking the binaries out into step-cli and step-ca, so i updated to readme to reflect this change.  

💔Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ Certificate:
 We are using the [Arch User Repository](https://aur.archlinux.org) to distribute
 `step` binaries for Arch Linux.
 
-* [Optional] The `step` binary tarball can be found [here](https://aur.archlinux.org/packages/step-cli-bin/).
+* [Optional] The `step-cli` binary tarball can be found [here](https://aur.archlinux.org/packages/step-cli-bin/).
 * The `step-ca` binary tarball can be found [here](https://aur.archlinux.org/packages/step-ca-bin/).
 
 You can use [pacman](https://www.archlinux.org/pacman/) to install the packages.
@@ -191,7 +191,7 @@ helm install step-certificates
 
 ### Test
 
-<pre><code><b>$ step version</b>
+<pre><code><b>$ step-cli version</b>
 Smallstep CLI/0.10.0 (darwin/amd64)
 Release Date: 2019-04-30 19:01 UTC
 
@@ -213,9 +213,9 @@ communication. Let's get started!
 
 ### Let's get started!
 
-#### 1. Run `step ca init` to create your CA's keys & certificates and configure `step-ca`:
+#### 1. Run `step-cli ca init` to create your CA's keys & certificates and configure `step-ca`:
 
-<pre><code><b>$ step ca init</b>
+<pre><code><b>$ step-cli ca init</b>
 ✔ What would you like to name your new PKI? (e.g. Smallstep): <b>Example Inc.</b>
 ✔ What DNS names or IP addresses would you like to add to your new CA? (e.g. ca.smallstep.com[,1.1.1.1,etc.]): <b>localhost</b>
 ✔ What address will your new CA listen at? (e.g. :443): <b>127.0.0.1:8080</b>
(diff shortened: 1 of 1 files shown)


nimbius:
ugh, nevermind.  I have to grant CLA assistant access to my entire account for just six lines of code? no thanks.

please update your documentation when you get a chance.

dcow:
@nimbius CLAassistant is an OAuth application with a sole permission allowing read-only access to your email address. You do not grant it access to your entire account.

dopey:
Also, while we're calling the `step` CLI package `step-cli`, the actual name of the executable is `step`. The naming is confusing for sure. I think we originally decided to call the package `step-cli` because debian had an existing package called `step`.

So, for the most part, I think the documentation that you are referring to is actually correct. Just, maybe, a bit confusing. If you have any suggestions on how we can make that less confusing, we definitely want to hear them.

mmalone:
Couple things...

First, it looks like this might not be a docs problem, but a packaging one. The canonical name of the binary is actually `step`, not `step-cli`. Some of our packages I think install as `step-cli` and symlink `step` to that. I’m not sure why we’re doing that but, regardless, the canonical name is `step` and that’s the name used in all our docs and blog posts and whatnot. If it’s not possible to update the arch linux package for some reason we should probably just note that the binary name is different for that distribution in the install notes, rather than changing the binary name everywhere. Ideally we change or symlink the binary though.

Second, I understand your frustration with CLA assistant. We’ve had this come up before and, for minor stuff like this, we can accept without you agreeing to the CLA. But it is easier for us, from a governance perspective, if you do agree to the CLA. As @dcow said, it doesn’t really get much “access”, it just needs you to authenticate. But I get it if you really don’t want to.

Thanks for your help!

nimbius:
@dopey 
if it helps, i think renaming the cli might be a good idea.  Arch has step as a package as well :(
'step ca' and 'step-ca' initially are a pretty confusing pair of commands.  maybe stepctl/stepcli  and stepd?  Assuming Lennart isnt using it.  It should try to be consistent across all platforms for folks provisioning debian from arch or vice versa.  If there isnt a systemd service file for step-ca id like to contribute one as well.

@mmalone 
CLA is a 21st century devil we know i guess. Every company wants to make sure they have the rights to properly handle thorny issues like redistribution or DMCA.  its OT, but worth mentioning.  The off-putting part of the two step connection is this screen in particular, when CLA Assistant goes for more access:

auth_screen2

_write_ access to my gists, commit statuses, webhooks and services? for something that isnt even in my CI chain?

maraino:
Weird, I do cla-assistant.io only asks me for the `Personal user data - Email address (read-only)`.

dcow:
@nimbius woah I see the same thing now, too, when visiting https://cla-assistant.io/. I wouldn't grant CLA assistant those permissions just for signing something, either. When I've used it previously, I only needed to grant email read access and that's the extent of my current permissions grant:

cla-assistant

Curious, if you click this link: https://cla-assistant.io/smallstep/certificates?pullRequest=251 are you still asked to grant it all those permissions, or does it just ask you for read-only access to your email? I wonder if this is a discrepancy between signing in through their main site with the (presumed) intention to setup an integration (which would need those additional permissions) vs simply trying to view an arbitrary CLA which should only need an identifier for your signature.

nimbius:
@dcow yep, same condition.  The service wants an integration, which frankly for confirming the CLA seems a touch heavy handed for a one-time sig?  nbd though since its just read-public-stuff.  Its the second screen that starts asking for not-so-public stuff.

dcow:
@nimbius I've opened an issue with CLA assistent: https://github.com/cla-assistant/cla-assistant/issues/566. Like @mmalone said, we're happy to accept your contributions irrespective of whether you're ultimately able to sign using cla-assistent or not. 

Re: systemd, please use https://github.com/smallstep/certificates/issues/122 to discuss. I'll can add an update later about where we are in our current thought process, but we'd love to hear any suggestions you have over there so we can kick off some fresh discussion.

dcow:
@nimbius according to https://github.com/cla-assistant/cla-assistant/issues/566 the issue is indeed the route used to access cla-assistant's services. Please double check that you're using this link:

 https://cla-assistant.io/smallstep/certificates?pullRequest=251

to access the CLA. You should not be prompted for additional permissions beyond email read access.

PR #252, opened 2020-05-08
Yubikey support

maraino:
### Description

This PR adds experimental support for YubiKey, currently, only TLS certificates are supported.

Yubikey support requires CGO and it's not going to be enabled on CI builds.

Yubikey uses [go-piv/piv-go](https://github.com/go-piv/piv-go), and it requires PCSC lite support, this is available by default in macOS and Windows systems, but not on Linux, see https://github.com/go-piv/piv-go#installation for installation instructions.

### DIFF
--- a/cmd/step-yubikey-init/main.go
+++ b/cmd/step-yubikey-init/main.go
@@ -1,0 +1,289 @@
+package main
+
+import (
+	"context"
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/sha1"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"flag"
+	"fmt"
+	"math/big"
+	"os"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/smallstep/certificates/kms/apiv1"
+	"github.com/smallstep/certificates/kms/yubikey"
+	"github.com/smallstep/cli/crypto/pemutil"
+	"github.com/smallstep/cli/ui"
+	"github.com/smallstep/cli/utils"
+)
+
+type Config struct {
+	RootOnly bool
+	RootSlot string
+	CrtSlot  string
+	RootFile string
+	KeyFile  string
+	Pin      string
+}
+
+func (c *Config) Validate() error {
+	switch {
+	case c.RootFile != "" && c.KeyFile == "":
+		return errors.New("flag `--root` requires flag `--key`")
+	case c.KeyFile != "" && c.RootFile == "":
+		return errors.New("flag `--key` requires flag `--root`")
+	case c.RootOnly && c.RootFile != "":
+		return errors.New("flag `--root-only` is incompatible with flag `--root`")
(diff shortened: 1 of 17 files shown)


cmd/step-yubikey-init/main.go
```diff
@@ -0,0 +1,322 @@
...
+		ui.PrintSelected("Root Certificate", "root_ca.crt")
+	}
+
+	// Intermediate Certificate
+	var keyName string
+	var publicKey crypto.PublicKey
+	if c.RootOnly {
+		priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+		if err != nil {
+			return errors.Wrap(err, "error creating intermediate public key")
```
dopey, line 228: why "public"?

cmd/step-yubikey-init/main.go
```diff
@@ -0,0 +1,322 @@
...
+
+		ui.PrintSelected("Root Key", resp.Name)
+		ui.PrintSelected("Root Certificate", "root_ca.crt")
+	}
+
+	// Intermediate Certificate
+	var keyName string
+	var publicKey crypto.PublicKey
+	if c.RootOnly {
+		priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
dopey, line 226: Why not use `GenerateDefaultKey` from smallstep/cli here?

cmd/step-yubikey-init/main.go
```diff
@@ -0,0 +1,322 @@
...
+		NotBefore:             now,
+		NotAfter:              now.Add(time.Hour * 24 * 365 * 10),
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
+		BasicConstraintsValid: true,
+		MaxPathLen:            0,
+		MaxPathLenZero:        true,
+		Issuer:                root.Subject,
+		Subject:               pkix.Name{CommonName: "YubiKey Smallstep Intermediate"},
+		SerialNumber:          mustSerialNumber(),
+		SubjectKeyId:          mustSubjectKeyID(publicKey),
```
dopey, line 266: AuthorityKeyID?

cmd/step-yubikey-init/main.go
maraino: I want to use a supported method, in case we decide to change the default to something like Ed25519

cmd/step-yubikey-init/main.go
maraino: AuthorityKeyId is automatically created by x509.CreateCertificate, see line 2125-2128 in https://golang.org/src/crypto/x509/x509.go?s=66383:66500#L2085

AuthorityKeyId will be always rewritten for an intermediate, but I added it to the root after checking that GlobalSign and DigiCert root have it.

cmd/step-yubikey-init/main.go
maraino: Fixed in 97508ca

cmd/step-yubikey-init/main.go
maraino: Fixed in 97508ca

PR #254, opened 2020-05-08
Add identity if username contains dots.

### LINKED ISSUE #253: Allow dots in SSH usernames
We want an OIDC/ssh principal with dots in the name, so we can use firstname.lastname as a principal.

### Why this is needed
We want to use Step with OIDC for SSO ssh access. It is company policy to user firstname.lastname usernames. Step-ca doesn't allow this for OIDC/ssh, it only allows firstnamelastname and firstname.lastname@company.com

### Note
Where Debian complains about a "bad name" when adding a username with a dot, CentOS for example accepts it without warning. For POSIX Compliancy a "." is accepted but the Debian NAME_REGEX is more strict in this regard and this is reflected in the SanitizeSSHUserPrincipal function.

jvdvegt:
### Description
Before this patch, an email address like first.last@company.com would result in two identities:
firstlast
first.last@company.com

With this change, the first.last identity is also added.

Fixes #253

### DIFF
--- a/authority/provisioner/provisioner.go
+++ b/authority/provisioner/provisioner.go
@@ -253,13 +253,25 @@ func (l *List) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-var sshUserRegex = regexp.MustCompile("^[a-z][-a-z0-9_]*$")
+var strictSSHUserRegex = regexp.MustCompile("^[a-z][-a-z0-9_]*$")
+var sshUserRegex = regexp.MustCompile(`^[a-z][-a-z0-9_.]*$`)
 
 // SanitizeSSHUserPrincipal grabs an email or a string with the format
 // local@domain and returns a sanitized version of the local, valid to be used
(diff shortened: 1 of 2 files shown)


jvdvegt:
I accepted the CLA twice, but that doesn't seem to be working?

I have my email address set to private, is that causing this issue?

maraino:
I've missed this, It's a good idea, I'll take a look this coming week.

maraino:
Just noting here a different option https://github.com/smallstep/certificates/issues/253, but I like the option of adding the 3rd principal.

jvdvegt:
@maraino Your idea seems fine to me, but I can see a next feature request already: only use the local-part of the email until the '+' sign :). Though I fail to see a (business) use case for this, so please go ahead and implement the three principals you mentioned.

maraino:
@jvdvegt What do you think about replacing this with #274.

maraino:
Implemented with https://github.com/smallstep/certificates/pull/274

PR #257, opened 2020-05-08
ACME accountUpdate ignore fields not recognized by the server.

### LINKED ISSUE #194: ACME accountUpdated should succeed if unknown fields are in the request.
Currently the step-ca ACME implementation throws a MalformedError if the update request does not contain any of the fields expected by the AccountUpdateRequest endpoint. According to the [acme spec](https://tools.ietf.org/html/rfc8555#section-7.3.2) 

```
If the client wishes to update this information in the future, it
   sends a POST request with updated information to the account URL.
   The server MUST ignore any updates to the "orders" field,
   "termsOfServiceAgreed" field (see Section 7.3.3), the "status" field
   (except as allowed by Section 7.3.6), or any other fields it does not
   recognize.  If the server accepts the update, it MUST return a
   response with a 200 (OK) status code and the resulting account
   object.
```

we should be ignoring any fields we don't recognize. So the request should succeed, even though it doesn't update anything (I think).

dopey:
Fixes #194

### DIFF
--- a/acme/api/account.go
+++ b/acme/api/account.go
@@ -65,7 +65,9 @@ func (u *UpdateAccountRequest) Validate() error {
 		}
 		return nil
 	default:
-		return acme.MalformedErr(errors.Errorf("empty update request"))
+		// According to the ACME spec (https://tools.ietf.org/html/rfc8555#section-7.3.2)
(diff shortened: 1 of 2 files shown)


dopey:
A ringing endorsement if ever there was one ;)

I'm gonna merge.

PR #260, opened 2020-05-14
[Feature] Force CommonName for certificates from ACME provisioner

### LINKED ISSUE #259: [Feature Request] Support generating Subject.CN with ACME provisioner if it is empty
Make it possible to generate CommonName if it was not in the request (i.e. certbot does not send Subject within its request) with ACME protocol.

As discussed [in gitter](https://gitter.im/smallstep/community?at=5df024eef65fec1c8efb3a84), letsencrypt's boulder does it under the flag `forceCN`.

### Why this is needed

To support systems, which require `Subject` field to be non-empty. Example: Chrome on MacOS 10 does not support certificates with empty Subject.

anxolerd:
### Description
As described in #259, some systems require certificate's Subject field to be non-empty. At the same time, `certbot` does not send CommonName in its certificate request. LetsEncrypt automatically generates CN for certificate based on the first SAN in the request. It is implemented under the `forceCN` config option.

This PR duplicates the described behavior for step-ca.

Closes #259

### DIFF
--- a/acme/order.go
+++ b/acme/order.go
@@ -262,6 +262,13 @@ func (o *order) finalize(db nosql.DB, csr *x509.CertificateRequest, auth SignAut
 	if csr.Subject.CommonName != "" {
 		csr.DNSNames = append(csr.DNSNames, csr.Subject.CommonName)
 	}
+
+	// Generate Subject CommonName for supporting `conservative` systems
+	// which does not accept certificates with empty subject
+	if csr.Subject.CommonName == "" && p.(*provisioner.ACME).ForceCN {
+		csr.Subject.CommonName = csr.DNSNames[0]
+	}
+
 	csr.DNSNames = uniqueLowerNames(csr.DNSNames)
 	orderNames := make([]string, len(o.Identifiers))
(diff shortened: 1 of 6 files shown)


dopey:
Thanks for the PR @anxolerd!

We have an existing "framework" for applying modifications to new certificates that I think this feature is probably a good fit for. 

Every provisioner has a `AuthorizeSign` method that returns a number of validators and modifiers (to be applied to the end certificate), that is specific to that provisioner. [Here is that method](https://github.com/smallstep/certificates/blob/master/authority/provisioner/acme.go#L63-L75) for the ACME provisioner. In conforming with the existing framework, I think we should create a modifier (the ones already there are good examples) and add it to the list returned from this method (on the condition that `forceCN == true` for the provisioner. The `modifiers` are applied [here](https://github.com/smallstep/certificates/blob/master/authority/tls.go#L94-L97) -- `mods`. 

Let me know if those ramblings make sense.

And lastly we'll want some unit tests.

anxolerd:
@dopey, thanks for the review. I hope I understood your remarks correctly and implemented the functionality using the framework properly.

PR #268, opened 2020-05-20
Set nbf and nbf for ACME orders even when they are not set in the request.

dopey:
According to the ACME spec (https://tools.ietf.org/html/rfc8555#section-7.1.3), ACME Order responses must contain the `notBefore` and `notAfter` attributes that will later be used to set the validity bounds on the certificate. If these values are not set as part of the NewOrder request `step-ca` does not set them in the Order object stored in the database (and returned to the user). This results in Order responses that have the `NotBefore` and `NotAfter` values set to the zero timestamp (e.g. `\"notAfter\":\"0001-01-01T00:00:00Z\"`). 

`step-ca` should set `notBefore` and `notAfter` at the time of order creation to the proper values based on the inputs and the ACME provisioner claims.

### DIFF
--- a/acme/authority.go
+++ b/acme/authority.go
@@ -47,9 +47,26 @@ type Interface interface {
 
 // Authority is the layer that handles all ACME interactions.
 type Authority struct {
+	backdate provisioner.Duration
 	db       nosql.DB
 	dir      *directory
 	signAuth SignAuthority
+}
+
+// AuthorityOptions required to create a new ACME Authority.
+type AuthorityOptions struct {
+	Backdate provisioner.Duration
+	// DB is the database used by nosql.
+	DB nosql.DB
+	// DNS the host used to generate accurate ACME links. By default the authority
+	// will use the Host from the request, so this value will only be used if
+	// request.Host is empty.
+	DNS string
+	// Prefix is a URL path prefix under which the ACME api is served. This
+	// prefix is required to generate accurate ACME links.
+	// E.g. https://ca.smallstep.com/acme/my-acme-provisioner/new-account --
+	// "acme" is the prefix from which the ACME api is accessed.
+	Prefix string
 }
 
 var (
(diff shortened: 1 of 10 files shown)


acme/authority.go
```diff
@@ -47,11 +47,28 @@ type Interface interface {
...
 type Authority struct {
+	backdate *provisioner.Duration
 	db       nosql.DB
 	dir      *directory
 	signAuth SignAuthority
 }
 
+// Options required to create a new ACME Authority.
+type AuthorityOptions struct {
+	Backdate *provisioner.Duration
```
maraino, line 58: The `Backdate` is a pointer in `authority.AuthConfig` because we want the ability to be able to set it to `0` using the ca.json. By default, `nil`, we set it to one minute.

If you are using the `NewAuthority` programmatically outside of our CA and you do not set it, it will be null and it will panic. A better approach would be to use just `provisioner.Duration` that will be 0 by default.

maraino: You might also make an alias as you did with TimeDuration, it will be easier to use.

acme/authority.go
```diff
@@ -64,22 +81,22 @@ var (
 )
 
 // NewAuthority returns a new Authority that implements the ACME interface.
-func NewAuthority(db nosql.DB, dns, prefix string, signAuth SignAuthority) (*Authority, error) {
-	if _, ok := db.(*database.SimpleDB); !ok {
+func NewAuthority(ops AuthorityOptions, signAuth SignAuthority) (*Authority, error) {
```
maraino, line 84: `NewAuthority` is used by Caddy, I'll suggest to create a `New()` and just call `New()` in new authority.

And perhaps that new could have the arguments reversed, `New(a SignAuthority, opts AutorityOptions)`, looks more natural to have the options at the end.

acme/order.go
```diff
@@ -45,10 +45,12 @@ func (o *Order) GetID() string {
...
 // OrderOptions options with which to create a new Order.
 type OrderOptions struct {
-	AccountID   string       `json:"accID"`
-	Identifiers []Identifier `json:"identifiers"`
-	NotBefore   time.Time    `json:"notBefore"`
-	NotAfter    time.Time    `json:"notAfter"`
+	AccountID       string        `json:"accID"`
+	Identifiers     []Identifier  `json:"identifiers"`
+	NotBefore       TimeDuration  `json:"notBefore"`
+	NotAfter        TimeDuration  `json:"notAfter"`
```
maraino, line 51: We should use `time.Time` to avoid the extension of the spec by being able to use durations too.

authority/provisioner/provisioner.go
```diff
@@ -29,6 +30,7 @@ type Interface interface {
 	AuthorizeSSHRevoke(ctx context.Context, token string) error
 	AuthorizeSSHRenew(ctx context.Context, token string) (*ssh.Certificate, error)
 	AuthorizeSSHRekey(ctx context.Context, token string) (*ssh.Certificate, []SignOption, error)
+	DefaultTLSCertDuration() time.Duration
```
maraino, line 33: If you want to avoid adding this, you can make add a new interface acme, `acme.Provisioner`, that either extends for `provisioner.Interface` and adds that method, or perhaps, just define the methods required.

acme/authority.go
```diff
@@ -64,22 +81,32 @@ var (
 )
 
 // NewAuthority returns a new Authority that implements the ACME interface.
+// NOTE: this method is deprecated in favor of NewAuthorityOps.
```
maraino, line 84: Use this format https://github.com/golang/go/wiki/Deprecated

PR #269, opened 2020-05-20
Support for AWS IMDSv2

joshathysolate:
### Description
Modified AWS provisioner to use the IMDSv2 API for all requests.  IMDSv2 utilizes a token in the header to authorize all calls to the metadata service for added security.

IMDSv2 should always work as only IMDSv1 can be disabled.  Note that IMDS can be completely disabled or limited in scope now as well, so additional checks were added to verify the StatusCode  is not indicating an error so a more meaningful message is returned to the user.

### DIFF
--- a/authority/provisioner/aws.go
+++ b/authority/provisioner/aws.go
@@ -29,6 +29,19 @@ const awsIdentityURL = "http://169.254.169.254/latest/dynamic/instance-identity/
 // awsSignatureURL is the url used to retrieve the instance identity signature.
 const awsSignatureURL = "http://169.254.169.254/latest/dynamic/instance-identity/signature"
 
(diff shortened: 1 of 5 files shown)


joshathysolate:
Looks like I'll need to fix the utils_test.go to make the AWS mock server support the token test as well

joshathysolate:
I see what’s going on - need to fix my readURL() code as well and add a new config param to AWS object.  Will make the fixes and push changes tomorrow

joshathysolate:
Will open a new PR - I am going to add a new flag that will allow you to specify which version(s) of IMDS to attempt: v2 only, v1 only or v2 then fallback to v1

Found some references online to instances where customers may want to only use v1.

PR #270, opened 2020-05-20
Support for AWS Instance Metadata Service v2

joshathysolate:
### Description
This PR includes code to support the new IMDSv2 protocol which requires token authentication before calling any metadata-related queries.

I've included the ability to specify which versions of IMDS to use for a particular AWS provisioner so customers will have the choice to use the default (use v2 first and fall back to v1 if an error or HTTP status code of 400 or greater is returned) or use a specific version (ie: only use v1 or only use v2).  In addition, the code allows for specifying an order by using --imds-version multiple times so you can change the order from v2 fall back to v1 to v1 fall back to v2.

Existing AWS provisioners without a specific "imdsVersions" property in the JSON config files will default to v2, v1.

Note that I'll be pushing a separate PR for the step client in order to enable the new --imds-version flag when creating a provisioner.

### DIFF
--- a/authority/provisioner/aws.go
+++ b/authority/provisioner/aws.go
@@ -29,6 +29,19 @@ const awsIdentityURL = "http://169.254.169.254/latest/dynamic/instance-identity/
 // awsSignatureURL is the url used to retrieve the instance identity signature.
 const awsSignatureURL = "http://169.254.169.254/latest/dynamic/instance-identity/signature"
 
+// awsAPITokenURL is the url used to get the IMDSv2 API token
+const awsAPITokenURL = "http://169.254.169.254/latest/api/token"
+
+// awsAPITokenTTL is the default TTL to use when requesting IMDSv2 API tokens
+// -- we keep this short-lived since we get a new token with every call to readURL()
+const awsAPITokenTTL = "30"
+
+// awsMetadataTokenHeader is the header that must be passed with every IMDSv2 request
+const awsMetadataTokenHeader = "X-aws-ec2-metadata-token"
+
+// awsMetadataTokenTTLHeader is the header used to indicate the token TTL requested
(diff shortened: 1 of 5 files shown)


joshathysolate:
I thought about it but probably an edge case that it will be V1 only - saw references to this in certain K8s environments where they're proxying connections to instance metadata and those proxies may not support V2 yet.

I did test it manually by breaking the V2 response as well to make sure things worked properly whether it was V2 -> V1, V1 -> V2, V2 only or V1 only and everything checked out.

If you guys want me to add some more tests to aws_test.go and utils_test.go for the scenarios, LMK

dcow:
@joshathysolate I floated the question by the team. We'd feel most comfortable with a test for the v1 behavior if it's not crazy incremental effort.

authority/provisioner/aws.go
```diff
@@ -332,12 +366,87 @@ func (p *AWS) checkSignature(signed, signature []byte) error {
...
+		case "v1":
+			resp, err = p.readURLv1(url)
+			if err == nil && resp.StatusCode < 400 {
+				return p.readResponseBody(resp)
+			}
+		case "v2":
+			resp, err = p.readURLv2(url)
+			if err == nil && resp.StatusCode < 400 {
+				return p.readResponseBody(resp)
+			}
```
maraino, line 383: You need to close the body if err == nil and resp.StatusCode >= 400
See https://golang.org/pkg/net/http/#Response

dcow:
@joshathysolate is this issue still on your radar? If not, we can take it over. Just let us know.

joshathysolate:
Yeah, sorry I've gotten quite busy with my day job so I haven't gotten back to it but most of it is done.  I'll let you guys take it from here and finish whatever other tweaks need to be made.

PR #274, opened 2020-05-21
Allow dots and other symbols in principals for OIDC

### LINKED ISSUE #253: Allow dots in SSH usernames

maraino:
### Description
This PR will create principals with dot symbols in it when an OIDC provisioner is used. We are not changing the sanitation logic, we append the third principal if necessary:

This PR will create SSH certificates with the following principals:
```
firsstlast@example.com => ["firstlast", "firstlast@example.com"]
first.last@example.com => ["firstlast", "first.last", "first.last@example.com"]

FirstLast@example.com => ["firstlast", "FirstLlast@example.com"]
First.Last@example.com => ["firstlast", "First.Last", "First.Last@example.com"]
First+Last@example.com => ["first_last", "First+Last", "First+Last@example.com"]
```
Fixes #253, #254

### DIFF
--- a/authority/provisioner/provisioner.go
+++ b/authority/provisioner/provisioner.go
@@ -336,11 +336,24 @@ type GetIdentityFunc func(ctx context.Context, p Interface, email string) (*Iden
 func DefaultIdentityFunc(ctx context.Context, p Interface, email string) (*Identity, error) {
 	switch k := p.(type) {
 	case *OIDC:
+		// OIDC principals would be:
+		// 1. Sanitized local.
+		// 2. Raw local (if different).
+		// 3. Email address.
 		name := SanitizeSSHUserPrincipal(email)
 		if !sshUserRegex.MatchString(name) {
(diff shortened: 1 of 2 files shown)


jvdvegt:
I don't see a test nor code for one of the examples in this PR:

First+Last@example.com => ["first_last", "First+Last", "First+Last@example.com"]

(and then specifically First+Last --> first_last). Though that behavior didn't change in this PR, I think it's good to have a basic test for the SSH-username sanitation here.

Issue #209, opened 2020-03-16
Account Key Rollover RFC 8555 § 7.3.5


hbellur:
### Subject of the issue
Describe your issue here.

Directory object returns key change URL. But server returns 404 when attempting to perform key rollover
### Your environment
* OS - Linux (Ubuntu)
* Version - 16.04.6 LTS (Xenial Xerus)

dopey:
Yep, that's fair. We can remove that.

dcow:
@dopey where does it say that https://tools.ietf.org/html/rfc8555#section-7.3.5 is optional?

dopey:
@dcow we don't support it through the ACME api. So we shouldn't be returning it as part of the directory response. Not a question of whether it's optional.

dcow:
Is that conventional (I don't have tons of experience with ACME clients)? If we remove the key change url do clients know how to proceed? Or will we just get a new issue opened up with whatever error message someone sees when they try to do a key change but the url is not in the directory? Do we have an issue tracking support for § 7.3.5?

PR #276, opened 2020-05-26
acme: Return 501 for the key-change route

dcow:
RFC 8555 § 7.3.5 is not listed as optional but we do not currently
support it. Rather than 404, return a 501 to inform clients that this
functionality is not yet implemented.

The notImplmented error type is not an official error registered in the
ietf:params:acme:error namespace, so prefix if with step:acme:error. An
ACME server is allowed to return other errors and clients should display
the message detail to users.

Fixes: https://github.com/smallstep/certificates/issues/209

### DIFF
--- a/acme/api/handler.go
+++ b/acme/api/handler.go
@@ -59,6 +59,7 @@ func (h *Handler) Route(r api.Router) {
 
 	r.MethodFunc("POST", getLink(acme.NewAccountLink, "{provisionerID}", false, nil), extractPayloadByJWK(h.NewAccount))
 	r.MethodFunc("POST", getLink(acme.AccountLink, "{provisionerID}", false, nil, "{accID}"), extractPayloadByKid(h.GetUpdateAccount))
+	r.MethodFunc("POST", getLink(acme.KeyChangeLink, "{provisionerID}", false, nil, "{accID}"), extractPayloadByKid(h.NotImplemented))
 	r.MethodFunc("POST", getLink(acme.NewOrderLink, "{provisionerID}", false, nil), extractPayloadByKid(h.NewOrder))
 	r.MethodFunc("POST", getLink(acme.OrderLink, "{provisionerID}", false, nil, "{ordID}"), extractPayloadByKid(h.isPostAsGet(h.GetOrder)))
 	r.MethodFunc("POST", getLink(acme.OrdersByAccountLink, "{provisionerID}", false, nil, "{accID}"), extractPayloadByKid(h.isPostAsGet(h.GetOrdersByAccount)))
@@ -66,6 +67,7 @@ func (h *Handler) Route(r api.Router) {
 	r.MethodFunc("POST", getLink(acme.AuthzLink, "{provisionerID}", false, nil, "{authzID}"), extractPayloadByKid(h.isPostAsGet(h.GetAuthz)))
 	r.MethodFunc("POST", getLink(acme.ChallengeLink, "{provisionerID}", false, nil, "{chID}"), extractPayloadByKid(h.GetChallenge))
 	r.MethodFunc("POST", getLink(acme.CertificateLink, "{provisionerID}", false, nil, "{certID}"), extractPayloadByKid(h.isPostAsGet(h.GetCertificate)))
+
 }
 
 // GetNonce just sets the right header since a Nonce is added to each response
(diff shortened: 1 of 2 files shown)


acme/api/handler.go
```diff
@@ -59,13 +59,15 @@ func (h *Handler) Route(r api.Router) {
...
+	r.MethodFunc("POST", getLink(acme.KeyChangeLink, "{provisionerID}", false, nil, "{accID}"), extractPayloadByKid(h.NotImplemented))
 	r.MethodFunc("POST", getLink(acme.NewOrderLink, "{provisionerID}", false, nil), extractPayloadByKid(h.NewOrder))
 	r.MethodFunc("POST", getLink(acme.OrderLink, "{provisionerID}", false, nil, "{ordID}"), extractPayloadByKid(h.isPostAsGet(h.GetOrder)))
 	r.MethodFunc("POST", getLink(acme.OrdersByAccountLink, "{provisionerID}", false, nil, "{accID}"), extractPayloadByKid(h.isPostAsGet(h.GetOrdersByAccount)))
 	r.MethodFunc("POST", getLink(acme.FinalizeLink, "{provisionerID}", false, nil, "{ordID}"), extractPayloadByKid(h.FinalizeOrder))
 	r.MethodFunc("POST", getLink(acme.AuthzLink, "{provisionerID}", false, nil, "{authzID}"), extractPayloadByKid(h.isPostAsGet(h.GetAuthz)))
 	r.MethodFunc("POST", getLink(acme.ChallengeLink, "{provisionerID}", false, nil, "{chID}"), extractPayloadByKid(h.GetChallenge))
 	r.MethodFunc("POST", getLink(acme.CertificateLink, "{provisionerID}", false, nil, "{certID}"), extractPayloadByKid(h.isPostAsGet(h.GetCertificate)))
+
```
dcow, line 70: remove

acme/api/handler.go
```diff
@@ -89,6 +91,11 @@ func (h *Handler) GetDirectory(w http.ResponseWriter, r *http.Request) {
 	api.JSON(w, dir)
 }
 
+// NotImplemented returns a 501. This is a place holder for future functionality.
```
dcow, line 94: more precisely: _Semantically, this communicates a placeholder for future functionality._

acme/errors.go
```diff
@@ -398,10 +412,28 @@ func (e *Error) Cause() error {
 	return e.Err
 }
 
+// Official returns true if this error is registered with the IETF.
```
dcow, line 415: > Official returns true if this error type is registered under the acme URN namespace.

acme/api/handler.go
dopey: remove what?

dopey: Any reason to `ExtractPayloadByKid` if we're just going to respond 501? That method might return a 4xx instead of a 501.

dcow: The extra newline

acme/api/handler.go
```diff
@@ -59,13 +59,15 @@ func (h *Handler) Route(r api.Router) {
 
 	r.MethodFunc("POST", getLink(acme.NewAccountLink, "{provisionerID}", false, nil), extractPayloadByJWK(h.NewAccount))
 	r.MethodFunc("POST", getLink(acme.AccountLink, "{provisionerID}", false, nil, "{accID}"), extractPayloadByKid(h.GetUpdateAccount))
+	r.MethodFunc("POST", getLink(acme.KeyChangeLink, "{provisionerID}", false, nil, "{accID}"), extractPayloadByKid(h.NotImplemented))
```
dcow, line 62: > Any reason to ExtractPayloadByKid if we're just going to respond 501? That method might return a 4xx instead of a 501. @dopey

dcow:
@dopey is there any reason we _wouldn't_ or _shouldn't_ support it eventually? If so maybe there's a different way to communicate it. But AFAIU it's part of the spec and probably worth it to support at some point.

acme/api/handler.go
dopey: what?

dopey: Oh, sorry, were you just copying my comment to the right place?

dopey:
I'm guessing we would never support it if no one ever asks for it. It may not be a super useful feature for a private ACME server. New accounts are cheap, so maybe our stance will just be "create a new account". 

Doesn't matter, I'm fine with 501. You've left good documentation for why you chose that and if someone has a problem with it downstream they can open an issue.

acme/api/handler.go
dcow: I figure we do as much as we can until we hit the "not implemented" point. That way we don't expose anything that doesn't need to be.

dcow:
@dopey reworded the comment to make sure a 501 in no way is a guarantee of future functionality.

dopey:
Looks like travis is failing so let's make sure we get that sorted before merging.

PR #281, opened 2020-05-30
Only retain `pending' orders in the `acme_account_orders_index`

### LINKED ISSUE #273: Data too long for column 'nvalue' at acme_account_orders_index mysql backend
### Subject of the issue
I'm using acme with mysql backend

after some time I get `Data too long for column 'nvalue'`

table `acme_account_orders_index` has only one row and all new values get added to the column nvalue, and after some time it crashes. 

If I delete the row it starts working again

### Your environment
* OS - docker
* Version - smallstep/step-ca:0.14.2

### Steps to reproduce
Use mysql backend and acme, and wait some days

### Additional context
full log 

```
ca_1  | time="2020-05-21T15:27:59Z" level=error duration=100.416961ms duration-ns=100416961 error="error creating order: error storing order IDs for account xxx: failed to set acme_account_orders_index/xxxx: Error 1406: Data too long for column 'nvalue' at row 1" fields.time="2020-05-21T15:27:59Z" method=POST name=ca nonce=xxx path=/acme/acme/new-order protocol=HTTP/1.1 referer= remote-address=10.1.1.133 request-id=xxx size=285 status=500 user-agent=simp_le/0.16.0 user-id=
```

### LINKED ISSUE #280: Transaction conflict when renewing certificates
### Subject of the issue
I use step-ca as an ACME directory for other services, Caddy being the biggest consumer. I've noticed that often renewal transactions are not completing. They work just fine for a while, both issuing and renewals, but then they stop renewing.

### Your environment
* OS - CentOS 8.1.1911
* Version - 0.14.4
* Podman - 1.9.1

### Steps to reproduce
I just followed [your blog post](https://smallstep.com/blog/private-acme-server/) on how to set up the ACME server, then configured Caddy to use the url.

### Expected behaviour
Certificates needs to be renewed.

### Actual behaviour
Certificates are not renewed, with the following log entry:
```
time="2020-05-29T11:07:02+02:00" level=error duration=38.041128ms duration-ns=38041128 error="error creating order: error storing order IDs for account xxxx: failed to commit badger transaction: Transaction Conflict. Please retry" fields.time="2020-05-29T11:07:02+02:00" method=POST name=ca nonce=xxxx path=/acme/acme/new-order protocol=HTTP/1.1 referer= remote-address=x.x.x.x request-id=xxxx size=228 status=500 user-agent="CertMagic Caddy/2.0.0 xenolf-acme/3.6.0 (release; linux; amd64)" user-id=
```

dopey:
According to the ACME spec "pending orders SHOULD be returned by the `GetOrders` api".
Invalid orders SHOULD not be returned. Other states, are not discussed and are therefore not necessary to return.

The `acme_account_orders_index` is a table mapping accountID to a list of orders owned by that account. Currently the list of orders is never audited -- it continues to grow with each new order. Orders are never removed from this list. 

This PR "audits" the list of orders each time the list is requested by the API. Only orders in the PENDING state are left in the list. This "should" fix errors seen by users where the byte size of the row becomes too large for the database to write.

Closes #273

### DIFF
--- a/acme/account.go
+++ b/acme/account.go
@@ -198,17 +198,46 @@ func getAccountByKeyID(db nosql.DB, kid string) (*account, error) {
 
 // getOrderIDsByAccount retrieves a list of Order IDs that were created by the
 // account.
-func getOrderIDsByAccount(db nosql.DB, id string) ([]string, error) {
-	b, err := db.Get(ordersByAccountIDTable, []byte(id))
+func getOrderIDsByAccount(db nosql.DB, accID string) ([]string, error) {
+	b, err := db.Get(ordersByAccountIDTable, []byte(accID))
 	if err != nil {
 		if nosql.IsErrNotFound(err) {
 			return []string{}, nil
 		}
-		return nil, ServerInternalErr(errors.Wrapf(err, "error loading orderIDs for account %s", id))
+		return nil, ServerInternalErr(errors.Wrapf(err, "error loading orderIDs for account %s", accID))
 	}
-	var orderIDs []string
-	if err := json.Unmarshal(b, &orderIDs); err != nil {
-		return nil, ServerInternalErr(errors.Wrapf(err, "error unmarshaling orderIDs for account %s", id))
+	var oids []string
+	if err := json.Unmarshal(b, &oids); err != nil {
+		return nil, ServerInternalErr(errors.Wrapf(err, "error unmarshaling orderIDs for account %s", accID))
 	}
-	return orderIDs, nil
+
+	// Remove any order that is not in PENDING state and update the stored list
+	// before returning.
+	//
+	// According to RFC 8555:
+	// The server SHOULD include pending orders and SHOULD NOT include orders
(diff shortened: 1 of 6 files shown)


acme/account.go
```diff
@@ -198,17 +199,46 @@ func getAccountByKeyID(db nosql.DB, kid string) (*account, error) {
...
+		if o, err = o.updateStatus(db); err != nil {
+			return nil, ServerInternalErr(errors.Wrapf(err, "error updating order %s for account %s", oid, accID))
+		}
+		if o.Status == StatusPending {
+			pendOids = append(pendOids, oid)
+		}
+	}
+	// If there have been changes to the pending orders list, then store the
+	// new list.
+	if !reflect.DeepEqual(pendOids, oids) {
```
maraino, line 236: I'm not sure about what's going on internally, but it looks like we can detect this in the previous `for` loop, or with a length check, if we can, we should avoid the use of DeepEqual.

PR #291, opened 2020-06-17
Add support for required template variables

maraino:
### Description
This PR adds support for checking template variables, to avoid `<no value>` in templates causing bad configuration.

Currently only the default `sshd_config.tpl` template requires user variables, the required variables are set in the ca.json, but the CA can backfill those values if the default template is used. Users configuring their own templates can either check the required values on the template or configure the required values in the ca.json.

With this change if we run `step` without the required variables we will see the following error:

```sh
$ step ssh config --host
required variable 'Certificate' is missing, please use `--set <key=value>` flag
```

Fixes https://github.com/smallstep/cli/issues/232

### DIFF
--- a/authority/authority.go
+++ b/authority/authority.go
@@ -202,6 +202,7 @@ func (a *Authority) init() error {
 	}
 
 	// Decrypt and load SSH keys
+	var tmplVars templates.Step
 	if a.config.SSH != nil {
 		if a.config.SSH.HostKey != "" {
 			signer, err := a.keyManager.CreateSigner(&kmsapi.CreateSignerRequest{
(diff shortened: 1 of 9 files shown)


authority/authority.go
```diff
@@ -218,6 +220,9 @@ func (a *Authority) init() error {
 			// Append public key to list of host certs
 			a.sshCAHostCerts = append(a.sshCAHostCerts, a.sshCAHostCertSignKey.PublicKey())
 			a.sshCAHostFederatedCerts = append(a.sshCAHostFederatedCerts, a.sshCAHostCertSignKey.PublicKey())
+			// Configure template variables
+			tmplVars.SSH.HostKey = a.sshCAHostCertSignKey.PublicKey()
+			tmplVars.SSH.HostFederatedKeys = append(tmplVars.SSH.HostFederatedKeys, a.sshCAHostFederatedCerts[1:]...)
```
dopey, line 225: Can you leave a comment here, and on the identical code below explaining what's going on here. Why are we only using [1:], etc.

dopey: I see that you're just moving the code around, but a bit of explanation here would still help (me).

PR #299, opened 2020-06-23
Refactor

dopey:
Some light refactoring.

### DIFF
--- a/acme/order.go
+++ b/acme/order.go
@@ -308,6 +308,16 @@ func (o *order) finalize(db nosql.DB, csr *x509.CertificateRequest, auth SignAut
 		}
 	}
 
+	if len(csr.IPAddresses) > 0 {
+		return nil, BadCSRErr(errors.Errorf("CSR contains IP Address SANs, but should only contain DNS Names"))
+	}
+	if len(csr.EmailAddresses) > 0 {
+		return nil, BadCSRErr(errors.Errorf("CSR contains Email Address SANs, but should only contain DNS Names"))
+	}
+	if len(csr.URIs) > 0 {
+		return nil, BadCSRErr(errors.Errorf("CSR contains URI SANs, but should only contain DNS Names"))
+	}
+
 	// Get authorizations from the ACME provisioner.
 	ctx := provisioner.NewContextWithMethod(context.Background(), provisioner.SignMethod)
 	signOps, err := p.AuthorizeSign(ctx, "")
--- a/authority/provisioner/jwk.go
+++ b/authority/provisioner/jwk.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/smallstep/certificates/errs"
-	"github.com/smallstep/cli/crypto/x509util"
 	"github.com/smallstep/cli/jose"
 )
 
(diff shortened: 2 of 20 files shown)


authority/provisioner/sign_options.go
```diff
@@ -195,6 +196,50 @@ func (v emailAddressesValidator) Valid(req *x509.CertificateRequest) error {
...
+	if !reflect.DeepEqual(want, got) {
+		return errors.Errorf("URIs claim failed - got %v, want %v", req.URIs, v)
+	}
+	return nil
+}
+
+func sansValidators(sans []string) []SignOption {
+	dnsNames, ips, emails, uris := x509util.SplitSANs(sans)
+	return []SignOption{dnsNamesValidator(dnsNames), emailAddressesValidator(emails),
+		ipAddressesValidator(ips), urisValidator(uris)}
```
maraino, line 222: As you're formatting this, can you format it like
```go
return []SignOption{
        dnsNamesValidator(dnsNames), 
        emailAddressesValidator(emails),
        ipAddressesValidator(ips), 
        urisValidator(uris),
}
```

authority/provisioner/jwk.go
```diff
@@ -152,19 +151,15 @@ func (p *JWK) AuthorizeSign(ctx context.Context, token string) ([]SignOption, er
...
 		profileDefaultDuration(p.claimer.DefaultTLSCertDuration()),
 		// validators
 		commonNameValidator(claims.Subject),
 		defaultPublicKeyValidator{},
-		dnsNamesValidator(dnsNames),
-		emailAddressesValidator(emails),
-		ipAddressesValidator(ips),
 		newValidityValidator(p.claimer.MinTLSCertDuration(), p.claimer.MaxTLSCertDuration()),
-	}, nil
+	}, sansValidators(claims.SANs)), nil
```
maraino, line 162: I don't think this is right, it works because SignOption is an interface, but I think the result array will look like:
```go
[]SignOption{
   newProvisionerExtensionOption(TypeJWK, p.Name, p.Key.KeyID),
   ...
   newValidityValidator(p.claimer.MinTLSCertDuration(), p.claimer.MaxTLSCertDuration()),
   []SignOption {
        dnsNamesValidator(dnsNames),
        ....
   },
}
```

authority/provisioner/x5c.go
```diff
@@ -194,21 +193,16 @@ func (p *X5C) AuthorizeSign(ctx context.Context, token string) ([]SignOption, er
...
 			claims.chains[0][0].NotBefore, claims.chains[0][0].NotAfter},
 		// validators
 		commonNameValidator(claims.Subject),
 		defaultPublicKeyValidator{},
-		dnsNamesValidator(dnsNames),
-		emailAddressesValidator(emails),
-		ipAddressesValidator(ips),
 		newValidityValidator(p.claimer.MinTLSCertDuration(), p.claimer.MaxTLSCertDuration()),
-	}, nil
+	}, sansValidators(claims.SANs)), nil
```
maraino, line 205: See my previous comment in jwk.go

authority/provisioner/sign_options.go
maraino: We might want to convert this to an actual Validator that checks all of them, so we can just add the validator in the `[]SignOption{}` array, see my previous comment int jwk.go

authority/tls.go
```diff
@@ -103,8 +103,8 @@ func (a *Authority) Sign(csr *x509.CertificateRequest, signOpts provisioner.Opti
 		}
 	}
 
-	// Certificate modifier after validation
-	for _, m := range forcedModifiers {
+	// Certificate modifiers after validation
+	for _, m := range append(forcedModifiers, provisioner.ExtraExtensionsEnforcer{}) {
```
maraino, line 107: Just add it in the initialization of the variable, line 67

go.mod
```diff
@@ -27,5 +27,6 @@ require (
 	gopkg.in/square/go-jose.v2 v2.4.0
 )
 
-//replace github.com/smallstep/cli => ../cli
+replace github.com/smallstep/cli => ../cli
+
```
maraino, line 31: Make sure to run `go mod tidy` after fixing this.

authority/provisioner/jwk.go
maraino: `sansValidators(claims.SANs)...` will work, but we might want to convert that sansValidators to a real validator so we don't need to append here and we can just add it in the array.

acme/order.go
```diff
@@ -308,6 +308,16 @@ func (o *order) finalize(db nosql.DB, csr *x509.CertificateRequest, auth SignAut
 		}
 	}
 
+	if len(csr.IPAddresses) > 0 {
+		return nil, BadCSRErr(errors.Errorf("CSR contains IP Address SANs, but should only contain DNS Names"))
+	}
```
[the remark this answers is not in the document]
dopey: See if it conflicts how? That RFC declares a new Challenge and Authorization type, instructing servers on how to validate "IP" type identifiers. `step-ca` explicitly does not accept any type of identifier, other than DNS. https://github.com/smallstep/certificates/blob/master/acme/api/order.go#L28-L32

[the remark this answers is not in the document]
maraino: Ok, if at some point we want to support IPs we'll change it.

PR #307, opened 2020-07-01
Add support for rekeying Fixes #292

### LINKED ISSUE #292: Re-keying for X509 certificates
The existing renew feature for X509 certificate doesn't enable getting renewed certificate for new key pairs. The renew function should accept json data as http body. This json should contain a field for public key. If this field this present, existing renew function should provide a new certificate for this public key.

### Why this is needed
The end devices use certificates with longer period (ex. 1year). This period is defined by the lifetime of the private key, beyond which it is believed to be not secure. Hence certificate renewal includes change of key pair and obtaining renewed certificate for the new key pair.

### Why not short lived certificate
Below are my personal view. Please correct if this is wrong
The main reason behind short lived certificate is the failure of active revocation model among existing web servers and web browsers. But in my environment i am going to address this shortcoming by properly implementing OCSP.
In my environment i have thousand of devices handling critical functionality and they use certificates for identification. With short lived certificates, each devices need to contact CA to get their certificate renewed daily. If the CA is down, this will affect operation of the multitude of devices.

dharanikumar-s:
### Description
I used the existing renew code for rekey functionality. 
Tested it. It works fine.  

💔Thank you!


authority/tls.go
```diff
@@ -199,6 +199,89 @@ func (a *Authority) Renew(oldCert *x509.Certificate) ([]*x509.Certificate, error
...
+		if err != db.ErrNotImplemented {
+			return nil, errs.Wrap(http.StatusInternalServerError, err, "authority.Renew; error storing certificate in db", opts...)
+		}
+	}
+
+	return []*x509.Certificate{serverCert, a.x509Issuer}, nil
+}
+
+// Rekey is similar to renew except that the certificate will be renewed with new key from csr argument.
+func (a *Authority) Rekey(oldCert *x509.Certificate, csr *x509.CertificateRequest) ([]*x509.Certificate, error) {
```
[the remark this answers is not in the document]
maraino: In fact, If we don't do any certificate request Rekey can just be 

```go
func (a *Authority) Rekey(oldCert *x509.Certificate, pk *crypto.PublicKey) ([]*x509.Certificate, error)
```

and call Rekey from Renew.

dharanikumar-s:
> I'm still wondering what should the API accept, a certificate request, or just a public key ...
Certificate request is required to display the proof of possession of client's private key which is not possible if we just send the public key.

maraino:
> Certificate request is required to display the proof of possession of client's private key which is not possible if we just send the public key.

I know I was talking about this yesterday with @dopey and that was the only point in favor for a CSR, we can leave it like that as it can also be extended to require a csr or a public key, but I think the `Authority.Rekey` should be 

```go
func (a *Authority) Rekey(oldCert *x509.Certificate, pk *crypto.PublicKey) ([]*x509.Certificate, error)
```

And if Renew and Rekey share the same code Renew would be:
```go
func (a *Authority) Renew(oldCert *x509.Certificate, pk *crypto.PublicKey) ([]*x509.Certificate, error) {
       return  a.Rekey(oldCert, oldCert.PublicKey)
}
```

authority/tls.go
```diff
@@ -135,92 +136,17 @@ func (a *Authority) Sign(csr *x509.CertificateRequest, signOpts provisioner.Opti
...
-
-	return []*x509.Certificate{serverCert, a.x509Issuer}, nil
+	return a.RenewOrRekey(oldCert, oldCert.PublicKey)
 }
 
-// Rekey is similar to renew except that the certificate will be renewed with new key from csr argument.
-func (a *Authority) Rekey(oldCert *x509.Certificate, csr *x509.CertificateRequest) ([]*x509.Certificate, error) {
+// Rekey is similar to renew except that the certificate will be renewed with new key.
+// Function does rekeying if a new public key is passed as pk else if same key is passed, certificate will be just renewed.
+func (a *Authority) RenewOrRekey(oldCert *x509.Certificate, pk crypto.PublicKey) ([]*x509.Certificate, error) {
```
dharanikumar-s, line 144: Is this function fine? I am calling this from Renew. Not sure if RenewOrRekey is the right name, couldn't think of anything else. And for the test case, shall i modify the existing TestAuthority_Renew to test RenewOrRekey (i.e TestAuthority_RenewOrRekey)?

authority/tls.go
```diff
@@ -180,34 +189,48 @@ func (a *Authority) Renew(oldCert *x509.Certificate) ([]*x509.Certificate, error
...
+	// Copy all extensions except:
+	//	1. Authority Key Identifier - This one might be different if we rotate the intermediate certificate
+	//					and it will cause a TLS bad certificate error.
+	//	2. Subject Key Identifier - This should be calculated for the public key passed to this function.
 	for _, ext := range oldCert.Extensions {
-		if !ext.Id.Equal(oidAuthorityKeyIdentifier) {
+		if (!ext.Id.Equal(oidAuthorityKeyIdentifier)) && (!ext.Id.Equal(oidSubjectKeyIdentifier)) {
 			newCert.ExtraExtensions = append(newCert.ExtraExtensions, ext)
 		}
+		if ext.Id.Equal(oidSubjectKeyIdentifier) {
```
dopey, line 200: This (https://github.com/smallstep/cli/blob/master/crypto/x509util/profile.go#L305-L312) snippet of code will always set a SubjectKeyID for new profiles if one is not explicitly set in the template. The SubjectKeyID can be overwritten by an ExtraExtension at the time when the template is signed, so I think all we need to do is make sure we don't append the oidSubjectKeyIdentifier to the cert extra extensions. Let's let @maraino weigh in here before making any changes, but I think that would be easier.

authority/tls.go
dopey: I think the function should be called just `Rekey`. It's cleaner. But we can let @maraino weigh in as well.

authority/tls_test.go
```diff
@@ -370,9 +370,11 @@ ZYtQ9Ot36qc=
 	}
 }
 
-func TestAuthority_Renew(t *testing.T) {
```
dopey, line 373: I like the idea of changing the `renew` tests to be `rekey` tests, because that's where all the logic lives. However, I think we should add a separate test function (`TestAuthority_Renew`) that verifies 1) if an error happens it's correctly returned, 2) on success, among other things the public key in the new cert is equivalent to the public key in the original cert. 

This doesn't have to be added as part of this PR. Just leaving a note for myself so I don't forget.

authority/tls.go
dharanikumar-s: I checked this. Yes, It is needless to append the SubjectKeyIdentifier in tls.go.

authority/tls.go
maraino: Just `Rekey`. and calling just Rekey as you did in Renew.

authority/tls.go
maraino: This is tricky because the x509 RFC recommends but does not enforce the use of that method for subject key identifiers. So at least right now we should be backward compatible with the previous implementation and keep the Subject Key Identifier as it was.

authority/tls.go
```diff
@@ -180,34 +189,48 @@ func (a *Authority) Renew(oldCert *x509.Certificate) ([]*x509.Certificate, error
...
+				Value: append([]byte{4, 20}, hash[:]...),
+			}
+			newCert.ExtraExtensions = append(newCert.ExtraExtensions, skiExtension)
+		}
 	}
 
 	leaf, err := x509util.NewLeafProfileWithTemplate(newCert, a.x509Issuer, a.x509Signer)
 	if err != nil {
-		return nil, errs.Wrap(http.StatusInternalServerError, err, "authority.Renew", opts...)
+		return nil, errs.Wrap(http.StatusInternalServerError, err, "authority.RenewOrRekey", opts...)
```
maraino, line 217: This can be perfectly just Rekey, as the logs would tell us if the error was coming from the renew or rekey method, another option would be to encapsulate all the logic in a PRIVATE method and pass the method identifier for logging.

authority/tls.go
maraino: After thinking a little bit I think we need two do different things here for renew and rekey:
1. For `Renew`, we want to keep the certificate intact, so we should not change the Subject Key Identifier
2. For `Rekey`, we should change the subject key identifier because is a way to detect all the certificates with the same key. I believe `x509util.NewLeafProfileWithTemplate` will set the SubjectKeyId if it's not set, so, for `Rekey` only, we should skip this extension and set `newCert.SubjectKeyId` to nil to force `x509util.NewLeafProfileWithTemplate` to create a new one.

A simple way to detect a Renew or Rekey, can be to pass the pk as nil in Renew and use the existing one.

authority/tls.go
dharanikumar-s: I don't understand the second option. Meanwhile i will go with just renaming to Rekey everywhere.

api/api_test.go
```diff
@@ -551,6 +552,7 @@ type mockAuthority struct {
 	root                         func(shasum string) (*x509.Certificate, error)
 	sign                         func(cr *x509.CertificateRequest, opts provisioner.Options, signOpts ...provisioner.SignOption) ([]*x509.Certificate, error)
 	renew                        func(cert *x509.Certificate) ([]*x509.Certificate, error)
+	renewOrRekey                 func(oldCert *x509.Certificate, pk crypto.PublicKey) ([]*x509.Certificate, error)
```
dopey, line 555: This should be `rekey`.

api/api_test.go
```diff
@@ -611,6 +613,13 @@ func (m *mockAuthority) Renew(cert *x509.Certificate) ([]*x509.Certificate, erro
 	return []*x509.Certificate{m.ret1.(*x509.Certificate), m.ret2.(*x509.Certificate)}, m.err
 }
 
+func (m *mockAuthority) Rekey(oldcert *x509.Certificate, pk crypto.PublicKey) ([]*x509.Certificate, error) {
+	if m.renewOrRekey != nil {
+		return m.renewOrRekey(oldcert, pk)
```
dopey, line 618: `rekey`

authority/tls.go
```diff
@@ -135,11 +137,16 @@ func (a *Authority) Sign(csr *x509.CertificateRequest, signOpts provisioner.Opti
 // Renew creates a new Certificate identical to the old certificate, except
 // with a validity window that begins 'now'.
 func (a *Authority) Renew(oldCert *x509.Certificate) ([]*x509.Certificate, error) {
+	return a.Rekey(oldCert, nil)
+}
+
+// Func is used for renewing or rekeying based on the public key passed.
```
dopey, line 143: The comment must start with `Rekey ...`

PR #316, opened 2020-07-13
Fix typos

nop33:
### Description
In `ExecStart` the user used us `smallstep` so the same user should be defined in `useradd`.

PR #318, opened 2020-07-14
Getting Started docs fixes

nop33:
_superseding https://github.com/smallstep/certificates/pull/316_

### Description
1. The user (and optionally the group) should be the same as the one create in line 212 (`$ useradd step`)
2. Since the service unit name is `step-ca` and the program it uses is also called `step-ca`, the directory to store the logs should also be called `step-ca` imo.
3. Fixed the name of the service in line 246 (`$ systemctl start smallstep`): There's no service called `smallstep`.

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -224,7 +224,7 @@ After=syslog.target network.target
 
 User=step
 Group=step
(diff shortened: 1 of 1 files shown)


nop33:
> Merging #318 into master will increase coverage by 0.04%.

wut? 👀 How is editing a docs file increasing the code coverage?!

maraino:
Thanks @nop33, I'm gonna merge this. And somehow codecov looked into old changes in .go files 🤷🏻‍♂️

PR #329, opened 2020-07-25
SSH cert templates

### LINKED ISSUE #327: Intro templates to allow SSH cert flexibility [L]
Breaking out ssh cert flexibility:

- [x] ssh functionality for JWT prov
- [x] ssh functionality for other provs
- [x] test suite for ssh
- [ ] more work for cert library
- [ ] use new host libraries
- [ ] get code review done
- [ ] incorporate feedback

maraino:
### Description

This PR adds support for SSH certificate templates. 

The PR also includes a first step into removing cli dependencies, in this PR I've removed all `cli/crypto` and `cli/jose` dependencies. A few more are still pending, but they are out of the scope of `go.step.sm/crypto`.

Fixes #327


authority/provisioner/ssh_options.go
```diff
@@ -0,0 +1,108 @@
...
+	// templates.
+	TemplateData json.RawMessage `json:"templateData,omitempty"`
+}
+
+// HasTemplate returns true if a template is defined in the provisioner options.
+func (o *SSHOptions) HasTemplate() bool {
+	return o != nil && (o.Template != "" || o.TemplateFile != "")
+}
+
+// SSHTemplateOptions generates a CertificateOptions with the template and data
```
dopey, line 43: generates a SSHCertificateOptions, even though that's pretty obvious.

authority/ssh.go
```diff
@@ -205,59 +220,78 @@ func (a *Authority) GetSSHBastion(ctx context.Context, user string, hostname str
...
+		return nil, errs.Wrap(http.StatusBadRequest, err, "authority.SignSSH")
+	}
 
 	// Set backdate with the configured value
 	opts.Backdate = a.config.AuthorityConfig.Backdate.Duration
 
 	for _, op := range signOpts {
 		switch o := op.(type) {
+		// add options to NewCertificate
+		case provisioner.SSHCertificateOptions:
```
dopey, line 240: At some point I went through and changed as many instances of `Certificate` -> `Cert` as I could find. We should choose one or the other.

dopey: If `Cert` always means certificate in our code base, then there's probably no need to use the longer version (even if conforms with golang x509/ssh). Just my opinion. Either way, we should be consistent.

authority/ssh.go
```diff
@@ -205,59 +220,78 @@ func (a *Authority) GetSSHBastion(ctx context.Context, user string, hostname str
...
 	}
 
-	// Build base certificate with the key and some random values
-	cert := &ssh.Certificate{
-		Nonce:  []byte(nonce),
-		Key:    key,
-		Serial: serial,
-	}
+	// Get actual *ssh.Certificate and continue with provisioner modifiers.
+	cert := certificate.GetCertificate()
```
dopey, line 283: this is a certTemplate, as far as I understand.

authority/ssh.go
```diff
@@ -326,15 +344,15 @@ func (a *Authority) RenewSSH(ctx context.Context, oldCert *ssh.Certificate) (*ss
 	va := now.Add(-1 * backdate)
 	vb := now.Add(duration - backdate)
 
-	// Build base certificate with the key and some random values
+	// Build base certificate with the old key.
+	// Nonce and serial will be automatically generated on signing.
 	cert := &ssh.Certificate{
```
dopey, line 349: this is a template

dopey, line 349: this is technically only a certTemplate. I think we should stick to only referring to certificates as things that are already signed.

authority/ssh.go
maraino: Cert is the real `*ssh.Certificate` that will be signed later.

authority/ssh.go
maraino: You want me to the variable name, right?

dopey: https://github.com/smallstep/certificates/blob/193d18ee21bd9a476222d55337569fe17ab4988d/authority/tls.go#L389

dopey: I think I just want to be clear around the fact that this isn't a signed cert. It's more confusing when we use `GetCertificate` because it's not actually returning a signed certificate. In this case it's not as bad. But still, it would be nice to be consistent.

authority/tls.go
```diff
@@ -372,48 +358,62 @@ func (a *Authority) Revoke(ctx context.Context, revokeOpts *RevokeOptions) error
...
-	keyPEM, err := pemutil.Serialize(profile.SubjectPrivateKey())
+	// Create initial certificate request.
+	cr, err := x509util.CreateCertificateRequest("Step Online CA", a.config.DNSNames, signer)
 	if err != nil {
-		return nil, errs.Wrap(http.StatusInternalServerError, err, "authority.GetTLSCertificate")
+		return fatal(err)
 	}
 
+	// Generate certificate directly from the certificate request.
+	certificate, err := x509util.NewCertificate(cr)
```
dopey, line 382: Same problem here. Now we have the terms `certificate`, `cert`, `template`. It's confusing. You use the `certificate` to create a `template` which you then use to create a `cert`.

dopey: I'm not saying this is necessarily actionable. Just that it's unnecessarily confusing.

authority/tls.go
```diff
@@ -372,48 +358,62 @@ func (a *Authority) Revoke(ctx context.Context, revokeOpts *RevokeOptions) error
...
+	keyPEM, err := pemutil.Serialize(priv)
 	if err != nil {
-		return nil, errs.Wrap(http.StatusInternalServerError, err,
-			"authority.GetTLSCertificate; error creating tls certificate")
+		return fatal(err)
 	}
 
-	// Get the 'leaf' certificate and set the attribute accordingly.
-	leaf, err := x509.ParseCertificate(tlsCrt.Certificate[0])
+	tlsCrt, err := tls.X509KeyPair(append(crtPEM, pem.EncodeToMemory(intermediatePEM)...), pem.EncodeToMemory(keyPEM))
```
dopey, line 412: Since we're here, let's change this to `tlsCert` and any other instances of `crt` to `cert` that we find.

authority/provisioner/ssh_options.go
maraino: Fixed in b7269b6

authority/ssh.go
maraino: I'm not changing it right now.

authority/ssh.go
maraino: Fixed in 4d375a0

authority/ssh.go
maraino: Fixed in 4d375a0

authority/tls.go
maraino: Fixed in cef0475

authority/tls.go
maraino: Fixed in cef0475?

PR #336, opened 2020-08-02
introduce docker-buildx

### LINKED ISSUE #285: Docker Image for ARM
ARM32/64 Docker image

### Why this is needed
In my particular case I want to deploy it on Raspberry Pi as container and since step-cli already has images for this architectures, would make sense to also have it for ARM

dopey:
Fixes #285

### DIFF
--- a/make/docker.mk
+++ b/make/docker.mk
@@ -1,0 +1,67 @@
+#########################################
+# Building Docker Image
+#
+# This uses a multi-stage build file. The first stage is a builder (that might
+# be large in size). After the build has succeeded, the statically linked
+# binary is copied to a new image that is optimized for size.
+#########################################
+
+docker-prepare:
+	# Ensure, we can build for ARM architecture
+	[ -f /proc/sys/fs/binfmt_misc/qemu-arm ] || docker run --rm --privileged docker/binfmt:a7996909642ee92942dcd6cff44b9b95f08dad64
+
+	# Register buildx builder
+	mkdir -p $$HOME/.docker/cli-plugins
+
+	wget -O $$HOME/.docker/cli-plugins/docker-buildx https://github.com/docker/buildx/releases/download/v0.3.1/buildx-v0.3.1.linux-amd64
+	chmod +x $$HOME/.docker/cli-plugins/docker-buildx
+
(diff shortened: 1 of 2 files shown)


make/docker.mk
```diff
@@ -0,0 +1,67 @@
...
+
+docker-prepare:
+	# Ensure, we can build for ARM architecture
+	[ -f /proc/sys/fs/binfmt_misc/qemu-arm ] || docker run --rm --privileged docker/binfmt:a7996909642ee92942dcd6cff44b9b95f08dad64
+
+	# Register buildx builder
+	mkdir -p $$HOME/.docker/cli-plugins
+
+	wget -O $$HOME/.docker/cli-plugins/docker-buildx https://github.com/docker/buildx/releases/download/v0.3.1/buildx-v0.3.1.linux-amd64
+	chmod +x $$HOME/.docker/cli-plugins/docker-buildx
```
maraino, line 17: Should we have this in .travis.yml?

make/docker.mk
```diff
@@ -0,0 +1,67 @@
...
+# Building Docker Image
+#
+# This uses a multi-stage build file. The first stage is a builder (that might
+# be large in size). After the build has succeeded, the statically linked
+# binary is copied to a new image that is optimized for size.
+#########################################
+
+docker-prepare:
+	# Ensure, we can build for ARM architecture
+	[ -f /proc/sys/fs/binfmt_misc/qemu-arm ] || docker run --rm --privileged docker/binfmt:a7996909642ee92942dcd6cff44b9b95f08dad64
```
maraino, line 11: What is `docker run --rm --privileged docker/binfmt:a7996909642ee92942dcd6cff44b9b95f08dad64`?
This will be executed if we run this on macOS.

dopey: https://www.docker.com/blog/multi-platform-docker-builds/ -- apparently in order to build docker for multiple architectures you need to register the platforms with the kernel.

make/docker.mk
dopey: I can run the `docker-prepare` from the travis.yml. Then it will get run for every branch and not just releases/master. Any reason why?

maraino: It's ok, usually preparing the environment, installing dependencies can be done in .travis.yml, but as you said, as we don't need this always it's ok.

PR #340, opened 2020-08-07
Add note about ACME concurrency limitations

zachwalton:
### Description
We've exercised the ACME server pretty well, and noticed this issue.

`newOrder` in order.go: https://github.com/smallstep/certificates/blob/1951669e13348e2634ad5772abd982cda03f3015/acme/order.go#L71

...calls `getOrderIDsByAccount`: https://github.com/smallstep/certificates/blob/1951669e13348e2634ad5772abd982cda03f3015/acme/account.go#L201

This function does an inline and non-atomic update in the DB to transition the state of expired orders and update the index of active orders by account ID, which it does by updating an array in the index table with a compare and swap without retries.

So creating orders for the same account concurrently results in (best case) compare and swap errors or (worst case) mismatched index / order state in the DB. We're using a different database backend (hoping to contribute it back), but this is the case for all DB backends supported by `smallstep/nosql` due to the architecture of the `getOrderIDsByAccount` function.

Adding a note about how we're handling this internally. Concurrency seems to be ok for all other code paths for the ACME server.

Thanks for the amazing product! ♥️

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -718,6 +718,10 @@ A few things to consider / implement when running multiple instances of `step-ca
 * Use `MySQL` DB: The default `Badger` DB cannot be read / written by more than one
 process simultaneously. The only supported DB that can support multiple instances
 is `MySQL`. See the [database documentation][4] for guidance on configuring `MySQL`.
+  * On this note, for the ACME server there are known concurrency limitations when using
+    the same account to manage multiple orders. It is recommended to generate an ephemeral
+    account keypair for every ACME order, or to ensure that multiple orders for the same
+    account are managed serially, not concurrently.
 
 * Synchronize `ca.json` across instances: `step-ca` reads all of it's
 configuration (and all of the provisioner configuration) from the `ca.json` file


dopey:
Hey, thanks for opening the PR / issue @zachwalton. I added the index table update a few months back because otherwise that table would grow to a size unsupported by badger and mysql. It was my intention to implement something that was "atomic-like", hence using cmp-and-swap w/out retries. 

It sounds like you're using the CA pretty heavily. I'm still surprised that the DB is getting into an inconsistent state - I was hoping the fail w/out retry would avoid that. But it sounds like forcing atomicity is more complex than I thought. 

How many concurrent requests were you making using the same account? Also, thanks for sleuthing around and figuring out what the root cause of the issue was!

We'll need to fix this at some point soon. I'll go ahead and open an issue to track it. We were planning on switching to a SQL backend so the timing might coincide with this fix. An even shorter term fix might be to just not try to return a list of orders that are still processing. I kind of doubt people are using that API. But maybe I'm wrong.

In the mean time I like the idea of leaving a comment. Any reason you closed the pull request?

zachwalton:
Hey!

> I'm still surprised that the DB is getting into an inconsistent state

I think that the issue is that there are several potential DB calls that update orders in the path of `getOrderIDsByAccount`, any of which can fail to compare and swap at any time. So with multiple concurrent calls to that function for the same account it can fail at various stages of completion.

> How many concurrent requests were you making using the same account? Also, thanks for sleuthing around and figuring out what the root cause of the issue was!

I think probably half a dozen. This was a customer who has automated ACME certificate generation for new DNS endpoints for their service, and their product is self-service so any number of customers can generate ACME certs at any time via their service using the same account keypair.

>  An even shorter term fix might be to just not try to return a list of orders that are still processing.

That's reasonable for us, though IIRC the RFC specifies returning the list.

> In the mean time I like the idea of leaving a comment. Any reason you closed the pull request?

Unfortunately my employer doesn't allow me to sign license agreements without going through an internal process first- hoping to do that for future contributions.

dopey:
No worries, on the license agreement. That makes sense. Would you mind if I paraphrased your addition? If you're uncomfortable with that, again, no worries.

With regards to the RFC (https://tools.ietf.org/html/rfc8555#section-7.1.2.1) -- they specify what the returned list MUST look like. But the contents of the list is a SHOULD. 🤷‍♂️ I'd prefer not to just return an empty list, but I also don't want this to be breaking for folks. 

Hopefully, we can schedule time to plan/push a fix soon.

zachwalton:
> No worries, on the license agreement. That makes sense. Would you mind if I paraphrased your addition? If you're uncomfortable with that, again, no worries.

That is ok :)

> With regards to the RFC (https://tools.ietf.org/html/rfc8555#section-7.1.2.1) -- they specify what the returned list MUST look like. But the contents of the list is a SHOULD. 🤷‍♂️ I'd prefer not to just return an empty list, but I also don't want this to be breaking for folks.

👍

> Hopefully, we can schedule time to plan/push a fix soon.

Just a thought for a short-term solution (I've looked at this code path a lot) is that you could create a new `lock` table and insert a row to use as a global lock around the entire `getOrderIDsByAccount` function, releasing it when it's done. That would obviously be a concurrency performance bottleneck at high scale but would provide concurrency safety without re-architecting things.

It still wouldn't solve for atomicity with the multiple serial DB calls in that function but that may fix the index / order state inconsistency.

(definitely not trying to be prescriptive, just throwing out ideas)

dopey:
> (definitely not trying to be prescriptive, just throwing out ideas)

No, no. That's a great idea. It's been a minute since I've had to do anything serious with database locking, so I definitely appreciate the recommendation. 

I think the main point of contention will be what you already noted: concurrency. We're planning on offering a hosted CA for free (which will basically just be out open source CA but with some HA and scaling built around it, since those are common asks). So we may favor a solution that allows for higher concurrency even if it takes a bit longer to implement in the short term. We'll have to discuss.

In the mean time, I created a bug-issue to track what you've found (#341). If you have any more recommendations / thoughts please post them there. 

Once again, I appreciate you taking the time to dig into the issue. Makes things way easier on our end :)

PR #364, opened 2020-09-02
Update Dockerfile.step-ca to match best practices

tashian:
- Use `COPY` instead of `ADD`, because `ADD` does Magical Things that we don't need (like auto-ungzipping)
- Use `WORKDIR` instead of `RUN mkdir`
- Added a `.dockerignore` file to reduce the build context size
- Added a `HEALTHCHECK` for the CA (runs every 30 seconds and reports health status in `docker ps`)
- The `CMD exec` is already a [shell form `CMD`](https://docs.docker.com/engine/reference/builder/#cmd), so it doesn't need `/bin/sh -c` (which ends up being redundant). I tested that `docker kill -s HUP` still works properly.

See https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

### DIFF
--- a/docker/Dockerfile.step-ca
+++ b/docker/Dockerfile.step-ca
@@ -1,24 +1,27 @@
 FROM golang:alpine AS builder
 
-RUN mkdir /src
-ADD . /src
+WORKDIR /src
+COPY . .
 
-RUN apk add --no-cache make git curl && \
-    cd /src && \
-    make V=1 bin/step-ca 
+RUN apk add --no-cache \
+   curl \
+   git \
+   make && \
+   make V=1 bin/step-ca
 
 FROM smallstep/step-cli:latest
 
 COPY --from=builder /src/bin/step-ca /usr/local/bin/step-ca
 
-ENV CONFIGPATH="/home/step/config/ca.json"
-ENV PWDPATH="/home/step/secrets/password"
-
 USER root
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.step-ca
```diff
@@ -1,24 +1,27 @@
...
 USER root
 RUN apk add --no-cache libcap && setcap CAP_NET_BIND_SERVICE=+eip /usr/local/bin/step-ca
 USER step
 
+ENV CONFIGPATH="/home/step/config/ca.json"
+ENV PWDPATH="/home/step/secrets/password"
+
 VOLUME ["/home/step"]
 STOPSIGNAL SIGTERM
+HEALTHCHECK CMD curl --cacert /home/step/certs/root_ca.crt -sSf https://localhost/health >/dev/null || exit 1
```
dopey, line 25: Will this work if the port is not 443?

Johannestegner: I'd say no. Also, if the certificate is not in the correct location, that will be an issue too. I would recommend using a shell script which checks the configuration for the correct port and root cert and does a http call from the script itself, (preferably without installing `jq` or similar software so that the image doesn't grow too much).

Other than that, a healthcheck is great stuff :)

tashian: Good catch. I cleared this up by using `step ca health` instead.

dopey (approved):
lgtm

PR #367, opened 2020-09-09
Support for CAS Interface and CloudCAS

maraino:
### Description

This PR adds support for the CAS interface, a Certificate Authority Service currently for only TLS certificates. The PR includes a CAS implementation for Google's CloudCAS as well as a SoftCAS that would be the default, the default implementation will use Golang's crypto or a KMS.

CloudCAS also implements a way to retrieve the root certificate from the configured certificate authority, so there is no need to configured `"root"` nor `"crt"` in the `ca.json`.

A configured ca.json for cloudCAS looks like:
```json
{
  "address": ":9000",
   "dnsNames": ["ca.smallstep.com"],
   "cas": {
      "type": "cloudCAS",
      "credentialsFile": "/path/to/credentials.json",
      "certificateAuthority": "projects/<name>/locations/<loc>/certificateAuthorities/<intermediate-name>"
  }, 
  "..."
}
```


db/db.go
```diff
@@ -187,6 +188,19 @@ func (db *DB) RevokeSSH(rci *RevokedCertificateInfo) error {
 	}
 }
 
+// GetCertificate retrieves a certificate by the serial number.
+func (db *DB) GetCertificate(serialNumber string) (*x509.Certificate, error) {
+	ans1Data, err := db.Get(certsTable, []byte(serialNumber))
```
dopey, line 193: asn1Data

cas/apiv1/extension.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/pkg/errors"
+)
+
+var (
+	oidStepRoot                 = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 37476, 9000, 64}
+	oidStepCertificateAuthority = append(asn1.ObjectIdentifier(nil), append(oidStepRoot, 2)...)
+)
+
+// CertificateAuthorityExtension is type used to encode the certificate
```
dopey, line 16: is type -> type is

db/db.go
maraino: Fixed in 8381e9b

cas/apiv1/extension.go
maraino: Fixed in 8381e9b

authority/tls.go
```diff
@@ -340,7 +352,28 @@ func (a *Authority) Revoke(ctx context.Context, revokeOpts *RevokeOptions) error
...
 		err = a.db.RevokeSSH(rci)
-	} else { // default to revoke x509
+	} else {
+		// Revoke an X.509 certificate using CAS. If the certificate is not
+		// provided we will try to read it from the db.
+		var revokedCert *x509.Certificate
+		if revokeOpts.Crt != nil {
+			revokedCert = revokeOpts.Crt
+		} else if rci.Serial != "" {
+			revokedCert, _ = a.db.GetCertificate(rci.Serial)
```
dopey, line 362: Assuming we need to throw an error here?

dopey: if GetCertificate fails

maraino: The certificate is might be only required by specific CAS implementations. CloudCAS requires it, and it will throw an error if the the certificate is nil, but SoftCAS doesn't care.

maraino: Add comment in d644274

dopey (approved):
lgtm

PR #374, opened 2020-09-18
Create a hash of a token if a token id is empty.

### LINKED ISSUE #373: If a JWT has no Nonce or ID then step-ca should use the JWT fingerprint as the ID when storing to the DB
Current behavior is that we 4xx or 5xx due to an error writing to the DB

maraino:
### Description

We've found some OIDC tokens without a `nonce`, a `nonce` is currently used as token identifier to detect used tokens. This PR will check for empty values, and use a SHA256 of the if the provisioner returned token id is empty.

Fixes #373

### DIFF
--- a/authority/authorize.go
+++ b/authority/authorize.go
@@ -2,7 +2,9 @@ package authority
 
 import (
 	"context"
+	"crypto/sha256"
 	"crypto/x509"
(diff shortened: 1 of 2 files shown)


dopey (approved):
lgtm

PR #375, opened 2020-09-21
Use new admin template for K8ssa and admin-OIDC provisioners.

maraino:
### Description

This change replaces the .Insecure.CR template to one that sets
all the SANs, but uses key usages and extended key usages for
regular TLS certificates.


dopey (approved):
lgtm

PR #402, opened 2020-10-20
Add support for CloudCAS on step ca init

maraino:
### Description

This PR adds support for CloudCAS in step ca init.


dopey (approved):
lgtm

PR #403, opened 2020-10-20
[acme] Use lock for ordersByAccID and type to house methods

dopey:


acme/order.go
```diff
@@ -16,6 +17,9 @@ import (
 
 var defaultOrderExpiry = time.Hour * 24
 
+// Mutex for locking ordersByAccount index operations.
+var ordersByAccountMux = &sync.Mutex{}
```
maraino, line 21: Not important but, `sync.Mutex` is usually used as a value. This should be equivalent.

```go
var orderByAccountMux sync.Mutex
```

acme/order.go
```diff
@@ -111,17 +115,81 @@ func newOrder(db nosql.DB, ops OrderOptions) (*order, error) {
...
+	}
+	return o, nil
+}
+
+type orderIDsByAccount struct{}
+
+// addOrderID adds an order ID to a users index of in progress order IDs.
+// This method will also cull any orders that are no longer in the `pending`
+// state from the index before returning it.
+func (oiba orderIDsByAccount) addOrderID(db nosql.DB, accID string, oid string) ([]string, error) {
```
maraino, line 131: It's curious the use of an empty type for these two methods. Any reason for this?

maraino (approved):
LGTM but I added a couple of comments

acme/order.go
dopey: I figured these would be methods on the authDB once I make that change. This is sort of staging that.

PR #415, opened 2020-11-03
Add support for using ssh-agent as a KMS

glance-:
This adds a new KMS, SSHAgentKMS, which is a KMS to provide signing keys
for issuing ssh certificates signed by a key managed by a ssh-agent. It
uses the golang.org/x/crypto package to get a native Go implementation
to talk to a ssh-agent.

This was primarly written to be able to use gpg-agent to provide the
keys stored in a YubiKeys openpgp interface, but can be used for other
setups like proxying a ssh-agent over network.

That way the signing key for ssh certificates can be kept in a
"sign-only" hsm.

This code was written for my employer Intinor AB, but for simplicity
sake gifted to me to contribute upstream.

Signed-off-by: Anton Lundin <glance@acc.umu.se>


maraino:
Oh nice, back in April I also did a POC, I guess that would have been very helpful
Here it is https://gist.github.com/maraino/ea3a8015f655eaf106e42e8f0c48c5da

I don't know about the status of that POC, but I remember seeing differences in the way the agent was signing, but I don't remember exactly what.

mmalone:
@glance- this is awesome! Thanks so much to you and your employer for the contribution.

maraino:
@glance- I've had time today to review, and test your changes, and I've merged them. 
I've made some minor changes in the comments, and I'm returning an error instead of using log.Fatal.

There's a simple way to get rid of the changes in authority.go, but for the moment I've decided to leave it as it is.

maraino (commented):
👍

PR #436, opened 2020-12-17
Upgrade crypto and validate token id

maraino:
### Description

This PR upgrades crypto dependency with template fix, and validates the token id.


dopey (approved):
lgtm

PR #438, opened 2020-12-18
Fix broken ValidateChallenge test

dopey:

### DIFF
--- a/acme/authority_test.go
+++ b/acme/authority_test.go
@@ -2,9 +2,14 @@ package acme
 
 import (
 	"context"
+	"crypto"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"net/url"
+	"strings"
 	"testing"
 	"time"
 
@@ -1331,11 +1336,14 @@ func TestAuthorityValidateChallenge(t *testing.T) {
 	prov := newProv()
 	ctx := context.WithValue(context.Background(), ProvisionerContextKey, prov)
 	ctx = context.WithValue(ctx, BaseURLContextKey, "https://test.ca.smallstep.com:8080")
+
 	type test struct {
 		auth      *Authority
 		id, accID string
 		err       *Error
 		ch        challenge
+		jwk       *jose.JSONWebKey
+		server    *httptest.Server
 	}
 	tests := map[string]func(t *testing.T) test{
 		"fail/getChallenge-error": func(t *testing.T) test {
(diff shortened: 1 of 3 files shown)


acme/authority_test.go
```diff
@@ -1375,8 +1383,24 @@ func TestAuthorityValidateChallenge(t *testing.T) {
...
 		},
 		"fail/validate-error": func(t *testing.T) test {
-			ch, err := newHTTPCh()
+			keyauth := "temp"
+			keyauthp := &keyauth
+			// Create test server that returns challenge auth
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				fmt.Fprintf(w, "%s\r\n", *keyauthp)
+			}))
+
```
maraino, line 1392: t.Cleanup(func() { ts.Close() })

acme/authority_test.go
```diff
@@ -1375,8 +1383,24 @@ func TestAuthorityValidateChallenge(t *testing.T) {
...
+			assert.FatalError(t, err)
+
+			jwk, _, err := jose.GenerateDefaultKeyPair([]byte("pass"))
+			assert.FatalError(t, err)
+
+			thumbprint, err := jwk.Thumbprint(crypto.SHA256)
 			assert.FatalError(t, err)
+			encPrint := base64.RawURLEncoding.EncodeToString(thumbprint)
+			*keyauthp = fmt.Sprintf("%s.%s", ch.getToken(), encPrint)
+
```
maraino, line 1403: Can you just move this before httptest.NewService so you don't need to create a temporary value and all that.

maraino (changes requested):
Added a couple of comments, it would be nice to have another ok test that completes the full flow with a server, when I was debugging yesterday, the validation was returning here:
https://github.com/smallstep/certificates/blob/bae209741fc402652e96d54ba3e639027d183a45/acme/challenge.go#L306-L308

dopey:
With regards to adding another tests that does the full flow through Validate(), that's not really what this test is for. This suite is testing ValidateChallenge() - it checks what happens when ch.Validate() returns true, false, error, etc. Testing the flows inside ch.Validate() (the http, dns, and tls cases) is done by other test suites specific to those challenges.

acme/authority_test.go
dopey: It's kind of a chicken and egg problem. httptest.NewService creates a server and sets a URL at which it can be reached. I need that URL when I create the challenge (it's stored in the challenge). However, the http Handler needs to respond with a token that comes from the challenge. So the challenge needs the server to be created, and the server needs the challenge to be created. So the way I worked around this is by using the pointer indirection. I'm all for a simpler solution if you can see one.

acme/authority_test.go
```diff
@@ -1467,15 +1507,18 @@ func TestAuthorityValidateChallenge(t *testing.T) {
 				}
 			} else {
 				if assert.Nil(t, tc.err) {
+					fmt.Printf("acmeCh = %+v\n", acmeCh)
```
maraino, line 1510: Debug statement

maraino (approved):
LGTM. Just remove the debug statements.

PR #440, opened 2020-12-22
Begins to fix issue 87

mkkeffeler:
### Description
This is the start of enabling a by provisioner appended certificate to the bundles that get returned. 

I need help getting this to work. I believe the best configuration for our customers would be an attribute in ca.json like this (open to renaming, of course, or moving)

```
      "provisioners": [
         {
                "options": {
                    "x509": {
                        "templateFile": "templates/certs/x509/all.tpl",
			"AppendedCertsFile": "/Users/mkeffele/foo.crt2",
```

Where the contents of ``` /Users/mkeffele/foo.crt2 ``` is the cert they want appended.

This works almost completely, and should be a completely optional thing, but I can't get it working quite. When i replace the variable with a hardcoded filepath, I can get the bundle to return that new cert. I seem to be missing something about being able to pass in the value of what is in ca.json in that JSON to this function. Help appreciated!

Tests to come after.

### DIFF
--- a/cas/apiv1/options.go
+++ b/cas/apiv1/options.go
@@ -21,7 +21,7 @@ type Options struct {
 	// `projects/*/locations/*/certificateAuthorities/*`.
 	CertificateAuthority string `json:"certificateAuthority"`
 
-	// Issuer and signer are the issuer certificate and signer used in SoftCAS.
+	// Certificate and signer are the issuer certificate,along with any other bundled certificates to be returned in the chain for consumers, and signer used in SoftCAS.
 	// They are configured in ca.json crt and key properties.
 	CertificateChain []*x509.Certificate
 	Signer           crypto.Signer `json:"-"`
--- a/cas/apiv1/options.go
+++ b/cas/apiv1/options.go
@@ -23,8 +23,8 @@ type Options struct {
 
 	// Issuer and signer are the issuer certificate and signer used in SoftCAS.
 	// They are configured in ca.json crt and key properties.
-	Issuer *x509.Certificate `json:"-"`
-	Signer crypto.Signer     `json:"-"`
+	CertificateChain []*x509.Certificate
+	Signer           crypto.Signer `json:"-"`
 
 	// IsCreator is set to true when we're creating a certificate authority. Is
 	// used to skip some validations when initializing a CertificateAuthority.
--- a/cas/softcas/softcas.go
+++ b/cas/softcas/softcas.go
@@ -35,7 +35,7 @@ type SoftCAS struct {
 // crypto.
 func New(ctx context.Context, opts apiv1.Options) (*SoftCAS, error) {
 	if !opts.IsCreator {
-		switch {			
+		switch {
 		case len(opts.CertificateChain) == 0:
 			return nil, errors.New("softCAS 'CertificateChain' cannot be nil")
 		case opts.Signer == nil:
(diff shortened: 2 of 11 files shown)


maraino:
Hi @mkkeffeler, adding a custom issuer (intermediate) per provisioner is way more complicated. I would stick to supporting multiple intermediates in ca.json's `crt`, and using the `cas/softcas/softcas.go` to support them.

maraino:
From what we talked, this:
https://github.com/smallstep/certificates/blob/3e6137110bed7baf563b9559cd54d7398801c573/authority%2Fauthority.go#L167-L170

Should become something like:
```go
options.CertificateChain, err = pemutil.ReadCertificateBundle(a.config.IntermediateCert)
if err != nil {
	return err
}
```

mkkeffeler:
successfully built, passes tests, and works as expected by appending the full content of intermediate_ca.crt

mkkeffeler:
Worth noting, the actual cert for the intermediate CA needs to be at the top of the file (aka, first in the bundle)

mkkeffeler:
This seems to solve my issue when testing locally, as i can now get a cert with a full bundle back to my offline root CA. :)

cas/softcas/softcas.go
```diff
@@ -26,26 +26,26 @@ var now = func() time.Time {
...
 }
 
 // New creates a new CertificateAuthorityService implementation using Golang or KMS
 // crypto.
 func New(ctx context.Context, opts apiv1.Options) (*SoftCAS, error) {
 	if !opts.IsCreator {
 		switch {
-		case opts.Issuer == nil:
-			return nil, errors.New("softCAS 'issuer' cannot be nil")
+		case opts.CertificateChain == nil:
```
maraino, line 39: Here is safer to check for `len(opts.CertificateChain) == 0`

cas/apiv1/options.go
```diff
@@ -23,8 +23,8 @@ type Options struct {
 
 	// Issuer and signer are the issuer certificate and signer used in SoftCAS.
 	// They are configured in ca.json crt and key properties.
-	Issuer *x509.Certificate `json:"-"`
-	Signer crypto.Signer     `json:"-"`
+	CertificateChain []*x509.Certificate
```
maraino, line 26: Fix comment, Issuer doesn't exist anymore.

maraino (changes requested):
Hi @mkkeffeler Looks great, I've just added a couple of comments to address. 

Regarding your previous comment:
> Worth noting, the actual cert for the intermediate CA needs to be at the top of the file (aka, first in the bundle)

This is the expected behavior, it's possible to validate it, but we were not doing it before. It's probably worth doing it, as well as making sure the crt matches the private key. But there's no need to add it in this PR.

Have you tested it with multiple intermediates?

mkkeffeler:
What do you mean multiple? I have tested actually returning a bundle, and a single cert, both are working. I tested it against multiple provisioners, meaning JWK and ACME.

maraino:
I meant using two intermediates, but I think you did.

mkkeffeler:
correct, I tested using 2 intermediates being returned in the bundle, as well as with 2 provisioners. your code update requests are completed as well :)

mkkeffeler (commented):
Perfect.

maraino:
@mkkeffeler Looks good to me, but there are some formatting issues, can you run `make fmt` and push again.

maraino (approved):
Thanks @mkkeffeler LGTM

PR #457, opened 2021-01-27
Add support for PKCS #11 KMS.

### LINKED ISSUE #301: Support for YubiHSM2
### Description

Add support to be able to sign x509 and SSH certificates using keys in a YubiHSM2. If possible we should use a PKCS#11 interface to interoperability with other HSMs.

The initial scope is to add YubiHSM2/PKCS#11 as a CA KMS, not to add support for the CLI yet, that will be part of a different project.

maraino:
### Description

This PR adds support for PKCS #11 modules. 
Fixes #301

The implementation has been tested with:
- [x] YubiHSM2
- [x] SoftHSM2
- [x] Nitrokey HSM 2


maraino:
To test in an environment we need:
* YubiHSM2-SDK 
* Run `yubihsm-connector`
* Initialize the PKI using `bin/step-pkcs11-init --ssh` (requires CGO)
* Create a ca.json with the generated `root_ca.crt` and `intermediate_ca.crt`
* Set `key` to the generated uri:
```
"key": "pkcs11:id=7331;object=intermediate-key"
```
* Set `ssh` with the generated uris:
```
"ssh": {
	"hostKey": "pkcs11:id=7332;object=ssh-host-key",
	"userKey": "pkcs11:id=7333;object=ssh-user-key"
}
```
* Set `kms` like:
```
"kms": {
	"type": "pkcs11",
	"uri": "pkcs11:module-path=/usr/local/lib/pkcs11/yubihsm_pkcs11.dylib;token=YubiHSM",
	"pin": "0001password"
}
```
or 
```
"kms": {
	"type": "pkcs11",
	"uri": "pkcs11:module-path=/usr/local/lib/pkcs11/yubihsm_pkcs11.dylib;token=YubiHSM?pin-value=0001password"
}

File: ca/ca.go [hunk 1]
```diff
@@ -227,9 +227,11 @@ func (ca *CA) Reload() error {
 	}
 
 	// 1. Stop previous renewer
-	// 2. Replace ca properties
+	// 2. Close key manager
```
dopey, line 230: Change this line to something like: `Safely shutdown any internal resources (e.g. key manager)`

kms/pkcs11/pkcs11.go
```diff
@@ -0,0 +1,354 @@
...
+	"github.com/ThalesIgnite/crypto11"
+	"github.com/pkg/errors"
+	"github.com/smallstep/certificates/kms/apiv1"
+	"github.com/smallstep/certificates/kms/uri"
+)
+
+// Scheme is the scheme used in uris.
+const Scheme = "pkcs11"
+
+// DefaultRSASize is the number of bits of a new RSA key if not bitsize has been
```
dopey, line 25: if no 'bitsize' has been specified.

kms/uri/uri.go
```diff
@@ -82,5 +86,51 @@ func ParseWithScheme(scheme, rawuri string) (*URI, error) {
 // Get returns the first value in the uri with the give n key, it will return
```
dopey, line 86: with the `given` not `give n`

kms/uri/uri.go
```diff
@@ -82,5 +86,51 @@ func ParseWithScheme(scheme, rawuri string) (*URI, error) {
...
 func (u *URI) Get(key string) string {
-	return u.Values.Get(key)
+	v := u.Values.Get(key)
+	if v == "" {
+		v = u.URL.Query().Get(key)
+	}
+	return v
+}
+
+// GetEncoded returns the first value in the uri with the give n key, it will
```
dopey, line 96: 'given' not `give n'

kms/uri/uri.go
```diff
@@ -82,5 +86,51 @@ func ParseWithScheme(scheme, rawuri string) (*URI, error) {
...
+// return empty nil if that field is not present or is empty. If the return
+// value is hex encoded it will decode it and return it.
+func (u *URI) GetEncoded(key string) []byte {
+	v := u.Get(key)
+	if v == "" {
+		return nil
+	}
+	if len(v)%2 == 0 {
```
dopey, line 104: How does `len(v)%2` verify if the value is hexadecimal?

cmd/step-pkcs11-init/main.go
```diff
@@ -0,0 +1,457 @@
...
+	Pin              string
+	NoCerts          bool
+	EnableSSH        bool
+	Force            bool
+}
+
+// Validate checks the flags in the config.
+func (c *Config) Validate() error {
+	switch {
+	case c.KMS == "":
```
dopey, line 53: if `--kms` is required then why have it be an optional flag?

kms/uri/uri.go
maraino: Well, this enforces the value to be like `0A` instead of `A`, `hex.DecodeString()` will return an error if you just pass `A`, so at the end of the day we'll get the same result.

cmd/step-pkcs11-init/main.go
```diff
@@ -0,0 +1,457 @@
...
+			c.SSHHostKeyObject = ""
+			c.SSHUserKeyObject = ""
+		}
+		return nil
+	}
+}
+
+func main() {
+	var kmsuri string
+	switch runtime.GOOS {
```
dopey, line 84: Should we have a way to configure this value in case it's not in one of the default locations?

kms/uri/uri.go
dopey: What if the value was `foop` (or something with an even number of chars but not intended to be hexadecimal)? Wouldn't be error when trying to decode it? Should we be trying to decode it under those circumstances?

cmd/step-pkcs11-init/main.go
maraino: I've change the message, `flag --kms is required`. It's not an optional flag.

cmd/step-pkcs11-init/main.go
maraino: you can always pass it through `--kms pkcs11:module-path=/foo/bar/zar/yubihsm_pkcs11.so;token=YubiHSM`
or even use a different module. `--kms pkcs11:module-path=/usr/local/lib/softhsm/libsofthsm2.so;token=MySoftHSM2`

maraino: This switch picks a default one, based on the installers for each platform.

dopey (approved):
lgtm

PR #458, opened 2021-01-28
Add systemd files

tashian:
This PR includes the main service unit `step-ca.service`, and the service templates for the renewal service & timer.

### DIFF
--- a/systemd/README.md
+++ b/systemd/README.md
@@ -1,5 +1,5 @@
 ### Systemd unit files for `step-ca`
 
-For documetation on `step-ca.service`, see [Running `step-ca` As A Daemon](https://smallstep.com/docs/step-ca/certificate-authority-server-production#running-step-ca-as-a-daemon).
+For documentation on `step-ca.service`, see [Running `step-ca` As A Daemon](https://smallstep.com/docs/step-ca/certificate-authority-server-production#running-step-ca-as-a-daemon).
 
 For documentation on `cert-renewer@.*`, see [Automating Certificate Renewal](https://smallstep.com/docs/step-ca/certificate-authority-server-production#automate-x509-certificate-lifecycle-management)
--- a/systemd/README.md
+++ b/systemd/README.md
@@ -1,0 +1,5 @@
+### Systemd unit files for `step-ca`
+
+For documetation on `step-ca.service`, see [Running `step-ca` As A Daemon](https://smallstep.com/docs/step-ca/certificate-authority-server-production#running-step-ca-as-a-daemon).
+
+For documentation on `cert-renewer@.*`, see [Automating Certificate Renewal](https://smallstep.com/docs/step-ca/certificate-authority-server-production#automate-x509-certificate-lifecycle-management)
--- a/systemd/step-ca.service
+++ b/systemd/step-ca.service
@@ -48,6 +48,7 @@ RemoveIPC=true
 RestrictRealtime=true
 PrivateDevices=true
 SystemCallFilter=@system-service
+SystemCallArchitectures=native
 MemoryDenyWriteExecute=true
 ReadWriteDirectories=/etc/step-ca/db
 
--- a/systemd/step-ca.service
+++ b/systemd/step-ca.service
@@ -31,6 +31,7 @@ NoNewPrivileges=yes
 
 ; Sandboxing
 ProtectSystem=full
+ProtectHome=true
 RestrictNamespaces=true
 RestrictAddressFamilies=AF_UNIX AF_INET AF_INET6
 PrivateTmp=true
--- a/systemd/step-ca.service
+++ b/systemd/step-ca.service
@@ -30,6 +30,7 @@ SecureBits=keep-caps
 NoNewPrivileges=yes
 
 ; Sandboxing
+; This works with YubiKey PIV (via pcscd), and presumably with YubiHSM2 via http connector
 ProtectSystem=full
 ProtectHome=true
 RestrictNamespaces=true
@@ -44,8 +45,8 @@ LockPersonality=true
 RestrictSUIDSGID=true
 RemoveIPC=true
 RestrictRealtime=true
-; confirmed this works, even with YubiKey PIV, and presumably with YubiHSM2:
 PrivateDevices=true
+SystemCallFilter=@system-service
 MemoryDenyWriteExecute=true
 ReadWriteDirectories=/etc/step-ca/db
 
--- a/systemd/step-ca.service
+++ b/systemd/step-ca.service
@@ -30,7 +30,8 @@ SecureBits=keep-caps
 NoNewPrivileges=yes
 
 ; Sandboxing
-; This works with YubiKey PIV (via pcscd), and presumably with YubiHSM2 via http connector
+; This sandboxing works with YubiKey PIV (via pcscd HTTP API), but it is likely
+; too restrictive for PKCS#11 HSMs.
 ProtectSystem=full
 ProtectHome=true
 RestrictNamespaces=true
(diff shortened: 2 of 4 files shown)


systemd/step-ca.service
```diff
@@ -0,0 +1,52 @@
...
+
+; Process capabilities & privileges
+AmbientCapabilities=CAP_NET_BIND_SERVICE
+CapabilityBoundingSet=CAP_NET_BIND_SERVICE
+SecureBits=keep-caps
+NoNewPrivileges=yes
+
+; Sandboxing
+ProtectSystem=full
+RestrictNamespaces=true
```
mohe2015, line 34: Is there a reason there is no `ProtectHome=true`? https://github.com/smallstep/docs/blob/master/step-ca/certificate-authority-server-production.mdx#running-step-ca-as-a-daemon contained it and according to https://www.freedesktop.org/software/systemd/man/systemd.exec.html this should do: "If true, the directories /home/, /root, and /run/user are made inaccessible and empty for processes invoked by this unit."

systemd/step-ca.service
```diff
@@ -0,0 +1,52 @@
...
+ProtectKernelTunables=true
+ProtectKernelLogs=true
+ProtectKernelModules=true
+LockPersonality=true
+RestrictSUIDSGID=true
+RemoveIPC=true
+RestrictRealtime=true
+; confirmed this works, even with YubiKey PIV, and presumably with YubiHSM2:
+PrivateDevices=true
+MemoryDenyWriteExecute=true
```
mohe2015, line 48: I think `SystemCallArchitectures=native` may prevent exploits that use syscalls from other architectures (although I don't fully understand how this works in linux). Also I would still like to discuss `SystemCallFilter=~@clock @debug @module @mount @raw-io @reboot @swap @privileged @resources @cpu-emulation @obsolete` (https://www.freedesktop.org/software/systemd/man/systemd.exec.html) because for me it seems that these are pretty obscure syscalls. In my opinion it should at least be `SystemCallFilter=~@clock @debug @module @mount @reboot @swap @resources @cpu-emulation @obsolete` (no raw-io and no privileged) @midchildan @tashian What do you both think? `@obsolete` e.g. should be possible for sure.

systemd/step-ca.service
tashian: This should be there, good call.

systemd/step-ca.service
tashian: How does `SystemCallFilter` differ from things like `ProtectClock` or `ProtectKernelModules`?

mohe2015: @tashian 
`ProtectClock` and `ProtectKernelModules` is implemented among other ways (like blocking directory access) using SystemCallFilter (see https://www.freedesktop.org/software/systemd/man/systemd.exec.html). So these two would then be duplicated which shouldn't matter (I got the list of recommended blocked syscalls from `systemd-analyze security step-ca`). The relevant part would probably be `SystemCallFilter=~@debug @mount @reboot @swap @resources @cpu-emulation @obsolete` as the others *should* already be there implicitly. I think that these would just reduce the attack surface for linux kernel exploits / breaking things (although there are no capabilities so I don't think in this case), e.g. @obsolete contains (https://github.com/systemd/systemd/blob/57353d2909e503e3e5c7e69251ba95a31e1a72ce/src/shared/seccomp-util.c#L631) historical syscalls that still are supported partially for backwards compatibility (afaik).

tashian: Hmm. How about `SystemCallFilter=@system-service`? It's simple and it covers most of the major items raised by `systemd-analyze security`.

mohe2015: @tashian Sure if it works.

midchildan: From [systemd.exec(5)](https://www.freedesktop.org/software/systemd/man/systemd.exec.html):

> System call filtering is not equally effective on all architectures. For example, on x86 filtering of network socket-related calls is not possible, due to ABI limitations — a limitation that x86-64 does not have, however. On systems supporting multiple ABIs at the same time — such as x86/x86-64 — it is hence recommended to limit the set of permitted system call architectures so that secondary ABIs may not be used to circumvent the restrictions applied to the native ABI of the system. In particular, setting SystemCallArchitectures=native is a good choice for disabling non-native ABIs.

So it seems many restrictions set by systemd service directives can't be enforced on some systems due to limitations of the underlying architecture. The main purpose of `SystemCallArchitectures` is to prevent processes from using syscalls for architecures with lax enforcement policies. This setting is effective on systems that support multiple syscall ABIs, namely x86_64 which supports x86 (32 bit) and x86_64 (64 bit) syscall ABIs.

This would be a problem if `step-ca` legitimately uses x86 syscalls for whatever reason, though I'd imagine this to be unlikely. 

> How does `SystemCallFilter` differ from things like `ProtectClock` or `ProtectKernelModules`?

I think they all use the same mechanism. `ProtectClock` and `ProtectKernelModule` is the high level configuration while `SystemCallFilter` provides more granular options.

> In my opinion it should at least be `SystemCallFilter=~@clock @debug @module @mount @reboot @swap @resources @cpu-emulation @obsolete`

Including `@resources` in the denylist would cause processes to crash when it calls `nice()`, so that seems a bit risky. The other options look okay, though it's hard to be completely sure.

midchildan: I also noticed there's yubikey support in `step-ca`, I wonder how some options would affect this functionality.

mohe2015: > > How does `SystemCallFilter` differ from things like `ProtectClock` or `ProtectKernelModules`?
> 
> I think they all use the same mechanism. `ProtectClock` and `ProtectKernelModule` is the high level configuration while `SystemCallFilter` provides more granular options.

The manpage mentions some additional hardenings like making "/dev/rtc0, /dev/rtc1" (the clocks) readonly and making `/usr/lib/modules` inaccessible.

> > In my opinion it should at least be `SystemCallFilter=~@clock @debug @module @mount @reboot @swap @resources @cpu-emulation @obsolete`
> 
> Including `@resources` in the denylist would cause processes to crash when it calls `nice()`, so that seems a bit risky. The other options look okay, though it's hard to be completely sure.

You're right, I would also be fine with using `@system-service` (which would be nice if it works as it's an allowlist). This contains `@resources`

> I also noticed there's yubikey support in step-ca, I wonder how some options would affect this functionality.

@tashian Tested this (and said it worked) but I don't own such keys so I can't test this.

tashian: > I also noticed there's yubikey support in step-ca, I wonder how some options would affect this functionality.

Yes, I've tested this and it works (including with `@system-service` syscall filtering).

So far, our hardware support only uses HTTP connector APIs (pcscd for YubiKey PIV, and [YubiHSM connector](https://developers.yubico.com/YubiHSM2/Component_Reference/yubihsm-connector/) for PKCS#11).

@maraino do you expect further HSM support might need direct hardware access via syscalls?

mohe2015: I don't know but maybe still add `SystemCallArchitectures=native` ?

systemd/README.md
```diff
@@ -0,0 +1,5 @@
+### Systemd unit files for `step-ca`
+
+For documetation on `step-ca.service`, see [Running `step-ca` As A Daemon](https://smallstep.com/docs/step-ca/certificate-authority-server-production#running-step-ca-as-a-daemon).
```
mohe2015, line 3:
```suggestion
For documentation on `step-ca.service`, see [Running `step-ca` As A Daemon](https://smallstep.com/docs/step-ca/certificate-authority-server-production#running-step-ca-as-a-daemon).
```

mohe2015 (approved):
LGTM

dopey (approved):
lgtm

PR #471, opened 2021-02-12
Note that libpcsclite is required to build from source

klingtnet:
### Description

If this package is not available `make` will fail with this error

```
ERRO Running error: buildir: failed to load package piv: could not load
export data: no export data for "github.com/go-piv/piv-go/piv"
```

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -16,7 +16,7 @@ to manage issues, etc.
 
 ## Building From Source
 
-The only prerequisites are [`go`](https://golang.org/) and make.
+The only prerequisites are [`go`](https://golang.org/), `make` and `libpcsclite` for YubiKey support.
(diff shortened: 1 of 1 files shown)


maraino:
@klingtnet This is a tricky situation, because we can simple avoid this using `CGO_ENABLED=0` in `make lint` (and since the last merge `make test`), but at the same time we want to lint/test our CGO enabled code in our build system. 

So I believe the right solution would be to put `$(GOFLAGS)` in `make lint` and restore it in `make test`, and add an alternative goal that lint, tests with CGO.

maraino:
@klingtnet I believe the PR https://github.com/smallstep/certificates/pull/474 would solve the issue.

maraino (commented):
Thanks for letting us know, I've fixed the issue and closed this PR.

PR #474, opened 2021-02-12
Permit linting and testing without CGO support.

maraino:
### Description

Do not require CGO (and dependencies like libpcsclite) when doing `make`, but require CGO for testing and linting in travis.
Fixes #471


dopey (approved):
lgtm

PR #485, opened 2021-02-18
Convert to github actions

dopey:


.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
...
+      is_prerelease: ${{ steps.is_prerelease.outputs.IS_PRERELEASE }}
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '^1.15.6'
```
maraino, line 23: I think we should build what the latest release of go ^1.16.0. If we want to keep backward compatibility with at least a version, we should do a new job that builds ands tests with older versions.

.github/workflows/test.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '^1.15.6'
+      - name: Install Deps
+        id: install-deps
+        run: sudo apt-get -y install libpcsclite-dev
+      - name: Lint, Test, Build
+        id: lintTestBuild
+        run: V=1 make -j1 bootstrap travis
```
maraino, line 27: We might want to change the `travis` goal to `ci` or something similar.

.github/workflows/test.yml
```diff
@@ -0,0 +1,34 @@
...
+  lintTestBuild:
+    name: Lint, Test, Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '^1.15.6'
```
maraino, line 21: 1.16.

maraino (commented):
LGTM but I added similar comments to the ones in the cli.

.github/workflows/test.yml
dopey: We're actually going to wait on switching the `test` job to go1.16 because there broken tests and linter warning that we'd like to fix first.

maraino (approved):
Looks good to me, but use latest Go version in 1.15 (1.15.8)

PR #502, opened 2021-03-10
Add support for cli-utils with powershell support.

maraino:
### Description

This PR updates go.step.sm/cli-utils to v0.2.0. This version sets a windows terminal to use VT codes, so colors and bold text is supported.


dopey (approved):
lgtm

PR #514, opened 2021-03-23
PKCS11 Init Pin Flag

gdbelvin:
Adds a "pin" flag to cmd/step-pkcs11-init so that it can take it as input from another program

### DIFF
--- a/cmd/step-pkcs11-init/main.go
+++ b/cmd/step-pkcs11-init/main.go
@@ -119,7 +119,7 @@ func main() {
 		fatal(err)
 	}
 
-	if u.Pin() == "" {
+	if u.Pin() == "" && c.Pin == "" {
 		pin, err := ui.PromptPassword("What is the PKCS#11 PIN?")
 		if err != nil {
 			fatal(err)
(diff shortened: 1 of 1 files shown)


maraino:
@gdbelvin I don't think this is actually doing anything, if the pin in the `--kms` URI is not set it will always ask for one.

You can pass a pin with the parameters `pin-value` or `pin-source` like this:
```sh
step-pkcs11-init --kms 'pkcs11:module-path=/usr/lib/x86_64-linux-gnu/pkcs11/yubihsm_pkcs11.so;token=YubiHSM?pin-value=0001password'
# or 
step-pkcs11-init --kms 'pkcs11:module-path=/usr/lib/x86_64-linux-gnu/pkcs11/yubihsm_pkcs11.so;token=YubiHSM?pin-source=/tmp/password.txt'
```

maraino:
This URI is based on the [RFC7515](https://tools.ietf.org/html/rfc7512)

gdbelvin:
Thanks for pointing this out

maraino:
Right now pin source must be a text file with the pin in it

maraino:
Changing your PR to actually work is just one extra condition here 
https://github.com/smallstep/certificates/blob/9146fe8055d2132ee4133143e2265e0c44d7f7eb/cmd/step-pkcs11-init/main.go#L121

gdbelvin:
Thanks, like this?

maraino (approved):
That looks good Gary.

PR #534, opened 2021-04-13
Update sshd_config.tpl to a Match all block.

### LINKED ISSUE #479: Update sshd template for hosts to use `matchall`
sshd -t revealed that the configuration Smallstep adds to sshd_config was somehow interpreted as being inside the MatchUser block (and probably the Match* family of directives)

The fix is easy, it just needs to be moved before the Match* sections. IDK if you guys wanna put that in the Host Enrollment Guide or maybe update the scripts to handle this situation. This causes the sshd daemon to not respawn, so if a user disconnects from the existing session, they will be locked out of the host.

Or probably use a prophylactic Match all

https://unix.stackexchange.com/questions/67334/openssh-how-to-end-a-match-block

maraino:
### Description
Update sshd_config.tpl to a Match all block.

Fixes #479


dopey (approved):
lgtm

PR #539, opened 2021-04-14
Use different method for unescpaed paths for the router

dopey:


acme/api/middleware.go
```diff
@@ -88,15 +87,23 @@ func (h *Handler) addDirLink(next nextHTTP) nextHTTP {
...
-		ct := r.Header.Get("Content-Type")
-		var expected []string
-		if strings.Contains(r.URL.String(), h.linker.GetLink(r.Context(), CertificateLinkType, false, "")) {
+		var (
+			expected []string
+			provName string
+		)
+		if p, err := provisionerFromContext(r.Context()); err == nil && p != nil {
+			provName = p.GetName()
+		}
```
maraino, line 96: Should we worry about an error here?

maraino (approved):
Just the question that I added about the error being ignored. Otherwise, looks good to me.

PR #540, opened 2021-04-15
Always specify a Proxy in all custom transports.

### LINKED ISSUE #535: Add support for HTTP(S)_PROXY environent variables to work behind proxy
Running the following with HTTP_PROXY and HTTPS_PROXY environment does not work behind a proxy.
```
step ssh config --team xxx
```
Inspecting the code, indicates that the issue lies in the bootstrap code.
From looking at command/ca/bootstrap.go, this is because proxy support is not implemented at this time. This is making it difficult to use smallstep's managed services.

https://golang.org/pkg/net/http/#ProxyFromEnvironment

I'm happy to write up the patch if this would be accepted. 

### Why this is needed

To allow smallstep to work behind the corporate proxies. This will allow me to use your managed offering rather than running a local CA.

maraino:
### Description
This PR changes all the custom transports being used to add `Proxy: http.ProxyFromEnvironment`

Fixes #535

### DIFF
--- a/ca/client.go
+++ b/ca/client.go
@@ -57,6 +57,7 @@ func newInsecureClient() *uaClient {
 	return &uaClient{
 		Client: &http.Client{
 			Transport: &http.Transport{
+				Proxy:           http.ProxyFromEnvironment,
 				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
 			},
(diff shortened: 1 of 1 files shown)


dopey (approved):
lgtm

PR #543, opened 2021-04-16
Remove the creation on nonce on get acme directory

maraino:
### Description
This PR removes the replay nonce on get directory request. According to RFC 8555, the replay nonces are only required in POST requests. And of course in the new-nonce request.


dopey (approved):
lgtm. But test this with our client (and acme.sh + certbot) before merging. Our client is not that well written, and god only knows what sort of craziness it relies on.

PR #544, opened 2021-04-17
Update provisioners.md

ionrover2:
"Themself" over "himself" is gender neutral. Not sure that anyone would notice or care, but just want to make it a happy place for everyone. Grammar is not my strong suite, so I won't make a fuss if this doesn't get merged.

### Description
added Gender neutral vernacular though i've gone back and fourth on the grammar (themself vs Themselves). Again, won't be upset if this doesn't get merged.

💔Thank you!


tashian:
Thanks for pointing this out. I resolved it with [this commit](https://github.com/smallstep/certificates/commit/e50c5bc4b1c4ac8e0d1da07a66f4c05b02fbb1e7)

PR #553, opened 2021-04-26
Add extension of db.AuthDB to store the fullchain

maraino:
### Description

This PR adds a temporary solution to allow an extension of a `db.AuthDB` interface that logs the full chain of certificates instead of just the leaf.


dopey (approved):
lgtm

PR #557, opened 2021-04-27
use InsecureSkipVerify for validation

### LINKED ISSUE #554: http-01 (rfc 8555 8.3) redirects to https fail validation as the certificate is invalid
### Subject of the issue

When the initial http-01 challenge is sent to port 80, it is valid to accept redirects. Often, ingress servers are configured to automatically redirect all connections to https on port 443. Since the certificate served by the server will not yet be valid until it is received via this process, the validation should set InsecureSkipVerify in http01Validate.

The RFC is somewhat conflicting in its wording:
> Because many web servers  allocate a default HTTPS virtual host to a particular low-privilege tenant user in a subtle and non-intuitive manner, the challenge must be completed over HTTP, not HTTPS.

yet

> The server SHOULD follow redirects when dereferencing the URL. Clients might use redirects, for example, so that the response can be provided by a centralized certificate management server.

Let's Encrypt helped draft that RFC and their documentation states:
> Our implementation of the HTTP-01 challenge follows redirects, up to 10 redirects deep. It only accepts redirects to “http:” or “https:”, and only to ports 80 or 443. It does not accept redirects to IP addresses. When redirected to an HTTPS URL, it does not validate certificates (since this challenge is intended to bootstrap valid certificates, it may encounter self-signed or expired certificates along the way).
> 
> The HTTP-01 challenge can only be done on port 80. Allowing clients to specify arbitrary ports would make the challenge less secure, and so it is not allowed by the ACME standard.
https://letsencrypt.org/docs/challenge-types/#http-01-challenge

So I believe that it is valid to follow their example.

### Expected behaviour
A challenge should be successfully validated when the http-01 request, after connecting to port 80, is redirected to https on port 443.

### Actual behaviour
The validation fails

joejulian:
### Description

The server will not yet have a valid certificate so we need to disable
certificate validation in the HTTPGetter.

Fixes #554

### DIFF
--- a/acme/api/handler.go
+++ b/acme/api/handler.go
@@ -64,8 +64,14 @@ type HandlerOptions struct {
 
 // NewHandler returns a new ACME API handler.
 func NewHandler(ops HandlerOptions) api.RouterHandler {
+	transport := &http.Transport{
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
+		},
+	}
 	client := http.Client{
-		Timeout: 30 * time.Second,
+		Timeout:   30 * time.Second,
+		Transport: transport,
 	}
 	dialer := &net.Dialer{
 		Timeout: 30 * time.Second,


maraino:
@dopey @joejulian See my comment in https://github.com/smallstep/certificates/issues/554#issuecomment-827795341

PR #570, opened 2021-05-10
Add arm6 to goreleaser

tashian:


dopey (approved):
lgtm

tashian:
Tested in GH actions.
Also I discovered that we were producing a .tar.gz of step-ca for WIndows, rather than a .zip. (Fixed.)

PR #599, opened 2021-06-02
certificate manager

dopey:


authority/admin/collection.go
```diff
@@ -0,0 +1,180 @@
...
+func (c *Collection) LoadByID(id string) (*linkedca.Admin, bool) {
+	return loadAdmin(c.byID, id)
+}
+
+func subProvNameHash(sub, provName string) string {
+	subHash := sha1.Sum([]byte(sub))
+	provNameHash := sha1.Sum([]byte(provName))
+	_res := sha1.Sum(append(subHash[:], provNameHash[:]...))
+	return string(_res[:])
+}
```
maraino, line 67: In go you can also use structs as keys, anything that is comparable. You should not use pointers, because then you will be comparing a memory address:

```go
type Foo struct {
    A string
    B string
}

Foo{"foo","bar"} == Foo{"foo","bar"} // true
Foo{"foo","bar"} == Foo{"bar","bar"} // false
&Foo{"foo","bar"} == &Foo{"foo","bar"} // false
```

authority/admin/collection.go
```diff
@@ -0,0 +1,180 @@
...
+		return fmt.Errorf("provisioner %s not found", adm.ProvisionerId)
+	}
+	// Store admin always in byID. ID must be unique.
+	if _, loaded := c.byID.LoadOrStore(adm.Id, adm); loaded {
+		return errors.New("cannot add multiple admins with the same id")
+	}
+
+	provName := p.GetName()
+	// Store admin always in bySubProv. Subject <-> ProvisionerName must be unique.
+	if _, loaded := c.bySubProv.LoadOrStore(subProvNameHash(adm.Subject, provName), adm); loaded {
```
maraino, line 101: See my previous comment.

authority/admin/collection.go
```diff
@@ -0,0 +1,180 @@
...
+		c.byID.Delete(adm.Id)
+		return errors.New("cannot add multiple admins with the same subject and provisioner")
+	}
+
+	if admins, ok := c.LoadByProvisioner(provName); ok {
+		c.byProv.Store(provName, append(admins, adm))
+		c.superCountByProvisioner[provName]++
+	} else {
+		c.byProv.Store(provName, []*linkedca.Admin{adm})
+		c.superCountByProvisioner[provName] = 1
```
maraino, line 111: Writing and reading from this map using different is not safe.

authority/admin/collection.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// SuperCount returns the total number of admins.
+func (c *Collection) SuperCount() int {
+	return c.superCount
+}
+
+// SuperCountByProvisioner returns the total number of admins.
+func (c *Collection) SuperCountByProvisioner(provName string) int {
+	if cnt, ok := c.superCountByProvisioner[provName]; ok {
```
maraino, line 140: See my previous comment.

authority/admin/collection.go
```diff
@@ -0,0 +1,180 @@
...
+func (c *Collection) LoadByProvisioner(provName string) ([]*linkedca.Admin, bool) {
+	a, ok := c.byProv.Load(provName)
+	if !ok {
+		return nil, false
+	}
+	admins, ok := a.([]*linkedca.Admin)
+	if !ok {
+		return nil, false
+	}
+	return admins, true
```
maraino, line 84: You can use `loadAdmin`

api/errors.go
```diff
@@ -8,8 +8,10 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/smallstep/certificates/acme"
+	"github.com/smallstep/certificates/authority/mgmt"
```
maraino, line 11: We should move `authority/mgmt` to `authority/admin`

authority/authority.go
```diff
@@ -11,29 +11,34 @@ import (
...
 
 	"github.com/smallstep/certificates/cas"
+	"github.com/smallstep/certificates/linkedca"
+	"github.com/smallstep/certificates/scep"
 
 	"github.com/pkg/errors"
+	"github.com/smallstep/certificates/authority/admin"
+	"github.com/smallstep/certificates/authority/config"
+	"github.com/smallstep/certificates/authority/mgmt"
+	authMgmtNosql "github.com/smallstep/certificates/authority/mgmt/db/nosql"
```
maraino, line 21: `admin/db/nosql`

authority/authority.go
```diff
@@ -143,6 +210,56 @@ func (a *Authority) init() error {
...
+			certProv, err := ProvisionerToCertificates(prov)
+			if err != nil {
+				return mgmt.WrapErrorISE(err, "error converting provisioner to certificates type")
+			}
+			a.config.AuthorityConfig.Provisioners = []provisioner.Interface{certProv}
+
+			// Create First Admin
+			adm := &linkedca.Admin{
+				ProvisionerId: prov.Id,
+				Subject:       "step",
```
maraino, line 243: Decide if this is the right name.

maraino (changes requested):
Still reviewing

authority/mgmt/api/middleware.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (h *Handler) authorizeToken(r *http.Request, token string) (*linkedca.Admin, error) {
+	jwt, err := jose.ParseSigned(token)
+	if err != nil {
+		return nil, mgmt.WrapError(mgmt.ErrorUnauthorizedType, err, "adminHandler.authorizeToken; error parsing x5c token")
+	}
+
+	verifiedChains, err := jwt.Headers[0].Certificates(x509.VerifyOptions{
+		Roots: h.rootPool,
+	})
```
maraino: This actually verifies that the certificate has `x509.ExtKeyUsageServerAuth`. That all of our certs have by default, but it looks to me that requiring `x509.ExtKeyUsageClientAuth` (using `x509.VerifyOptions.KeyUsages`) is more appropriate.

authority/mgmt/api/handler.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	return h
+}
+
+// Route traffic and implement the Router interface.
+func (h *Handler) Route(r api.Router) {
+	authnz := func(next nextHTTP) nextHTTP {
+		return h.extractAuthorizeTokenAdmin(h.requireAPIEnabled(next))
+	}
```
maraino: Most of the same functionality can be achieved using mTLS. We should consider if we want to change this.

authority/mgmt/api/middleware.go
```diff
@@ -0,0 +1,154 @@
...
+	if claims.Subject == "" {
+		return nil, mgmt.NewError(mgmt.ErrorUnauthorizedType,
+			"x5c.authorizeToken; x5c token subject cannot be empty")
+	}
+
+	var adm *linkedca.Admin
+	adminFound := false
+	adminSANs := append([]string{leaf.Subject.CommonName}, leaf.DNSNames...)
+	adminSANs = append(adminSANs, leaf.EmailAddresses...)
+	for _, san := range adminSANs {
```
maraino: No URIs no, IPs. Sounds ok to me, but just noting this.

maraino (changes requested):
Still reviewing

authority/admin/collection.go
dopey: what do you mean by using different? Do you want me to use a sync map?

authority/mgmt/api/admin.go
```diff
@@ -0,0 +1,169 @@
...
+	admins, nextCursor := h.auth.GetAdminCollection().Find(cursor, limit)
+	api.JSON(w, &GetAdminsResponse{
+		Admins:     admins,
+		NextCursor: nextCursor,
+	})
+}
+
+// CreateAdmin creates a new admin.
+func (h *Handler) CreateAdmin(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
```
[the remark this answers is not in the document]
dopey: Right now, you'll get a 404 because the routes don't get defined.

authority/mgmt/api/handler.go
dopey: Had a convo with @mmalone and he wanted to use tokens because it allowed for proxying. But he wanted to so more to authenticate the body of the request. Like hashing the request in the token or signing over the entire request body using the private key.

ca/ca.go
```diff
@@ -167,16 +179,47 @@ func (ca *CA) Init(config *authority.Config) (*CA, error) {
...
-		if err := chi.Walk(mux, walkFunc); err != nil {
-			fmt.Printf("Logging err: %s\n", err.Error())
+	// Admin API Router
+	adminDB := auth.GetAdminDatabase()
+	if adminDB != nil {
+		adminHandler := adminAPI.NewHandler(auth)
+		mux.Route("/admin", func(r chi.Router) {
+			adminHandler.Route(r)
+		})
+	}
```
maraino, line 189: Only if admin is enabled?

linkedca/provisioners.go
```diff
@@ -0,0 +1,38 @@
...
+		v = new(ProvisionerDetails_Azure)
+	case Provisioner_ACME:
+		v = new(ProvisionerDetails_ACME)
+	case Provisioner_X5C:
+		v = new(ProvisionerDetails_X5C)
+	case Provisioner_K8SSA:
+		v = new(ProvisionerDetails_K8SSA)
+	case Provisioner_SSHPOP:
+		v = new(ProvisionerDetails_SSHPOP)
+	default:
```
maraino, line 30: SCEP is missing.

maraino: And I'm not sure if we want non-generated code here.

maraino (changes requested):

authority/admin/collection.go
dopey: `loadAdmin` returns a single Admin, not a list of admins.

authority/config/config.go
```diff
@@ -0,0 +1,306 @@
...
+	// DefaultTLSOptions represents the default TLS version as well as the cipher
+	// suites used in the TLS certificates.
+	DefaultTLSOptions = TLSOptions{
+		CipherSuites: CipherSuites{
+			"TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305",
+			"TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
+			"TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
+		},
+		MinVersion:    1.2,
+		MaxVersion:    1.2,
```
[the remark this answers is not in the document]
dopey: Let's do this in a different change. There's some logic in how min/max are set based off the defaults that I don't really want to change as part of this PR.

authority/config/config.go
```diff
@@ -0,0 +1,306 @@
...
+	// Check that only one K8sSA is enabled
+	var k8sCount int
+	for _, p := range c.Provisioners {
+		if p.GetType() == provisioner.TypeK8sSA {
+			k8sCount++
+		}
+	}
+	if k8sCount > 1 {
+		return errors.New("cannot have more than one kubernetes service account provisioner")
+	}
```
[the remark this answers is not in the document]
dopey: We do. Because the IDForToken is a static const for k8ssa provisioners. So we won't be able to add more than one.

authority/provisioner/claims.go
```diff
@@ -7,6 +7,21 @@ import (
...
+	UserDuration *Durations `json:"userDurations"`
+	HostDuration *Durations `json:"hostDuration"`
+}
+
+type Durations struct {
+	Min     string `json:"min"`
+	Max     string `json:"max"`
+	Default string `json:"default"`
+}
+
```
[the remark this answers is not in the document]
dopey: nope

authority/provisioner/collection.go
```diff
@@ -131,22 +148,7 @@ func (c *Collection) LoadByCertificate(cert *x509.Certificate) (Interface, bool)
...
-			case TypeACME:
-				return c.Load("acme/" + string(provisioner.Name))
-			case TypeX5C:
-				return c.Load("x5c/" + string(provisioner.Name))
-			case TypeK8sSA:
-				return c.Load(K8sSAID)
-			default:
-				return c.Load(string(provisioner.CredentialID))
-			}
+			return c.LoadByName(string(provisioner.Name))
```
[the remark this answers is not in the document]
dopey: Good question. Let's discuss at design meeting. Will add to list.

ca/adminClient.go
```diff
@@ -0,0 +1,572 @@
...
+	x5cJWK      *jose.JSONWebKey
+	x5cCertFile string
+	x5cCertStrs []string
+	x5cCert     *x509.Certificate
+	x5cIssuer   string
+	x5cSubject  string
+}
+
+// NewAdminClient creates a new AdminClient with the given endpoint and options.
+func NewAdminClient(endpoint string, opts ...ClientOption) (*AdminClient, error) {
```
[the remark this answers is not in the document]
dopey: Let's hold off on this. We're using x5c-cert and x5c-key which the cli already uses for different things. So we could screw the client up for a `step ca certificate` request.

linkedca/provisioners.go
dopey: made sense to me to have it in linkedca, since it's only acting on linkedca types.

maraino (changes requested):
Mostly ok, I've added a few new comments.

authority/admin/api/provisioner.go
```diff
@@ -0,0 +1,175 @@
...
+
+	var (
+		p   provisioner.Interface
+		err error
+	)
+	if len(id) > 0 {
+		if p, err = h.auth.LoadProvisionerByID(id); err != nil {
+			api.WriteError(w, admin.WrapErrorISE(err, "error loading provisioner %s", id))
+			return
+		}
```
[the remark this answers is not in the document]
dopey: I'm not sure I understand the question. The api accepts either a name from the url or an id query parameter. So it attempts to load the provisioner either by name or id, depending on what the input was.

authority/admin/api/provisioner.go
```diff
@@ -0,0 +1,175 @@
...
```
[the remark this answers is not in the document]
dopey: Same answer as before. It's not a name AND an ID. It's either one or the other. If a query parameter is submitted, the name in the URL is completely ignored.

authority/admin/db/nosql/admin.go
```diff
@@ -0,0 +1,178 @@
...
+
+// dbAdmin is the database representation of the Admin type.
+type dbAdmin struct {
+	ID            string              `json:"id"`
+	AuthorityID   string              `json:"authorityID"`
+	ProvisionerID string              `json:"provisionerID"`
+	Subject       string              `json:"subject"`
+	Type          linkedca.Admin_Type `json:"type"`
+	CreatedAt     time.Time           `json:"createdAt"`
+	DeletedAt     time.Time           `json:"deletedAt"`
```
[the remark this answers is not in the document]
dopey: I think I always check if DeletedAt is zero

authority/administrator/collection.go
```diff
@@ -0,0 +1,243 @@
...
+
+// Update updates the given admin in all related lists and collections.
+func (c *Collection) Update(id string, nu *linkedca.Admin) (*linkedca.Admin, error) {
+	adm, ok := c.LoadByID(id)
+	if !ok {
+		return nil, admin.NewError(admin.ErrorNotFoundType, "admin %s not found", adm.Id)
+	}
+	if adm.Type == nu.Type {
+		return nil, admin.NewError(admin.ErrorBadRequestType, "admin %s already has type %s", id, adm.Type)
+	}
```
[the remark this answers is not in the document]
dopey: yep.

authority/authority.go
```diff
@@ -386,16 +438,44 @@ func (a *Authority) init() error {
...
-		if err := a.provisioners.Store(p); err != nil {
-			return err
+
+		provs, err := a.adminDB.GetProvisioners(context.Background())
+		if err != nil {
+			return admin.WrapErrorISE(err, "error loading provisioners to initialize authority")
+		}
+		if len(provs) == 0 {
+			// Create First Provisioner
+			prov, err := CreateFirstProvisioner(context.Background(), a.adminDB, a.config.Password)
```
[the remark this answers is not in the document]
dopey: We're actually going to change completely how this is done (based on recent design discussion). The decision is to do this during `step ca init`. I think we're going to leave it as is for now and just not encourage anyone to use it until we've pushed out the new onboarding flows in `step ca init`.

authority/provisioners.go
```diff
@@ -19,25 +29,573 @@ func (a *Authority) GetEncryptedKey(kid string) (string, error) {
...
+	if err := a.adminDB.DeleteProvisioner(ctx, provID); err != nil {
+		if err := a.reloadAdminResources(ctx); err != nil {
+			return admin.WrapErrorISE(err, "error reloading admin resources on failed provisioner remove")
+		}
+		return admin.WrapErrorISE(err, "error deleting provisioner %s", provName)
+	}
+	return nil
+}
+
+func CreateFirstProvisioner(ctx context.Context, db admin.DB, password string) (*linkedca.Provisioner, error) {
```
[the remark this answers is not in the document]
dopey: Same comment as above. We're not going to encourage users to turn on `enableAdmin` in the configuration until we button up onboarding. It will just be a hidden flag for now.

authority/authority.go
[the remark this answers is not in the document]
maraino: sounds good

authority/admin/collection.go
maraino: Ups I meant using different goroutines. I think we are now locking at a different level so it should be ok.

authority/admin/api/provisioner.go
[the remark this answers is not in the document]
maraino: I mean that the URL will be `/admin/provisioners/the-name?id=uuid` with the current code If `the-name` and `uuid` do not belong to the same provisioner the name is totally ignored.

authority/admin/api/provisioner.go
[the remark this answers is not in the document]
maraino: Is there any benefit in the API in allowing 2 unique identifiers (name and UUID)?

maraino (approved):
LGTM

ca/adminClient.go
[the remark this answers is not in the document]
maraino: Ok for now, but we should move those to the client, adding something like `ca.WithAdminCredentials(cert, key, ...)` should not screw up anything. For example, the sign with an x5c provisioner has the token in the body, not as a header.

authority/admin/api/admin.go
```diff
@@ -0,0 +1,139 @@
...
+// CreateAdminRequest represents the body for a CreateAdmin request.
+type CreateAdminRequest struct {
+	Subject     string              `json:"subject"`
+	Provisioner string              `json:"provisioner"`
+	Type        linkedca.Admin_Type `json:"type"`
+}
+
+// Validate validates a new-admin request body.
+func (car *CreateAdminRequest) Validate() error {
+	return nil
```
maraino, line 21: We should check the properties are not empty and the type is a valid one.

authority/admin/api/admin.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// UpdateAdminRequest represents the body for a UpdateAdmin request.
+type UpdateAdminRequest struct {
+	Type linkedca.Admin_Type `json:"type"`
+}
+
+// Validate validates a new-admin request body.
+func (uar *UpdateAdminRequest) Validate() error {
+	return nil
```
maraino, line 37: We should check the type and return bad request if is not valid.

authority/admin/api/middleware.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/smallstep/certificates/authority/admin"
+)
+
+type nextHTTP = func(http.ResponseWriter, *http.Request)
+
+// requireAPIEnabled is a middleware that ensures the Administration API
+// is enabled before servicing requests.
+func (h *Handler) requireAPIEnabled(next nextHTTP) nextHTTP {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if h.db == nil {
```
maraino, line 17: This is not exactly true, EnableAdmin is also required.

authority/admin/api/admin.go
```diff
@@ -0,0 +1,139 @@
...
+func (h *Handler) GetAdmin(w http.ResponseWriter, r *http.Request) {
+	id := chi.URLParam(r, "id")
+
+	adm, ok := h.auth.LoadAdminByID(id)
+	if !ok {
+		api.WriteError(w, admin.NewError(admin.ErrorNotFoundType,
+			"admin %s not found", id))
+		return
+	}
+	api.JSON(w, adm)
```
maraino, line 55: Protojson will probably create a nicer json, specially in enumerations like type.

authority/admin/api/admin.go
```diff
@@ -0,0 +1,139 @@
...
+			"error parsing cursor and limit from query params"))
+		return
+	}
+
+	admins, nextCursor, err := h.auth.GetAdmins(cursor, limit)
+	if err != nil {
+		api.WriteError(w, admin.WrapErrorISE(err, "error retrieving paginated admins"))
+		return
+	}
+	api.JSON(w, &GetAdminsResponse{
```
maraino, line 72: Protojson?

authority/admin/api/admin.go
```diff
@@ -0,0 +1,139 @@
...
+		Subject:       body.Subject,
+		Type:          body.Type,
+	}
+	// Store to authority collection.
+	if err := h.auth.StoreAdmin(r.Context(), adm, p); err != nil {
+		api.WriteError(w, admin.WrapErrorISE(err, "error storing admin"))
+		return
+	}
+
+	api.JSON(w, adm)
```
maraino, line 107: Protojson?

authority/admin/api/admin.go
```diff
@@ -0,0 +1,139 @@
...
+
+	id := chi.URLParam(r, "id")
+
+	adm, err := h.auth.UpdateAdmin(r.Context(), id, &linkedca.Admin{Type: body.Type})
+	if err != nil {
+		api.WriteError(w, admin.WrapErrorISE(err, "error updating admin %s", id))
+		return
+	}
+
+	api.JSON(w, adm)
```
maraino, line 138: protozoon

authority/admin/api/provisioner.go
```diff
@@ -0,0 +1,175 @@
...
+
+	p, next, err := h.auth.GetProvisioners(cursor, limit)
+	if err != nil {
+		api.WriteError(w, errs.InternalServerErr(err))
+		return
+	}
+	api.JSON(w, &GetProvisionersResponse{
+		Provisioners: p,
+		NextCursor:   next,
+	})
```
maraino, line 69: Provisioners won't be nicely marshaled without using protojson.

authority/admin/api/provisioner.go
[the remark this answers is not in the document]
maraino: Having to unique identifiers at this level is a little upsetting. At the db layer makes total sense.

authority/admin/db/nosql/nosql.go
```diff
@@ -0,0 +1,88 @@
...
+			return errors.Wrapf(err, "error marshaling authority type: %s, value: %v", typ, nu)
+		}
+	}
+	var oldB []byte
+	if old == nil {
+		oldB = nil
+	} else {
+		oldB, err = json.Marshal(old)
+		if err != nil {
+			return errors.Wrapf(err, "error marshaling acme type: %s, value: %v", typ, old)
```
maraino, line 57: acme type?

authority/admin/errors.go
```diff
@@ -0,0 +1,223 @@
...
+	errorServerInternalMetadata = errorMetadata{
+		typ:     ErrorServerInternalType.String(),
+		details: "the server experienced an internal error",
+		status:  500,
+	}
+	errorMap = map[ProblemType]errorMetadata{
+		ErrorNotFoundType: {
+			typ:     ErrorNotFoundType.String(),
+			details: "resource not found",
+			status:  400,
```
maraino, line 76: 404

authority/admin/errors.go
```diff
@@ -0,0 +1,223 @@
...
+		},
+		ErrorAuthorityMismatchType: {
+			typ:     ErrorAuthorityMismatchType.String(),
+			details: "resource not owned by authority",
+			status:  401,
+		},
+		ErrorDeletedType: {
+			typ:     ErrorDeletedType.String(),
+			details: "resource is deleted",
+			status:  http.StatusUnauthorized,
```
maraino, line 86: 404?

maraino (commented):
Pretty much ok, but added a few comments.

authority/admin/api/admin.go
dopey: Agreed. I'm not checking the provisioner create inputs either. I'm not going to have time to test those validators, so didn't want to get too deep.

dopey: I can do some simple property checking though.

authority/admin/api/admin.go
dopey: Can't use it. Throws an error.

authority/admin/api/provisioner.go
dopey: The provisioners in this list aren't protojson linkedca provisioners anyway. They're the certificates type of provisioner. This method isn't being used right now. We're still using the existing API with `step ca provisioner list`.

I can take this method out for now if you want.

dopey:
Update pushed. I addressed those comments which I could easily fix. Take a look when you have a chance and let me know.

authority/admin/api/admin.go
maraino: Makes sense, a solution to this, to get a nicer JSON is to use a proto message like the [AdminList](
https://github.com/smallstep/linkedca/blob/27424aae8d2588aad217c75897cba5eebcb9aa91/admin.proto#L24-L26). With a new cursor property.

authority/admin/api/provisioner.go
maraino: Ok for now, but to get a nicer JSON the same that we can use [ProvisionerList](https://github.com/smallstep/linkedca/blob/27424aae8d2588aad217c75897cba5eebcb9aa91/provisioners.proto#L50-L52) with a cursor. 

It's also possible to add custom MarshalJSON and UnmarshalJSON to GetProvisionersResponse that uses protojson to marshal the list of Provisioners/Admins.

maraino (approved):
LGTM.

I've added some comments on mixed json/protojson types, but taking into account this is a beta release I'm ok with fixing them later. It's not ideal, because of backward compatibility with the cli, but ok for now.

Issue #123, opened 2019-10-07
Option(s) to denylist certain common names / subjects / alternative names


kbabioch:
### What would you like to be added

I might have missed something very obvious, but there seems to be no way to limit and constrain the issuance of certificate in general and/or for specific provisioners.

### Why this is needed

There is probably a lot of possible options imaginable here, but the particular use-case I have in mind is to make sure that certificates for some domains cannot be issued using the ACME provisioner, even though someone might have control over the domain and/or the webserver on this domain.

mmalone:
You didn't miss anything. This is a problem with ACME. The challenge here is that the sorts of policies people want to capture vary quite a bit. We have some tech to address this problem generically but it's not released yet :/.

In the meantime, I've been thinking about basic filtering based on domain names... perhaps borrowing from the [name constraint](https://tools.ietf.org/html/rfc5280#section-4.2.1.10) mechanism from Web PKI. It'd probably be provisioner-level and you'd be able to add a parameter like `"domains": ["foo.example.com", ".dev.example.com"]` which, in this case, would limit the provisioner to issuing certificates for "foo.example.com" and any subdomain under ".dev.example.com". Would that be enough for you?

darix:
name constraints is something on top of it 

lets say I have an intermediate CA file with a nameConstraint=dns:example.com

but I also want to prevent people from getting certs for some shared production servers under example.com. only a subset of people should be allowed to get certs for those.

mmalone:
Sorry, yea, what I meant was we could borrow the syntax and semantics of name constraints and re-use them to restrict certificate issuance on a per-provisioner basis. The question of whether CA root/intermediate certificates contain name constraints is separate.

What I'm proposing is that you could have a provisioner configuration like this:

```
{
    "type": "ACME",
    "name": "acme-issuer",
    "domains": {
        "allowed": ["foo.example.com", ".dev.example.com"],
        "denied": ["ca.dev.example.com"]
    }
}
```

The "foo.example.com" and ".dev.example.com" syntax is borrowed from name constraints. The former is an exact match, and the latter is a wildcard. So, in this case, we're saying that the ACME provisioner should only issue certificates for "foo.example.com" and for any subdomain under "dev.example.com", but _not_ "ca.dev.example.com". We wouldn't put name constraints in the intermediate certificate or anywhere else. We're just re-using the name constraint syntax to define which domains a provisioner can/can't issue certificates for.

[...]

PR #602, opened 2021-06-03
IP Identifier Validation [RFC8738]

hslatman:
Adds support for RFC8738: IP Identifier Validation


acme/api/order_test.go
```diff
@@ -60,6 +61,68 @@ func TestNewOrderRequest_Validate(t *testing.T) {
...
+						{Type: "ip", Value: "192.168.42.42"},
+					},
+					NotAfter:  naf,
+					NotBefore: nbf,
+				},
+				nbf: nbf,
+				naf: naf,
+			}
+		},
+		"ok/mixed-ipv4-and-ipv6": func(t *testing.T) test {
```
dopey, line 110: Can we add one more test for the case when the IP is parsed but invalid? Otherwise tests look great.

acme/order.go
```diff
@@ -213,6 +187,130 @@ func (o *Order) Finalize(ctx context.Context, db DB, csr *x509.CertificateReques
...
+
+	// TODO: limit what IP addresses can be used? Only private? Only certain ranges (i.e. only allow the specific ranges by default, configuration for all?)
+	// TODO: can DNS already be limited to a certain domain? That would probably be nice to have too, but maybe not as part of this PR
+	// TODO: if it seems not too big of a change, make consts/enums out of the stringly typed identifiers (challenge types, identifier types)
+	// TODO: only allow IP based identifier based on configuration? Some additional configuration and validation on the provisioner for this case.
+
+	// Validate identifier names against CSR alternative names.
+	//
+	// Note that with certificate templates we are not going to check for the
+	// absence of other SANs as they will only be set if the templates allows
```
dopey, line 225: if the template allows them

acme/order.go
```diff
@@ -213,6 +187,130 @@ func (o *Order) Finalize(ctx context.Context, db DB, csr *x509.CertificateReques
...
+// adresses are NOT considered equal to the IPv4 address in this case.
+// Both IPs should be the same version AND equal to each other.
+func ipsAreEqual(x, y net.IP) bool {
+	if matchAddrFamily(x, y) {
+		return x.Equal(y)
+	}
+	return false
+}
+
+// matchAddrFamily returns if two IPs are both IPv4 OR IPv6
```
dopey, line 308: returns `true` if

dopey (changes requested):
I left a few comments. In general, this looks great. We need some more unit tests, but otherwise looking good.

acme/api/order_test.go
hslatman: Added that test with an invalid IP

hslatman:
@dopey: Let me know if the tests look good like this or would like me to move more Finalize() tests over to (just) the sans function test. I think they both have their purpose, but certain test cases are effectively duplicates right now.

acme/order.go
```diff
@@ -296,16 +300,16 @@ func canonicalize(csr *x509.CertificateRequest) (canonicalized *x509.Certificate
 }
 
 // ipsAreEqual compares IPs to be equal. IPv6 representations of IPv4
-// adresses are NOT considered equal to the IPv4 address in this case.
-// Both IPs should be the same version AND equal to each other.
+// adresses are considered equal to the IPv4 address in this case.
+// TODO: is this behavior OK to keep?
```
dopey, line 304: What about this behavior exactly are we questioning? Current behavior (as written) is that they will NOT be considered equal. Is that correct? I could be misunderstanding the code. My interpretation conflicts with the current comment above the function.

dopey:
Test coverage and changes look great. It would be awesome to move more tests out of Finalize and over to SANs and try to get rid of duplicates, if you have the bandwith. 

The last remaining question is the one you asked about ipv6(ipv4) == ipv4. My gut says that the ips should have to be exactly equal, but I may be missing some context here. e.g. Maybe this would make my life difficult if my system required the use of ipv6. But in that case I would expect the order to have the ipv6 value in it.

mholt:
I believe it is typical to explicitly have both the IPv4 and IPv6 subjects.

dopey:
@mholt In which case it sounds like ipv6 should be compared to ipv6 and ipv4 to ipv4. So if the CSR has two IPs (one ipv6 and ipv4) then the order should have the same as well. Am I understanding that correctly?

mholt:
@dopey That's my understanding, the IP has to be exactly equal. (RFC 8738 doesn't really expound on this, but it does seem to imply that IPv4 and IPv6 are treated as separate subjects.)

hslatman:
I agree that IPv4 vs. IPv6 should indeed be a full equality check. When I wrote the tests I actually found out that Go considers them the same if it's IPv4-in-IPv6 (and generally, that's OK, I think) and my initial implementation was in fact wrong 😅, resulting in the IPv4-in-IPv6 case to be equal to the IPv4 address. In this use case they should be fully the same to make it work correct in all cases, so I'll implement it like that.

I'll move the Finalize test cases that effectively test the sans() function (and don't need to be mocked) to the tests for sans. Shouldn't take too long.

dopey:
@hslatman gotcha. That confused me too. I saw the code didn't change but the comment did. 
Everything you said sounds great!

hslatman:
Unfortunately, it's not trivial to do 😅 

Due to how net.ParseIP() works, an IPv4-in-IPv6 address will actually be stored just like the IPv4. Every IPv4 address is essentially 16 bytes, just like the IPv6 address, but prefixed with 12 x \x00 . This is also tracked here: https://github.com/golang/go/issues/37921.

mholt:
FWIW, my use case is probably OK with this if it remains a "known issue" -- not a big deal; but I can only speak for me.

dopey:
I support the "known issue" route as well.

hslatman:
Refactored the tests and made the issue with IPv4 mapped IPv6 addresses clear in a comment.

dopey (approved):
lgtm

hslatman:
Hey @tashian: currently there are no configuration options available. An ACME client can get a cert for IPv4 or IPv6 identifier(s) (either for just one or more IPs, or combined with DNS names, in a single CSR) from an ACME provisioner.

Some thoughts about configurability of this part: if I remember correctly the matter of authorising certain IP ranges was brought up in a team discussion. It was considered similar to what could be implemented for allowing specific (sub)domains (also see #123). The "Inventory" functionality was mentioned as one of the ways to do that. Another solution is to implement support for checking Name Constraints (also mentioned in that issue) in the CA certificate (and enforcing them when issuing a certificate). I think it's also possible to use the certificate flexibility functionality to filter out CSRs with (certain) IP ranges, if that's the goal.

PR #612, opened 2021-06-15
Allow reading pin from kms string

gdbelvin:
Originally `pin-source` would be read twice, once by `u.Pin()` and again inside `kms.New()`. 
This double read is problematic when reading pins from FIFO pipes. 

This PR fixes this by removing the double read.

### DIFF
--- a/cmd/step-pkcs11-init/main.go
+++ b/cmd/step-pkcs11-init/main.go
@@ -119,19 +119,6 @@ func main() {
 		fatal(err)
 	}
 
-	kmsPin := u.Pin()
-	if c.Pin == "" && kmsPin != "" {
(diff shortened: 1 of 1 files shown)


maraino:
@gdbelvin Isn't the result equivalent to what it is right now, meaning if the pin is in the uri, it will be used by pkcs11
https://github.com/smallstep/certificates/blob/c4d0c8a18e51cb3f4f8d16863ca21115789aefc7/kms/pkcs11/pkcs11.go#L54-L60

maraino (approved):
Looks good, but I think the result is the same as before, can you confirm?

maraino (approved):
👍 nice catch for your use case

PR #613, opened 2021-06-17
Extractable private keys and certs

gdbelvin:
By default, pkcs11 objects are not extractable from the HSM, even when the permission to do so are available.
This might be fine for some scenarios, but best practices specify that objects on HSMs should be exportable
to support availability and disaster recovery. 

This PR adds support for setting the extractable bit in PKCS11 during key generation and certificate storage.


maraino (approved):
Thanks @gdbelvin

gdbelvin:
LGTM

PR #631, opened 2021-07-09
Log serial number as hex byte string

### LINKED ISSUE #630: Certificate serial numbers should not be logged as integer literals
When the CA writes requests to its JSON-formatted log, serial numbers are output as decimal-formatted integer literals (because apparently that's what Go's `math/big.Int.MarshalJSON` does. While this is strictly valid JSON, many JSON parsers behave just like 1990s Javascript implementations, i.e. they have no concept of integer values and treat everything as a 64bit IEEE-754 float, limiting usable integer values to 0…2⁵³-1. This will cause trouble when trying to analyze logs with automated tools. Formatting serial numbers as a string makes the problem go away.

I suggest changing the output format to match the behavior of `openssl x509 -text` default: byte hex values separated by colons: `de:ad:c0:d3:11:22:33:44:55:66`

hillu:
This change should cause less confusion in JSON parser implementations that think that every number is a 64 bit float.

Close #630

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -400,7 +400,7 @@ func logOtt(w http.ResponseWriter, token string) {
 func LogCertificate(w http.ResponseWriter, cert *x509.Certificate) {
 	if rl, ok := w.(logging.ResponseLogger); ok {
 		m := map[string]interface{}{
-			"serial":      cert.SerialNumber,
+			"serial":      fmtBytes(cert.SerialNumber.Bytes()),
 			"subject":     cert.Subject.CommonName,
 			"issuer":      cert.Issuer.CommonName,
 			"valid-from":  cert.NotBefore.Format(time.RFC3339),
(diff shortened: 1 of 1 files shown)


hillu:
The labeler/label error message does not tell me anything useful. What went wrong?

dopey:
@hillu Don't worry about that. It's just a job that's supposed to label the PR as `needs triage` for us. But Github made a change recently so that users outside the org no longer have the permission to run that job. Now we've gotta fix that.

maraino (changes requested):
I think is better to convert the serial number to a string, so the logs show the same representation that we use in the DB.

@dopey What do you think?

hillu:
@maraino Logging the serial as a stringified decimal number fixes my problem, too. I chose the hexadecimal number because that's what OpenSSL uses in `x509 -text` output.

maraino:
`step certificate inspect` shows both, base 10 and base 16 (hex). I think commands like `step ca revoke` uses always the base 10 version so I think it makes more sense to use it here too.

A different question is if base 10 was the right decision for this.

dopey:
@maraino I was thinking the same thing. So, for the purpose of this PR I think we should change to decimal representation. That's what's used in the database and that's what's expected by the `step ca revoke --serial <#>` command. 

@hillu if you think a hex representation would be better more conventional, let's open up another issue to discuss that. A change like that would need some thought since it would break backward compatibility.

hillu:
I have opened a new PR #641 with a simple one-line patch and am closing this.

PR #641, opened 2021-07-14
Log certificate's serial number as stringified decimal number

hillu:
Using a JSON string fixes a common issue with JSON parsers that
deserialize all numbers to a 64-bit IEEE-754 floats. (Certificate
serial numbers are usually 128 bit values.)

This change is consistent with existing log entries for revocation
requests.

See also: #630, #631

### Description

💔Thank you!


maraino (approved):
lgtm

PR #644, opened 2021-07-15
Fix logging provisioner name as string

hslatman:
### Description
The provisioner name was printed as a byte slice due to a regression. This PR prints it as a string again when logging a certificate.

💔Thank you!

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -418,7 +418,7 @@ func LogCertificate(w http.ResponseWriter, cert *x509.Certificate) {
 				if len(val.CredentialID) > 0 {
 					m["provisioner"] = fmt.Sprintf("%s (%s)", val.Name, val.CredentialID)
 				} else {
-					m["provisioner"] = val.Name
+					m["provisioner"] = string(val.Name)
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -418,7 +418,7 @@ func LogCertificate(w http.ResponseWriter, cert *x509.Certificate) {
 				if len(val.CredentialID) > 0 {
 					m["provisioner"] = fmt.Sprintf("%s (%s)", val.Name, val.CredentialID)
 				} else {
-					m["provisioner"] = val.Name
+					m["provisioner"] = string(val.Name)
```
dopey, line 421: @maraino is this a result of a recent change to how we store the oid extension?

maraino: No, it looks like it has been like this forever. 

For most of the provisioners is not a problem, but provisioners like X5C that do not have a CredentialID shows `provisioner="[88 53 67]` instead of `provisioner="X5C"`

maraino (approved):
lgtm

PR #650, opened 2021-07-17
ACME External Account Binding

hslatman:
### Description

This PR adds External Account Binding support to the ACME provisioner.

The implementation currently supports key material creation by using a new admin (beta) endpoint at `/admin/acme/eab`. Changes to the CLI can be found [here](https://github.com/smallstep/cli/pull/512/files).

Still need to iron out some details and have to add tests, but the basics work with `acmez` already 🙂


acme/api/account.go
```diff
@@ -205,3 +224,83 @@ func (h *Handler) GetOrdersByAccountID(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	acmeProv, ok := prov.(*provisioner.ACME) // TODO: rewrite into providing configuration via function on acme.Provisioner
+	if !ok || acmeProv == nil {
+		return nil, acme.NewErrorISE("provisioner in context is not an ACME provisioner")
+	}
+
+	shouldSkipAccountBindingValidation := !acmeProv.RequireEAB
```
dopey, line 241: Do we need a variable for this or can we just check `if !acmeProv.RequireEAB`?

acme/api/handler.go
```diff
@@ -145,12 +153,27 @@ func (d *Directory) ToLog() (interface{}, error) {
 // for client configuration.
 func (h *Handler) GetDirectory(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
+	prov, err := provisionerFromContext(ctx)
+	if err != nil {
+		api.WriteError(w, err)
+		return
+	}
+
+	acmeProv, ok := prov.(*provisioner.ACME) // TODO: rewrite into providing configuration via function on acme.Provisioner
```
dopey, line 162: We may want to just add this section to `provisionerFromContext`, right? Is there anywhere where we load a provisioner using this method where we don't want to make sure it's an ACME provisioner? We only use this method in the ACME code. But if we want to be extra safe, we could just make it a new method like `acmeProvisionerFromContext`.

dopey:
https://github.com/smallstep/certificates/blob/0dd6564b1e16627b160e491f9e4b4395221e2d90/ca/ca.go#L154-L171 -- In this section we create the acmeDB, and then directly below we instantiate the admin API handler. We can just pass the acmeDB to the admin Handler as another argument. 

Or we can just add another attribute to authority (here https://github.com/smallstep/certificates/blob/0dd6564b1e16627b160e491f9e4b4395221e2d90/authority/authority.go#L35-L73) like `acmeDB` and just store the acmeDB there when you create it (using a setter method since you won't have access to the private variable from ca/ca.go). The admin handler already takes a *authority as its only argument. Then you'd just add a method like `GetACMEDatabase` (here https://github.com/smallstep/certificates/blob/0dd6564b1e16627b160e491f9e4b4395221e2d90/authority/authority.go#L500-L509).

acme/api/account.go
hslatman: More of a personal preference, so that it's clear what is happening. Removed it in this case; it's not really hard to understand what's happening without the var.

acme/api/handler.go
hslatman: The `provisionerFromContext` is indeed only used in the ACME part (there's another, similar function for SCEP, though), but I think the idea was to provide just the interface for reasons related to overall architecture. the `RequireEAB` is the first property that is used directly and didn't really fit in with the way other properties for the (ACME) provisioner were handled, so I chose to implement `acmeProvisionerFromContext`. There're probably other ways, like having a function on the `acme.Provisioner` interface; perhaps I'll think of something nicer.

hslatman:
> I'm fine with either method. The second may look a bit cleaner.

I've chosen the first method, so that the `acmeDB` dependency is injected directly.

acme/api/account.go
```diff
@@ -225,49 +237,47 @@ func (h *Handler) GetOrdersByAccountID(w http.ResponseWriter, r *http.Request) {
...
 		return nil, acme.NewError(acme.ErrorExternalAccountRequiredType, "no external account binding provided")
 	}
 
+	// TODO: extract the EAB in a similar manner as JWS, JWK, payload, etc? That would probably move a lot/all of
+	// the logic of this function into the middleware. Should not be too hard, because the middleware does know
+	// about the handler and thus about its dependencies.
 	eabJSONBytes, err := json.Marshal(nar.ExternalAccountBinding)
 	if err != nil {
-		return nil, acme.WrapErrorISE(err, "error marshaling externalAccountBinding")
+		return nil, acme.WrapErrorISE(err, "error marshaling externalAccountBinding into JSON")
```
dopey, line 260: Isn't this marshaling to bytes?

acme/api/account.go
```diff
@@ -225,49 +237,47 @@ func (h *Handler) GetOrdersByAccountID(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	// TODO: verify supported algorithms against the incoming alg (and corresponding settings)?
 	// TODO: implement strategy pattern to allow for different ways of verification (i.e. webhook call) based on configuration
 
 	keyID := eabJWS.Signatures[0].Protected.KeyID
-	externalAccountKey, err := h.db.GetExternalAccountKey(ctx, keyID)
+	externalAccountKey, err := h.db.GetExternalAccountKey(ctx, acmeProv.Name, keyID)
 	if err != nil {
+		if _, ok := err.(*acme.Error); ok {
```
dopey, line 273: Why do we need this extra error typing block? Maybe that will become obvious once I've seen db.GetExternalAccountKey.

acme/db/nosql/account.go
```diff
@@ -177,55 +178,106 @@ func (db *DB) CreateExternalAccountKey(ctx context.Context, name string) (*acme.
...
 
 // GetExternalAccountKey retrieves an External Account Binding key by KeyID
-func (db *DB) GetExternalAccountKey(ctx context.Context, keyID string) (*acme.ExternalAccountKey, error) {
+func (db *DB) GetExternalAccountKey(ctx context.Context, provisionerName string, keyID string) (*acme.ExternalAccountKey, error) {
 	dbeak, err := db.getDBExternalAccountKey(ctx, keyID)
 	if err != nil {
 		return nil, err
 	}
 
+	if dbeak.ProvisionerName != provisionerName {
```
dopey, line 209: Why is it important that the provisioner name in the request match the original provisionerName used to bind the key?

acme/db/nosql/account.go
```diff
@@ -177,55 +178,106 @@ func (db *DB) CreateExternalAccountKey(ctx context.Context, name string) (*acme.
...
+		AccountID:       dbeak.AccountID,
+		KeyBytes:        dbeak.KeyBytes,
+		CreatedAt:       dbeak.CreatedAt,
+		BoundAt:         dbeak.BoundAt,
 	}, nil
 }
 
-func (db *DB) UpdateExternalAccountKey(ctx context.Context, eak *acme.ExternalAccountKey) error {
+func (db *DB) DeleteExternalAccountKey(ctx context.Context, keyID string) error {
+	_, err := db.db.Get(externalAccountKeyTable, []byte(keyID))
```
dopey, line 225: I think we could do these together in an `db.db.Update` call. But I'm not sure if there's any actual value to that. The idea for that method was that it would be sort of atomic, but I doubt it actually is. Depends on the details of the database.

acme/db/nosql/account.go
```diff
@@ -177,55 +178,106 @@ func (db *DB) CreateExternalAccountKey(ctx context.Context, name string) (*acme.
...
+		return nil, err
+	}
+
+	keys := make([]*acme.ExternalAccountKey, len(entries))
+	for i, entry := range entries {
+		dbeak := new(dbExternalAccountKey)
+		if err = json.Unmarshal(entry.Value, dbeak); err != nil {
+			return nil, errors.Wrapf(err, "error unmarshaling external account key %s into dbExternalAccountKey", string(entry.Key))
+		}
+		keys[i] = &acme.ExternalAccountKey{
```
dopey, line 249: Shouldn't we be filtering against the input provisionerName?

authority/admin/api/acme.go
```diff
@@ -0,0 +1,113 @@
...
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// CreateExternalAccountKeyRequest is the type for POST /admin/acme/eab requests
+type CreateExternalAccountKeyRequest struct {
+	ProvisionerName string `json:"provisioner"`
+	Name            string `json:"name"`
+}
+
+// Validate validates a new-admin request body.
```
dopey, line 19: new external account key body?

authority/admin/api/acme.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/go-chi/chi"
+	"github.com/smallstep/certificates/api"
+	"github.com/smallstep/certificates/authority/admin"
+	"go.step.sm/linkedca"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// CreateExternalAccountKeyRequest is the type for POST /admin/acme/eab requests
+type CreateExternalAccountKeyRequest struct {
+	ProvisionerName string `json:"provisioner"`
```
dopey, line 15: I think we can just use `Provisioner` instead of `ProvisionerName` since a name is now unique for provisioners within an authority.

authority/admin/api/acme.go
```diff
@@ -0,0 +1,113 @@
...
+		api.WriteError(w, admin.WrapError(admin.ErrorBadRequestType, err, "error reading request body"))
+		return
+	}
+
+	if err := body.Validate(); err != nil {
+		api.WriteError(w, err)
+		return
+	}
+
+	eak, err := h.acmeDB.CreateExternalAccountKey(r.Context(), body.ProvisionerName, body.Name)
```
dopey, line 49: Why are EAKs associated with a provisioner? I'm not asking because I think they shouldn't be, just want to make sure I understand.

go.mod
```diff
@@ -43,4 +43,6 @@ require (
 // replace go.step.sm/cli-utils => ../cli-utils
 // replace go.step.sm/linkedca => ../linkedca
 
+//replace go.step.sm/linkedca => ../linkedca
```
dopey, line 46: isn't this already in the file? just above here.

hslatman: Must've been a left over. Fixed now.

acme/api/account.go
hslatman: It is, indeed. Good find.

acme/api/account.go
hslatman: Within the lookup an acme.Error (`acme.ErrorUnauthorizedType`) is returned if the EAB credential does not belong to the provisioner. The additional type check allows an ACME client to interpret this specific error more clearly than without it. There may be a cleaner way to do this, but I haven't given that too much thought yet.

acme/db/nosql/account.go
hslatman: For future reference and as discussed on Slack: a user can enable or disable EAB per provisioner. It made sense to me to also link creation of the credentials to a provisioner, so that there are less surprises.

acme/db/nosql/account.go
hslatman: In this case I thought it would make sense to be able to indicate that a key could not be found. Didn't know about `db.db.Update()` yet, but it seems that it might able to do return a similar error. Will take a look.

acme/db/nosql/account.go
hslatman: Fixed in 02cd3b6b3bf763662b24bef35279b07dd000b107.

authority/admin/api/acme.go
hslatman: Renamed in 02cd3b6b3bf763662b24bef35279b07dd000b107.

ca/adminClient.go
```diff
@@ -664,13 +668,13 @@ retry:
 }
 
 // GetExternalAccountKeys returns all ACME EAB Keys from the GET /admin/acme/eab request to the CA.
-func (c *AdminClient) GetExternalAccountKeys(provisionerName string, opts ...AdminOption) ([]*linkedca.EABKey, error) {
+func (c *AdminClient) GetExternalAccountKeys(provisionerName, reference string, opts ...AdminOption) ([]*linkedca.EABKey, error) {
```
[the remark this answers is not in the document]
hslatman: @dopey: had another look at this. This also makes use of a call to `List()` to loop through all entries and could thus also go OOM. There will likely be less EAB credentials than authorizations, though, so I don't think the issue will happen soon. This could probably also benefit from the `Iterate()` approach described in #767.

hslatman:
@dopey: this PR now also includes tests for the Admin API (Admin, Provisioner, EAB). Thought it would be good to add those, so that new stuff can be added on top of the existing functionality.

acme/db/nosql/account.go
```diff
@@ -259,21 +259,42 @@ func (db *DB) DeleteExternalAccountKey(ctx context.Context, provisionerName, key
...
+	for _, entry := range entries { // entries is sorted alphabetically on the key (ID) of the EAK; no need to sort this again.
 		dbeak := new(dbExternalAccountKey)
 		if err = json.Unmarshal(entry.Value, dbeak); err != nil {
-			return nil, errors.Wrapf(err, "error unmarshaling external account key %s into ExternalAccountKey", string(entry.Key))
+			return nil, "", errors.Wrapf(err, "error unmarshaling external account key %s into ExternalAccountKey", string(entry.Key))
 		}
 		if dbeak.Provisioner != provisionerName {
 			continue
 		}
+		// skip the IDs not matching the cursor to look for in the sorted list.
```
dopey, line 286: Are we confident that `list` returns a consistent order of items?

acme/db/nosql/account.go
```diff
@@ -259,21 +259,42 @@ func (db *DB) DeleteExternalAccountKey(ctx context.Context, provisionerName, key
...
+		if cursor != "" && !foundCursorKey && cursor != dbeak.ID {
```
dopey, line 287: This logic was kinda difficult for me to follow and I wonder if it wouldn't be better to spit up the conditionals a bit:

```
if cursor != "" && !foundCursorKey {
    if cursor == dbeak.ID {
        founcCursorKey = true
    } else {
        continue
    }
}
```

This also cuts down on a few comparisons in a loop.

api/api.go
```diff
@@ -48,6 +51,21 @@ type Authority interface {
 	Version() authority.Version
 }
 
+type LinkedAuthority interface { // TODO(hs): name is not great; it is related to LinkedCA, though
```
dopey, line 54: I think it might make more sense to name this something associated with Admin (not linked). Everything is kinda related to linkedCA, because everything has to work as a linkedCA as well.

Maybe `AdminAuthority` or something with the word admin, since it is the admin API.

api/api.go
```diff
@@ -457,3 +475,296 @@ func fmtPublicKey(cert *x509.Certificate) string {
 	}
 	return fmt.Sprintf("%s %s", cert.PublicKeyAlgorithm, params)
 }
+
+type MockAuthority struct {
```
dopey, line 479: I think you're mocking `LinkedAuthority` here, right? I would say rename it to something else (as discussed in a note above) and then is there an existing MockAuthority type that you can pull in. Kinda sucks to have to mock out all the Authority methods you're not planning on testing. Might be better to have a separate MockAuthority that you pull into this struct. Or a real struct that implements the Authority interface, just so you don't have to mock out all these unnecessary methods.

dopey: Looks like we already had one of those in api_test.go. So, yea I would just split the mocks up into two. Is there any reason for them to be defined in the public facing files? Do you have tests that access this mock from other packages?

authority/admin/api/acme.go
```diff
@@ -49,7 +54,7 @@ func (h *Handler) requireEABEnabled(next nextHTTP) nextHTTP {
 
 // provisionerHasEABEnabled determines if the "requireEAB" setting for an ACME
 // provisioner is set to true and thus has EAB enabled.
-func (h *Handler) provisionerHasEABEnabled(ctx context.Context, provisionerName string) (bool, error) {
+func (h *Handler) provisionerHasEABEnabled(ctx context.Context, provisionerName string) (bool, *admin.Error) {
```
dopey, line 57: Why return a specific error type from this method? 99.99% of the time we return golang default error. Just want to make sure that if we're going to do something different, that there's a very good reason.

authority/admin/api/handler.go
```diff
@@ -3,19 +3,18 @@ package api
...
-	"github.com/smallstep/certificates/authority"
 	"github.com/smallstep/certificates/authority/admin"
 )
 
-// Handler is the ACME API request handler.
+// Handler is the Admin API request handler.
 type Handler struct {
 	db     admin.DB
-	auth   *authority.Authority
+	auth   api.LinkedAuthority // was: *authority.Authority
```
dopey, line 12: maybe AdminAuthority?

dopey (changes requested):
See comments.

authority/admin/api/handler.go
hslatman: Made it `adminAuthority`.

authority/admin/api/acme.go
hslatman: Good spot. It must've been left over from an earlier iteration. `error` is fine in this case.

api/api.go
hslatman: Move the old one back to `api_test.go` and created a new `mockAdminAuthority` for the `Admin` APIs.

acme/db/nosql/account.go
hslatman: Agree, your version is simpler. Fixed in 11a7f011778a4f771f838cef4e8635a2e73cea25.

acme/db/nosql/account.go
hslatman: `bbolt` uses byte-sorted ordering (on key), `badger` is sorted lexicographically (on key) and the `mysql` implementation (most likely) returns the results in order of primary key ascending, if I'm correct. Within a single call to `List`, and thus within a single database transaction, the ordering of the items is consistent. 

It's a different story if we expect multiple systems calling the APIs for creation and retrieval of EAKs at the same time. The keys are random, so new items will get a spot in between other items between calls to List. While paging through the items, this may thus cause an inconsistent view of the data. If we want to prevent this, we'd need to add an index using `CreatedAt` as the key and loop through the items based on that ordering. WDYT?

acme/db/nosql/eab.go
```diff
@@ -0,0 +1,365 @@
...
+	dbExternalAccountKeyReference := new(dbExternalAccountKeyReference)
+	if err := json.Unmarshal(k, dbExternalAccountKeyReference); err != nil {
+		return nil, errors.Wrapf(err, "error unmarshaling ACME EAB key for reference %s", reference)
+	}
+
+	return db.GetExternalAccountKey(ctx, provisionerID, dbExternalAccountKeyReference.ExternalAccountKeyID)
+}
+
+func (db *DB) UpdateExternalAccountKey(ctx context.Context, provisionerID string, eak *acme.ExternalAccountKey) error {
+	externalAccountKeyMutex.Lock()
```
dopey, line 230: Why are we using mutexes? And if they are necessary will this be (yet another) a problem when running multiple open source CAs that read/write from the same DB?

acme/db/nosql/eab.go
```diff
@@ -0,0 +1,365 @@
...
+		KeyBytes:      eak.KeyBytes,
+		CreatedAt:     eak.CreatedAt,
+		BoundAt:       eak.BoundAt,
+	}
+
+	return db.save(ctx, nu.ID, nu, old, "external_account_key", externalAccountKeyTable)
+}
+
+func (db *DB) addEAKID(ctx context.Context, provisionerID, eakID string) error {
+	referencesByProvisionerIndexMutex.Lock()
```
dopey, line 264: I can see that we might want the mutex here, but I want to understand why we need them for the other reads/writes/updates. Like if we're just doing a lookup by an ID. We don't generally use mutexes in other places for this.

acme/db/nosql/nosql.go
```diff
@@ -11,17 +11,18 @@ import (
...
+	accountByKeyIDTable                       = []byte("acme_keyID_accountID_index")
+	authzTable                                = []byte("acme_authzs")
+	challengeTable                            = []byte("acme_challenges")
+	nonceTable                                = []byte("nonces")
+	orderTable                                = []byte("acme_orders")
+	ordersByAccountIDTable                    = []byte("acme_account_orders_index")
+	certTable                                 = []byte("acme_certs")
+	certBySerialTable                         = []byte("acme_serial_certs_index")
+	externalAccountKeyTable                   = []byte("acme_external_account_keys")
+	externalAccountKeysByReferenceTable       = []byte("acme_external_account_key_reference_index")
```
dopey, line 24: it's an external_account_keyID by reference, right? You use keyID in the table and variable name below, but not in this one.

acme/db/nosql/eab.go
hslatman: Added them in these other places out of an abundance of caution. For the ones in `eakEADID` and removeEAKID` I have followed your pattern, because I thought they would certainly make sense there. For the others, I thought it wouldn't harm to add them there too. 

When retrieving ACME EAB keys, of which there can be many, it can take a while before they're loaded and sent to the client. Considering that adding EAB keys may be scripted, it's likely that there will be multiple calls to adding new keys, perhaps mixed with retrievals. I wanted to have a strong guarantee that the results would always be consistent. 

The various concrete `nosql` implementations do offer transactions and provide means for a consistent view of the data (based on views), but since these implementations currently don't allow us to explicitly open and commit transactions for multiple types of entities in a single transaction, I try to mimic that using (rw) mutexes. Doing it at the API level may not be the best way to do it, but so far it worked quite well 😄. Ideally I would've liked direct control over when a DB transaction was started, the total unit of work to operate on, and then explicitly committed all at once.

acme/db/nosql/eab.go
hslatman: See explanation about mutexes in the other comment.

Regarding running multiple open source CAs: using Badger will already stop you from doing that, as only a single process can work with the DB at a single time. When using MySQL, the mutexes are not going to safe our day, because they're not distributed. We'd need a distributed locking mechanism built on top of Raft or Paxos, for example, to do so (using the current database interface). I guess the same "guarantees" apply as for HA ACME: it might just work. A (fully) distributed CA would be super interesting from an engineering perspective, though 😉  

The above would probably be quite different if the DB layer provided full control over transactions (like explicit row locking) and a consistent view of the data within that transaction, for example through MVCC.

acme/db/nosql/nosql.go
hslatman: Fixed in bf21319e76bf65b946ce5723de19f10059ee7f86.

hslatman:
In bf21319e76bf65b946ce5723de19f10059ee7f86 I also fixed a bug with persistence. I must've missed this case in other manual tests due to using an existing ACME provisioner. It now seems to work correct doing multiple cycles of adding and removing keys.

dopey (approved):
lgtm

PR #654, opened 2021-07-21
Fix needs-renewal condition and switch to using ExecCondition

tashian:
I broke the renewal condition with my previous PR 🥺 
So, this fixes it. (And I actually tested it this time.)


dopey (approved):
lgtm

PR #657, opened 2021-07-28
RA install script

tashian:
See also: https://www.notion.so/smallstep/Use-ACME-with-Certificate-Manage-3add740028fe49569ce9c4a9f28291ce

### DIFF
--- a/scripts/install-step-ra.sh
+++ b/scripts/install-step-ra.sh
@@ -1,0 +1,253 @@
+#!/bin/bash
+set -e
+
+echo "This script will install and start a step-ca server running in Registration Authority (RA) mode."
+echo ""
+echo "You will need an upstream CA (URL and fingerprint)"
+echo "Don't have a CA? Sign up for a hosted CA at smallstep.com — or run your own."
+echo ""
+
+# Fail if this script is not run as root.
+if ! [ $(id -u) = 0 ]; then
+  echo "This script must be run as root"
+  exit 1
+fi
+
+# Architecture detection
+arch=$(uname -m)
+case $arch in
+  x86_64) arch="amd64" ;;
+  x86) arch="386" ;;
+  i686) arch="386" ;;
+  i386) arch="386" ;;
+  aarch64) arch="arm64" ;;
+  armv5*) arch="armv5" ;;
+  armv6*) arch="armv6" ;;
+  armv7*) arch="armv7" ;;
+esac
+
+if [ "$arch" = "armv5" ]; then
+  echo "This script doesn't work on armv5 machines"
+  exit 1
+fi
+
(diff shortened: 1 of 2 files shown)


scripts/install-step-ra.sh
```diff
@@ -0,0 +1,253 @@
...
+  echo "This script requires the jq commmand; please install it."
+  exit 1
+fi
+
+if ! hash curl &> /dev/null; then
+  echo "This script requires the curl commmand; please install it."
+  exit 1
+fi
+
+if ! hash tar &> /dev/null; then
```
maraino, line 44: It will also require gzip :)

scripts/install-step-ra.sh
```diff
@@ -0,0 +1,253 @@
...
+done
+
+# Install step
+if ! hash step &> /dev/null; then
+  echo "Installing 'step' in /usr/bin..."
+  STEP_VERSION=$(curl -s https://api.github.com/repos/smallstep/cli/releases/latest | jq -r '.tag_name')
+
+  curl -sLO https://github.com/smallstep/cli/releases/download/$STEP_VERSION/step_linux_${STEP_VERSION:1}_$arch.tar.gz
+  tar xvzf step_linux_${STEP_VERSION:1}_$arch.tar.gz
+  install -m 0755 -t /usr/bin step_${STEP_VERSION:1}/bin/step
```
maraino, line 94: I'm more of a `/usr/local/bin` guy for those packages that are not installed using a package manager.

scripts/install-step-ra.sh
```diff
@@ -0,0 +1,253 @@
...
+
+cat <<EOF > $(step path)/config/ca.json
+{
+  "address": "$RA_ADDRESS",
+  "dnsNames": ["$RA_DNS_NAMES"],
+  "db": {
+    "type": "badgerV2",
+    "dataSource": "/etc/step-ca/db"
+  },
+  "logger": {"format": "text"},
```
maraino, line 193: format json?

maraino (approved):
I've added a few recommendations, but it lgtm

scripts/install-step-ra.sh
tashian: I agree. @dopey and I had to make a call at some point, about where `step-ca` should live, because the path is hardcoded into the systemd unit and in one or two places in the docs. Maybe there's a way around this? I used to use `/usr/local/bin`, but since the Debian package installs it in `/usr/bin` we started using that everywhere. :(

scripts/install-step-ra.sh
tashian: Could you elaborate on this suggestion?

PR #672, opened 2021-08-11
Allow the reuse of azure tokens if DisableTrustOnFirstUse is true

### LINKED ISSUE #656: Azure provisioner does not allow for token reuse
### Subject of the issue
When attempting to issue two certificates from a single Azure Virtual Machine, the second attempt fails with an authorization error: `The request lacked necessary authorization to be completed. Please see the certificate authority logs for more info.`
The Azure provisioner is configured with `disableTrustOnFirstUse` to `true`
```
{
    "type": "Azure",
    "name": "AZURE IID Provisioner",
    "tenantID": "<redacted>",
    "resourceGroups": [],
    "claims": {
        "disableRenewal": false
    },
    "disableCustomSANs": false,
    "disableTrustOnFirstUse": true
}
```

### Your environment
* OS - Debian 10 (client and server)
* step-ca (server): Smallstep CA/0.15.14 (linux/amd64)
* step (client): Smallstep CLI/0.16.0 (linux/amd64)

### Steps to reproduce
- Deploy a CA server and configure an Azure Provisioner (with the same configuration as above).
- Deploy a Virtual Machine in Azure with System Identity enabled.
- Login to the machine, bootstrap the CA configuration
- Use `step` to issue certificate 1 -> Works
- Use `step` to issue certiticate 2 -> Does not work

### Expected behaviour
The second attempt to generate a certificate should have succeed 

### Actual behaviour
The second attempt fails with an authorization error: `The request lacked necessary authorization to be completed. Please see the certificate authority logs for more info.`

### Additional context
After looking through the source, it seems that Azure does not have a `jti` in the identity token so https://github.com/smallstep/certificates/blob/v0.15.14/authority/provisioner/azure.go#L126 will return an empty string.
The `authorizeToken` function in https://github.com/smallstep/certificates/blob/b7f8f15619a186e3e0aec05fa9545cb11649892c/authority/authorize.go#L77-L80 correctly handles this case where the `reuseKey` is empty, however Azure does not issue a new token for every call (each token is valid for 24 hours) to the identity API https://github.com/smallstep/certificates/blob/v0.15.14/authority/provisioner/azure.go#L26.
The hash generated at https://github.com/smallstep/certificates/blob/b7f8f15619a186e3e0aec05fa9545cb11649892c/authority/authorize.go#L78 will always be the same until the token is refreshed in Azure Identity API.

Did not find any information on forcing a token renew on Azure. Happy to contribute a fix if someone has an idea to get a unique id to prevent the token reuse issue.

maraino:
### Description

This PR allows the reuse of Azure tokens if DisableTrustOnFirstUse is true

Azure caches a token for up to 24h, and we cannot issue a new certificate for the same instance in that period of time.

The meaning of this parameter is to allow the signing of multiple certificates in one instance. This is possible in GCP, because we get a new token, and is possible in AWS because we can generate a new one. On Azure, there was no other way to do it unless you wait for 24h.

Fixes #656

### DIFF
--- a/authority/provisioner/provisioner.go
+++ b/authority/provisioner/provisioner.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/x509"
 	"encoding/json"
+	stderrors "errors"
 	"net/url"
 	"regexp"
 	"strings"
@@ -31,6 +32,17 @@ type Interface interface {
 	AuthorizeSSHRenew(ctx context.Context, token string) (*ssh.Certificate, error)
 	AuthorizeSSHRekey(ctx context.Context, token string) (*ssh.Certificate, []SignOption, error)
 }
+
+// ErrAllowTokenReuse is an error that is returned by provisioners that allows
+// the reuse of tokens.
+//
+// This is for example returned by the Azure provisioner when
+// DisableTrustOnFirstUse is set to true. For AWS and GCP DisableTrustOnFirst
+// use means that we allow the re-use of a token coming from a specific
+// instance, but in these providers we can always get a new token, but because
+// Azure caches the token for up to 24h we should add a mechanism to allow the
+// re-use.
+var ErrAllowTokenReuse = stderrors.New("allow token reuse")
 
 // Audiences stores all supported audiences by request type.
 type Audiences struct {
--- a/authority/provisioner/provisioner.go
+++ b/authority/provisioner/provisioner.go
@@ -36,12 +36,12 @@ type Interface interface {
 // ErrAllowTokenReuse is an error that is returned by provisioners that allows
 // the reuse of tokens.
 //
-// This is for example returned by the Azure provisioner when
-// DisableTrustOnFirstUse is set to true. For AWS and GCP DisableTrustOnFirst
-// use means that we allow the re-use of a token coming from a specific
-// instance, but in these providers we can always get a new token, but because
-// Azure caches the token for up to 24h we should add a mechanism to allow the
-// re-use.
+// This is, for example, returned by the Azure provisioner when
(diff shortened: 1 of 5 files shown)


authority/provisioner/provisioner.go
```diff
@@ -32,6 +33,17 @@ type Interface interface {
 	AuthorizeSSHRekey(ctx context.Context, token string) (*ssh.Certificate, []SignOption, error)
 }
 
+// ErrAllowTokenReuse is an error that is returned by provisioners that allows
+// the reuse of tokens.
+//
+// This is for example returned by the Azure provisioner when
+// DisableTrustOnFirstUse is set to true. For AWS and GCP DisableTrustOnFirst
```
dopey, line 40: DisableTrustOnFirstUse last use is split into two words.

authority/provisioner/provisioner.go
```diff
@@ -32,6 +33,17 @@ type Interface interface {
...
+// use means that we allow the re-use of a token coming from a specific
+// instance, but in these providers we can always get a new token, but because
+// Azure caches the token for up to 24h we should add a mechanism to allow the
+// re-use.
```
dopey, line 44: This is, for example, returned by the Azure provisioner when DisableTrustOnFirstUse is set to true. Azure caches tokens for up to 24hr and has no mechanism for getting a different token - this can be an issue when rebooting a VM. In contrast, AWS and GCP have facilities for requesting a new token. Therefore, for the Azure provisioner we are enabling token reuse, with the understanding that we are not following security best practices.

dopey (approved):
Considering replacing your paragraph with the one I wrote.

PR #677, opened 2021-08-17
Go 1.17 compatibility

maraino:
### Description

This PR fixes step-ca compatibility issues with Go 1.17. Compatibility changes are:
* Go 1.17 introduces a change in the net/url package disallowing the use of a semicolon (;) in URL queries. We used `url.ParseQuery` to decode the opaque string that is semicolon-separated. Now before decoding the opaque string we replace the semicolon with an ampersand. See https://golang.org/doc/go1.17#semicolons
* In Go 1.17 `tls.Dial` will fail if client and server configured protocols do not overlap. See https://golang.org/doc/go1.17#ALPN. The ALPN challenge tests that check that the negotiated protocol is `acme-tls/1` were failing because of this.  This change detects the error `no_application_protocol (120)` and fails as it was failing before.
* There's a change in the error message on one test case, with Go 1.17 the error returned is context exceeded instead of the one in the test.

### DIFF
--- a/acme/challenge.go
+++ b/acme/challenge.go
@@ -129,6 +129,12 @@ func tlsalpn01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSON
 
 	conn, err := vo.TLSDial("tcp", hostPort, config)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


dopey (approved):
lgtm

PR #680, opened 2021-08-23
Identity certificate lifetime

maraino:
### Description

This PR enforces that the identity cert lifetime matches the ssh certificate lifetime on ssh renewals.

### DIFF
--- a/api/sshRenew.go
+++ b/api/sshRenew.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"net/http"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/smallstep/certificates/authority/provisioner"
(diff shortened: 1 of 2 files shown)


api/sshRenew.go
```diff
@@ -74,13 +79,24 @@ func (h *caHandler) SSHRenew(w http.ResponseWriter, r *http.Request) {
...
+	cert := r.TLS.PeerCertificates[0]
+
+	// Enforce the cert to match another certificate, for example an ssh
+	// certificate.
+	if !notBefore.IsZero() {
+		cert.NotBefore = notBefore
+	}
+	if !notAfter.IsZero() {
+		cert.NotAfter = notAfter
+	}
```
dopey, line 97: Should we be using a deep copy of this certificate here? Just in case other downstream entities use this certificate.

PR #687, opened 2021-08-30
Bump golangci-lint to latest

dopey:


maraino (approved):
LGTM, I'm using the same one

PR #689, opened 2021-08-30
Azure OIDC

maraino:
### Description

This PR adds an extra way to distinguish Azure and Azure OIDC tokens.

We used to distinguish these tokens using the `azp` claim, but this claim does not appear on new azure oidc tokens, at least on some configurations.

This change will attempt to load by the audience (client id) if the token contains an email, required for OIDC

### DIFF
--- a/authority/authority_test.go
+++ b/authority/authority_test.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"reflect"
 	"testing"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/smallstep/assert"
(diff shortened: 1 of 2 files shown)


authority/authority_test.go
```diff
@@ -82,6 +83,10 @@ func testAuthority(t *testing.T, opts ...Option) *Authority {
 	}
 	a, err := New(c, opts...)
 	assert.FatalError(t, err)
+	// Avoid errors when test tokens are created before the test authority. This
+	// happens in some tests where we re-create the same authority to test
+	// special cases without re-creating the token.
+	a.startTime = a.startTime.Add(-1 * time.Minute)
```
dopey, line 89: yay! is this the thing that's been causing errant test failures?

dopey (approved):
lgtm

PR #692, opened 2021-08-31
Context management

dopey:


templates/values.go
```diff
@@ -67,18 +67,21 @@ var DefaultSSHTemplateData = map[string]string{
...
 	Include "{{.User.StepPath}}/ssh/config"
-{{- end }}`,
+{{ end }}# end`,
 
 	// config.tpl is the step ssh config file, it includes the Match rule and
 	// references the step known_hosts file.
 	//
 	// Note: on windows ProxyCommand requires the full path
-	"config.tpl": `Match exec "step ssh check-host %h"
+	"config.tpl": `Match exec "step ssh{{- if .User.Context }} --context {{ .User.Context }}{{- end }} check-host %h"
```
maraino, line 84: The `step ssh proxycommand` might also require the `--context` right?

templates/templates.go
```diff
@@ -257,11 +259,17 @@ func (o *Output) Write() error {
...
 
-	if o.Type == File {
+	switch o.Type {
+	case File:
 		return fileutil.WriteFile(path, o.Content, 0600)
+	case Snippet:
+		return fileutil.WriteSnippet(path, o.Content, 0600)
+	case Fragment:
+		lines := strings.Split(string(o.Content), "\n")
+		return fileutil.WriteFragment(path, o.Content, lines[0], lines[len(lines)-1], 0600)
```
maraino, line 269: Instead of having the header/footer in the template can we just add it here:
```go
header := "# authority " + authorityName + " - autogenerated by step"
footer := "# end"
fileutil.WriteFragment(path, o.Content, header, footer, 0600)
```

I'm not sure if we're able to get the authority name here.

templates/values.go
```diff
@@ -67,18 +67,21 @@ var DefaultSSHTemplateData = map[string]string{
...
+	"include.tpl": `{{- if .User.Authority }}# {{ .User.Authority }}
+{{ end }}# autogenerated by step
+# @ {{ now }}
+Host *
 {{- if or .User.GOOS "none" | eq "windows" }}
 	Include "{{ .User.StepPath | replace "\\" "/" | trimPrefix "C:" }}/ssh/config"
 {{- else }}
 	Include "{{.User.StepPath}}/ssh/config"
-{{- end }}`,
+{{ end }}# end`,
```
maraino, line 78: See my previous command and try to avoid the header/footers here.

maraino (changes requested):
Pretty much ok, just a few small things.

templates/values.go
```diff
@@ -64,30 +71,43 @@ var DefaultSSHTemplates = SSHTemplates{
...
 	User {{.User.User}}
 {{- end }}
 {{- if or .User.GOOS "none" | eq "windows" }}
 	UserKnownHostsFile "{{.User.StepPath}}\ssh\known_hosts"
-	ProxyCommand C:\Windows\System32\cmd.exe /c step ssh proxycommand %r %h %p
+	ProxyCommand C:\Windows\System32\cmd.exe /c step ssh proxycommand{{- if .User.Context }} --context {{ .User.Context }}{{- end }} %r %h %p
 {{- else }}
 	UserKnownHostsFile "{{.User.StepPath}}/ssh/known_hosts"
-	ProxyCommand step ssh proxycommand %r %h %p
+	ProxyCommand step ssh proxycommand{{- if .User.Context }} --context {{ .User.Context }}{{- end }} %r %h %p
```
maraino, line 110: Does this display an extra whitespace between --context and %r?

dopey: no

pki/pki.go
```diff
@@ -944,6 +949,11 @@ func (p *PKI) Save(opt ...ConfigOption) error {
 		if err = fileutil.WriteFile(p.defaults, b, 0644); err != nil {
 			return errs.FileError(err, p.defaults)
 		}
+		if p.profileDefaults != "" {
+			if err = fileutil.WriteFile(p.profileDefaults, []byte("{}"), 0644); err != nil {
+				return errs.FileError(err, p.profileDefaults)
+			}
+		}
```
maraino, line 956: This will ask if we want to overwrite that file, if we don't it will cancel the rest.

I think is going to be very common to use the same profile on multiple CAs. This scenario will probably affect us more than our users. So if the file exists we should do nothing.

templates/templates.go
```diff
@@ -20,6 +20,8 @@ type TemplateType string
 const (
 	// Snippet will mark a template as a part of a file.
 	Snippet TemplateType = "snippet"
+	// Line is a template for a single line in a file.
+	Line TemplateType = "line"
```
maraino, line 24: It's not exactly a single line as we will be prepending lines on it. Can you clarify the comment?

templates/templates.go
```diff
@@ -247,7 +249,10 @@ type Output struct {
 
 // Write writes the Output to the filesystem as a directory, file or snippet.
 func (o *Output) Write() error {
-	path := config.StepAbs(o.Path)
+	// Replace ${STEPPATH} with the base step path.
+	o.Path = strings.Replace(o.Path, "${STEPPATH}", step.BasePath(), -1)
+
+	path := step.Abs(o.Path)
```
maraino, line 255: I thought that you were going to do the replacement in `step.Abs()`, but it's ok for now.

templates/values.go
```diff
@@ -64,30 +71,43 @@ var DefaultSSHTemplates = SSHTemplates{
...
+	Include "{{ .User.StepBasePath | replace "\\" "/" | trimPrefix "C:" }}/ssh/includes"
+{{- else }}
 	Include "{{ .User.StepPath | replace "\\" "/" | trimPrefix "C:" }}/ssh/config"
+{{- end }}
+{{- else }}
+{{- if .User.Authority }}
+	Include "{{.User.StepBasePath}}/ssh/includes"
 {{- else }}
 	Include "{{.User.StepPath}}/ssh/config"
+{{- end }}
```
maraino, line 89: This is for backward compatibility with old clients right? Ideally old and new clients should always use the `~/.ssh/includes`. Is this the case?

templates/values.go
```diff
@@ -22,12 +22,19 @@ type StepSSH struct {
 var DefaultSSHTemplates = SSHTemplates{
 	User: []Template{
 		{
-			Name:         "include.tpl",
+			Name:         "base_config.tpl",
 			Type:         Snippet,
-			TemplatePath: "templates/ssh/include.tpl",
+			TemplatePath: "templates/ssh/base_config.tpl",
```
maraino, line 27: Can we leave it as `include.tpl` or perhaps `ssh_config.tpl` would be better. Base doesn't tell you anything.

pki/pki.go
maraino: And do we need this file in the helm template? You don't have any changes in helm.go.

maraino (changes requested):
A few more comments.

templates/templates.go
dopey: I need to store output on the template type.

authority/ssh.go
```diff
@@ -101,6 +101,13 @@ func (a *Authority) GetSSHConfig(ctx context.Context, typ string, data map[strin
 		if err != nil {
 			return nil, err
 		}
+
+		// Backwards compatibility for version of the cli older than v0.18.0
+		if o.Name == "step_includes.tpl" && (data == nil || data["Version"] != "v2") {
```
maraino, line 106: I would prepend Step to Version, but I don't want to make it StepVersion because that looks like the version number. StepTemplateVersion? Too long?

pki/pki.go
```diff
@@ -950,7 +960,11 @@ func (p *PKI) Save(opt ...ConfigOption) error {
 			return errs.FileError(err, p.defaults)
 		}
 		if p.profileDefaults != "" {
-			if err = fileutil.WriteFile(p.profileDefaults, []byte("{}"), 0644); err != nil {
+			if _, err := os.Stat(p.profileDefaults); os.IsNotExist(err) {
+				if err = fileutil.WriteFile(p.profileDefaults, []byte("{}"), 0644); err != nil {
```
maraino, line 964: It's weird to see 0644 where we always create everything as 0600/0700.
```suggestion
				if err = fileutil.WriteFile(p.profileDefaults, []byte("{}"), 0600); err != nil {
```

maraino (changes requested):
Just one small detail.

pki/pki.go
dopey: The `$(step path)/config/defaults.json` are written with `0644  right below this.  Should those also be `0600`? Why are those different?

authority/ssh.go
```diff
@@ -102,10 +102,14 @@ func (a *Authority) GetSSHConfig(ctx context.Context, typ string, data map[strin
...
 
-		// Backwards compatibility for version of the cli older than v0.18.0
-		if o.Name == "step_includes.tpl" && (data == nil || data["Version"] != "v2") {
-			o.Type = templates.File
-			o.Path = strings.TrimPrefix(o.Path, "${STEPPATH}/")
+		// Backwards compatibility for version of the cli older than v0.18.0.
+		// Before v0.18.0 we were not passing any value for SSHTemplateVersionKey
+		// from the cli.
+		if o.Name == "step_includes.tpl" {
+			if val, ok := data[templates.SSHTemplateVersionKey]; !ok || val == "" {
```
maraino, line 109: Looks good, but can be simpler
```suggestion
			if data[templates.SSHTemplateVersionKey] == "" {
```

maraino (approved):
Looks good, but added a simplification.

Issue #688, opened 2021-08-30
Store Attributes of Certificates in x509_certs alongside the whole certificate


mkkeffeler:
### What would you like to be added
We can't query smallstep for anything related to certs because the only thing in the DB is the bytes of the cert. Storing the cert alongside more columns like the common name, not after date, and more, would let us enable more complex queries, and optimize performance for future use cases.

### Why this is needed
Enable searching of Certificates signed by attributes of the certificate.

dopey:
Related #282, #239.

We recently released certificate observability features as part of our hosted product. Open source users can _link_ a standalone CA and monitor generated certificates (https://info.smallstep.com/certificate-manager-early-access-mvp/). We'll be adding enhanced querying, as well as CRL/OCSP in the coming weeks. Linking a single standalone CA is free. 

So, with that in mind, our focus, with respect to observability, will be improving that product. 

That said, we agree that storing more certificate attributes, so that they can be easily searched without parsing the whole certificate, would be a good thing. And, there's general consensus on enabling a SQL backend so that users can query the DB without writing bespoke ETLs. Since this is part of the product (and there is a free-tier), we're unlikely to focus internal engineering efforts here - at least in the short to mid-term. If folks from the community would like to contribute these types of enhancements, we'd be happy to help answer questions and facilitate development.

mkkeffeler:
Thanks @dopey so what would be acceptable to the team for the free tier?

To be more specific, I think we should store, notAfter, notBefore, all parts of the subject DN (so that you can query on locality separate from organizational unit, for example), SANS, list of non-critical and critical extensions (can query for certs with an extension).

Once we nail down the columns, we can enhance the logic to create the table if it doesn't exist (and also make sure if it does exist the columns match, for existing users)

dopey:
Our goal is to have everything that you mentioned available in the free tier of the hosted product. But not implemented in the open source. Let me explain a bit more.

The open source has hooks in it that allow us to replace the open source DB with one that implements the same interface. In our hosted stack (closed source) we have implemented that interface in a way that allows us to do observability, reporting, monitoring, and querying over the attributes of a cert. An open source standalone CA will be able to "link" up to our hosted product and take advantage of our proprietary implementation for free (up to a point - if you link more than one authority, you'll have to pay for it). 

[...]

PR #696, opened 2021-09-04
Use new methods in nosql to enable querying for Certificates.

mkkeffeler:
### Description
This resolves https://github.com/smallstep/certificates/issues/688 and enables users to perform a normal upgrade of smallstep, and the new schema will be migrated to. At which point users can begin to query for certificates the application has signed on a go forward basis. This is reliant on https://github.com/smallstep/nosql/pull/15, and then updating the version of the dependency used in this project.


dopey:
https://github.com/smallstep/nosql/pull/15

This relies on changes to the `nosql` package (PR linked above). We will not merge those changes (reasons given in the PR), therefore this PR would need modification as well. 

To recap, we are interested in a SQL implementation of the following interface: https://github.com/smallstep/certificates/blob/master/db/db.go#L45. The existing implementation in open source uses nosql, and is subsequently difficult to query against.

PR #697, opened 2021-09-07
Add deb checksum, add golang 1.17 for tests, build with golang 1.17

dopey:


maraino (approved):
Looks good to me, but we can probably just use 1.15 and 1.17 only. If it compiles in 1.15 it should compile in 1.16.

Right now, I think the main advantage of maintaining 1.15 compatibility, instead of doing only the last two, is that RH Go has a FIPS-140 version of Go but the last version is 1.15. 

Google's borigcrypto branch that is similar has support for 1.17 but they don't make any statements about the sutability of that branch for FIPS-140 purposes, although Rancher has a version of Kubernetes compiled using it, and it's FIPS-140 approved.

dopey:
Good to keep in mind about the FIPS stuff.

PR #698, opened 2021-09-08
CHANGELOG: use ISO-8601 dates

Slamdunk:
U.S. date format generates confusion arount the world, please use ISO-8601 dates


dopey (approved):
lgtm

PR #699, opened 2021-09-08
Converts group and subgroup to lowercase for comparison.

### LINKED ISSUE #679: Match ssh principals as case-insensitive
### Description

If a user using an OIDC provisioner has the email configured as myUser@domain.com, and the user signs as myuser@domain.com, `step-ca` will return an error because the principals passed by the cli do not match the expected ones:

```
# client
step ssh login --force --provisioner "Office 365" myuser@domain.com

# server
authority.SignSSH: ssh certificate principals does not match - got [myuser myuser@domain.com], want [myuser myUser myUser@domain.com]"
```

Alternatively, unless `--principals` is used, we might want to avoid sending principals at all if OIDC is used.

fofloinn:
Fixes #679

### Description
This should fix the above issue by converting the ssh principals to lower case for group comparison.
Please note I'm not sure what needs to be done with adding a test for this so any guidance would be appreciated. 

Thank you!

### DIFF
--- a/authority/provisioner/sign_ssh_options.go
+++ b/authority/provisioner/sign_ssh_options.go
@@ -454,20 +454,12 @@ func containsAllMembers(group, subgroup []string) bool {
 	if lsg > lg || (lg > 0 && lsg == 0) {
 		return false
 	}
-	groupLower := []string{}
-	subgroupLower := []string{}
-	for _, s := range group {
-		groupLower = append(groupLower, strings.ToLower(s))
-	}
-	for _, s := range subgroup {
-		subgroupLower = append(subgroupLower, strings.ToLower(s))
-	}
 	visit := make(map[string]struct{}, lg)
 	for i := 0; i < lg; i++ {
-		visit[groupLower[i]] = struct{}{}
+		visit[strings.ToLower(group[i])] = struct{}{}
 	}
 	for i := 0; i < lsg; i++ {
-		if _, ok := visit[subgroupLower[i]]; !ok {
+		if _, ok := visit[strings.ToLower(subgroup[i])]; !ok {
 			return false
 		}
 	}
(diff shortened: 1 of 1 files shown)


authority/provisioner/sign_ssh_options.go
```diff
@@ -453,12 +454,20 @@ func containsAllMembers(group, subgroup []string) bool {
...
+		subgroupLower = append(subgroupLower, strings.ToLower(s))
+	}
 	visit := make(map[string]struct{}, lg)
 	for i := 0; i < lg; i++ {
-		visit[group[i]] = struct{}{}
+		visit[groupLower[i]] = struct{}{}
 	}
 	for i := 0; i < lsg; i++ {
-		if _, ok := visit[subgroup[i]]; !ok {
+		if _, ok := visit[subgroupLower[i]]; !ok {
```
maraino, line 470: @fofloinn Can you just remove groupLower and subgroupLower, and just do
```
visit[strings.ToLower(group[I])] = struct{}{}
...
if _, ok := visit[strings.ToLower(subgroup[I])]; !ok {
```

maraino (changes requested):

authority/provisioner/sign_ssh_options.go
maraino: I've just pushed this changes myself.

maraino (approved):
I've pushed my suggestion. Thanks for fixing this, these quick fixes are not our main priority, and we really appreciate the community fixing this.

fofloinn:
> I've pushed my suggestion. Thanks for fixing this, these quick fixes are not our main priority, and we really appreciate the community fixing this.

No problem at all @maraino, happy to help where I can.

maraino:
@fofloinn I think we should create a new GitHub project `Community` or something similar where we can group these types of changes that are not our priority and the community can contribute.

PR #700, opened 2021-09-08
Allow to kms signers to define the SignatureAlgorithm

maraino:
### Description

This PR adds a new interface to CloudKMS Signers to define the signature algorithm.

CloudKMS keys signs data using a specific signature algorithm, in RSA keys, this can be PKCS#1 RSA or RSA-PSS, if the latter is used, x509.CreateCertificate will fail unless the template SignatureCertificate is properly set.

In contrast, AWS KMS RSA keys, are just RSA keys and can sign with PKCS#1 or RSA-PSS schemes, so right now the way to enforce one or the other is to use templates.

### DIFF
--- a/cas/softcas/softcas.go
+++ b/cas/softcas/softcas.go
@@ -68,7 +68,7 @@ func (c *SoftCAS) CreateCertificate(req *apiv1.CreateCertificateRequest) (*apiv1
 	}
 	req.Template.Issuer = c.CertificateChain[0].Subject
 
-	cert, err := x509util.CreateCertificate(req.Template, c.CertificateChain[0], req.Template.PublicKey, c.Signer)
+	cert, err := createCertificate(req.Template, c.CertificateChain[0], req.Template.PublicKey, c.Signer)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


cas/softcas/softcas.go
```diff
@@ -210,3 +210,16 @@ func (c *SoftCAS) createSigner(req *kmsapi.CreateSignerRequest) (crypto.Signer,
 	}
 	return c.KeyManager.CreateSigner(req)
 }
+
+// createCertificate sets the SignatureAlgorithm of the template if necessary
+// and calls x509util.CreateCertificate.
+func createCertificate(template, parent *x509.Certificate, pub crypto.PublicKey, signer crypto.Signer) (*x509.Certificate, error) {
+	// Signers can specify the signature algorithm. This is specially important
```
dopey, line 217: especially

cas/softcas/softcas.go
```diff
@@ -210,3 +210,16 @@ func (c *SoftCAS) createSigner(req *kmsapi.CreateSignerRequest) (crypto.Signer,
...
+	// when x509.CreateCertificates attempts to validate a RSAPSS signature.
```
dopey, line 218: Certificate -- no s on the end.

dopey (approved):
lgtm

PR #703, opened 2021-09-16
Update go.mozilla.org/pkcs7 

hslatman:
This includes the fix as described in mozilla-services/pkcs7#59, which was the reason a fork of the library was used.


dopey (approved):
lgtm

PR #707, opened 2021-09-16
Add support for setting individual password for ssh and tls keys

### LINKED ISSUE #693: Enhancement: Enable Configuration of Unique Passwords for Intermediate TLS, SSH Host and SSH Client Keys

I would like the following additional flags added to step-ca

- `--intermediate-tls-ca-key-password`
- `--ssh-host-ca-key-password`
- `--ssh-client-ca-key-password`

These new flags would override the `--password` flag for its respective key type and enable the user to specify unique passwords for each of their CA keys allowing the user to independently manage and rotate each CA key's password.

When not specified, the `--password` flag would use its current existing behavior for all key types.

### Why this is needed

Currently today, step-ca requires the user to keep the passwords for all CA keys the same.  This has three drawbacks:

- Any compromise of the CA password compromises all three keys.
- Any automated rotation plan that includes automated password rotation must synchronize rotation of all three keys or re-encrypt any keys not currently being rotated.

The current workaround for this is to operate three independent CA instances.

maraino:
### Description

This PR adds the flags `--ssh-host-password-file` and `--ssh-user-password-file`, if any of them are not set will default to the one passed in --password-file or the one in the ca.json

Fixes #693

### DIFF
--- a/authority/authority.go
+++ b/authority/authority.go
@@ -43,6 +43,8 @@ type Authority struct {
 	linkedCAToken string
 
 	// X509 CA
+	password           []byte
+	issuerPassword     []byte
 	x509CAService      cas.CertificateAuthorityService
 	rootX509Certs      []*x509.Certificate
 	rootX509CertPool   *x509.CertPool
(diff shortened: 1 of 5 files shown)


authority/authority.go
```diff
@@ -43,6 +43,8 @@ type Authority struct {
 	linkedCAToken string
 
 	// X509 CA
+	password           []byte
```
dopey, line 46: what's the difference between password and issuerPassword?

dopey: nvm, it's for the RA.

dopey (approved):
lgtm

PR #708, opened 2021-09-17
Use badgerv2 the default DB in helm

maraino:
### Description

This PR changes the default DB for helm charts to badgerv2, this is currently the default in `step ca init`.

We also use port 443 for the default `--ca-url`, as we usually access it through a k8s service. This option can be overridden using the `--with-ca-url` flag in the cli.


dopey (approved):
lgtm

PR #710, opened 2021-09-22
Upgrade go-jose.v2

### LINKED ISSUE #705: Ed25519 thumbprint canonicalization error in go-jose
### Description

Address possible issues caused by the wrong canonicalization of an Ed25519 key in `go-jose` (https://github.com/square/go-jose/commit/a10ff54)

One possible error if an Ed25519 account is used is that this condition might fail can be found here:
https://github.com/smallstep/certificates/blob/8df9f629b173182eb9c73768c87b377cb4f56d53/acme/challenge.go#L99-L106

It will fail if the thumbprint is calculated differently in the client and in the server.

Certbot, lego, and acme.sh do not use by default (and might not support) this key type

maraino:
### Description

Upgrade go.step.sm/crypto that includes an update of go-jose.v2. The new version of go-jose includes the fix of a typo in the OKP template causing bad fingerprints for Ed25519 keys.

See https://github.com/square/go-jose/commit/a10ff54e00bc6e833bf549e04ae976f0fe8ea2fd

Fixes #705


dopey (approved):
lgtm

PR #711, opened 2021-09-23
Check for admins in both emails and groups.

maraino:
### Description
This PR makes sure that admins are always checked on emails and groups. In the previous implementation, the groups were only checked on the SSH flow to decide the template to use.


dopey (approved):
lgtm

PR #715, opened 2021-09-29
Fix ssh template variables when CA is injected using options.

maraino:
### Description

This PR fixes an issue added in https://github.com/smallstep/certificates/pull/714, where templates variables are not included if you inject the SSH CAs using `WithSSHUserSigner` and `WithSSHHostSigner`.


dopey (approved):
lgtm

PR #728, opened 2021-10-27
Remove "env" and "expandenv"

maraino:
### Description
This PR removes the Sprig function `env` and `expandenv` from all the templates. It updates `go.step.sm/crypto` to do the same on the certificate templates.

At the same time, it replaces `promptui` with a fork that does not depend on github.com/juju/ansiterm, an LGPL project.


go.mod
```diff
@@ -44,6 +44,10 @@ require (
 	gopkg.in/square/go-jose.v2 v2.6.0
 )
 
+// avoid license conflict from juju/ansiterm until https://github.com/manifoldco/promptui/pull/181
+// is merged or other dependency in path currently in violation fixes compliance
+replace github.com/manifoldco/promptui => github.com/nguyer/promptui v0.8.1-0.20210517132806-70ccd4709797
```
francislavoie, line 49: `github.com/manifoldco/promptui` v0.9.0 is now released, I think this can be removed now after an update of that lib https://github.com/manifoldco/promptui/releases/tag/v0.9.0

maraino: We just got this notification, but this was changed after they released it, looking at the logs on Nov 1

francislavoie: Ah I missed that https://github.com/smallstep/certificates/commit/62a20c7db53bd055dcbb224b3da5da7c1198478a#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6

Thanks, and sorry for the noise!

PR #742, opened 2021-11-12
Pin golangci-lint to v1.43.0 and fix issues

### LINKED ISSUE #740: Golangci-lint not fixed to specific version in CI
### Golangci-lint not fixed to specific version in CI

Despite the setting in https://github.com/smallstep/certificates/blob/master/.golangci.yml#L79, it seems that golangci-lint is always using the latest version in CI. This results in linting issues being reported not related to new code that's being worked on.

In https://github.com/smallstep/certificates/pull/625 I've pinned the version to v1.41.0, which I'm using locally, and I had no new linting issues.

My proposal would be to update the linter version every once in a while, fixing all issues at that time. The hard thing is not forgetting to update it 😅 .

hslatman:
Fixes #740.

The golangci-lint version to use should be updated periodically.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        go: [ '1.15', '1.16', '1.17' ]
+        go: [ '1.16', '1.17' ]
     steps:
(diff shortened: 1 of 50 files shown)


dopey (approved):
lgtm

PR #748, opened 2021-11-17
Fix systemd renewer script

elyscape:
This corrects the systemctl command. Additionally, it simplifies the shell script used and switches to sh, which unlike bash is guaranteed to be available.

See also smallstep/docs#32.

### DIFF
--- a/systemd/cert-renewer@.service
+++ b/systemd/cert-renewer@.service
@@ -22,7 +22,7 @@ ExecStart=/usr/bin/step ca renew --force ${CERT_LOCATION} ${KEY_LOCATION}
 
 ; Try to reload or restart the systemd service that relies on this cert-renewer
 ; If the relying service doesn't exist, forge ahead.
-ExecStartPost=/usr/bin/env bash -c "if ! systemctl --quiet is-enabled %i.service ; then exit 0; fi; systemctl try-reload-or-restart %i"
+ExecStartPost=/usr/bin/env sh -c "! systemctl --quiet is-enabled %i.service || systemctl reload-or-try-restart %i.service"
 
 [Install]
 WantedBy=multi-user.target


systemd/cert-renewer@.service
```diff
@@ -22,7 +22,7 @@ ExecStart=/usr/bin/step ca renew --force ${CERT_LOCATION} ${KEY_LOCATION}
 
 ; Try to reload or restart the systemd service that relies on this cert-renewer
 ; If the relying service doesn't exist, forge ahead.
-ExecStartPost=/usr/bin/env bash -c "if ! systemctl --quiet is-enabled %i.service ; then exit 0; fi; systemctl try-reload-or-restart %i"
+ExecStartPost=/usr/bin/env sh -c "! systemctl --quiet is-enabled %i.service || systemctl reload-or-try-restart %i.service"
```
tashian, line 25: this should be `systemctl try-reload-or-restart`not `systemctl reload-or-try-restart`

tashian: and I think `is-enabled` should be `is-active`

tashian:
Thanks for the PR @elyscape 😄 just added a couple comments

tashian:
Fixed in [commit d925bc](https://github.com/smallstep/certificates/commit/d925bc6ba9611cb474d2d63bcebfd9b292ad097a)

systemd/cert-renewer@.service
elyscape: @tashian It used to be `reload-or-try-restart`, and was changed to `try-reload-or-restart` [in systemd 229][systemd]. Unfortunately, CentOS 7 uses systemd 219. The good news is that the old command is still accepted for backwards compatibility purposes.

[systemd]: https://github.com/systemd/systemd/blob/v229/NEWS#L115-L118

tashian: Thanks @elyscape for elaborating on that. It's been tricky to balance future-proofing with broad compatibility on our systemd stuff. After some research in January, I found that v229 is a good minimum version for us to support, and v239+ is what a lot of distros are running. That said, I've added a comment (in a067b3a) to the file about using `reload-or-try-restart` in <229.

elyscape: Sounds good!

PR #750, opened 2021-11-17
Report duration errors directly to the cli.

maraino:
### Description

This PR is the first part of a project to send better errors to the cli users. We will refactor the errs package in other PRs.

@dopey what do you think about the distinction between 400 and 401, are they ok, should all they be 400?

### DIFF
--- a/authority/provisioner/sign_options.go
+++ b/authority/provisioner/sign_options.go
@@ -8,12 +8,15 @@ import (
 	"crypto/x509/pkix"
 	"encoding/asn1"
 	"encoding/json"
+	"fmt"
 	"net"
+	"net/http"
 	"net/url"
 	"reflect"
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/smallstep/certificates/errs"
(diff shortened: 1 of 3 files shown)


authority/provisioner/sign_options.go
```diff
@@ -369,6 +372,28 @@ func newValidityValidator(min, max time.Duration) *validityValidator {
...
+func unauthorized(format string, args ...interface{}) error {
+	msg := fmt.Sprintf(format, args...)
+	return &errs.Error{
+		Status: http.StatusUnauthorized,
+		Msg:    msg,
+		Err:    errors.New(msg),
+	}
+}
+
 // Valid validates the certificate validity settings (notBefore/notAfter) and
```
dopey, line 397: remove extra 'and' here

dopey (approved):
lgtm, although I would return BadRequest for all of these.

authority/provisioner/sign_options.go
maraino: Fixed in acd0bac

PR #752, opened 2021-11-19
Bad request errors

maraino:
### Description

This PR is a first pass on sending better errors to the cli. This PR focuses only on 400 - bad request errors. With this change errs.BadRequest() and errs.BadRequestErr() always require a string to pass to the user.

Following PRs will change other HTTP statuses.

### DIFF
--- a/authority/ssh.go
+++ b/authority/ssh.go
@@ -94,7 +94,7 @@ func (a *Authority) GetSSHConfig(ctx context.Context, typ string, data map[strin
 
 		// Check for required variables.
 		if err := t.ValidateRequiredData(data); err != nil {
-			return nil, errs.BadRequestErr(err, errs.WithMessage("%v, please use `--set <key=value>` flag", err))
+			return nil, errs.BadRequestErr(err, "%v, please use `--set <key=value>` flag", err)
 		}
 
 		o, err := t.Output(mergedData)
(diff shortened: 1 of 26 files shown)


authority/ssh.go
```diff
@@ -258,7 +258,7 @@ func (a *Authority) SignSSH(ctx context.Context, key ssh.PublicKey, opts provisi
 // RenewSSH creates a signed SSH certificate using the old SSH certificate as a template.
 func (a *Authority) RenewSSH(ctx context.Context, oldCert *ssh.Certificate) (*ssh.Certificate, error) {
 	if oldCert.ValidAfter == 0 || oldCert.ValidBefore == 0 {
-		return nil, errs.BadRequest("renewSSH: cannot renew certificate without validity period")
+		return nil, errs.BadRequest("cannot renew a certificate without validity period")
```
dopey, line 261: it's not really necessary to add these here. I don't think error messages are expected to be grammatically correct. They're just meant to be informative. The extra `a` doesn't convey any info that isn't already there.

dopey (approved):
lgtm

authority/ssh.go
maraino: Let's keep it as it is, I think it looks more natural and this will be displayed to the user.

PR #753, opened 2021-11-21
Docker Example Fix: Remove Step Build

distributedOne:
### Description
The example for docker fails when attempting to run `make`, to get around this the base image for the `renewer` is changed from alpine to `smallstep/step-cli` and the attempt to build the `step` cli is removed from the Makefile.


maraino:
To avoid cyclic dependencies that can get complicated, I think that instead of building the binary like this:
```
CGO_ENABLED=0 GOOS=linux go build -o renewer/step github.com/smallstep/cli/cmd/step
```

We should do something like this:
```
curl -so renewer/step https://files.smallstep.com/step-linux-latest
```

distributedOne:
> To avoid cyclic dependencies that can get complicated, I think that instead of building the binary like this:
> 
> ```
> CGO_ENABLED=0 GOOS=linux go build -o renewer/step github.com/smallstep/cli/cmd/step
> ```

The dependencies really were not hard to resolve, at least not so much that it prevented me continuing on.

> We should do something like this:
> 
> ```
> curl -so renewer/step https://files.smallstep.com/step-linux-latest
> ```

From my own prospective I would prefer to build over download the finished binary from an external location.

maraino:
`smallstep/cli` already depends on `smallstep/certificates`, and if we add the reverse dependency it might make our release steps way more complicated, we had that cycle before, and we got rid of it because it was problematic.

A better option is to use our `smallstep/step-cli` as the base image instead of just alpine so `step` is already installed on the new containers.

distributedOne:
> `smallstep/cli` already depends on `smallstep/certificates`, and if we add the reverse dependency it might make our release steps way more complicated, we had that cycle before, and we got rid of it because it was problematic.
> 
> A better option is to use our `smallstep/step-cli` as the base image instead of just alpine so `step` is already installed on the new containers.

Even better, I just reverted the changes to the go.mod and go.sum as well as patched up the example as you've mentioned. The description and title to this PR has been updated to match.

maraino (approved):
Thanks @distributedOne

PR #773, opened 2021-12-13
Improve IP SANS support

hslatman:
To support IPs in the subject using `step-cli`, this PR ensures that
Subject Common Names that can be parsed as an IP are also checked
to have been authorized before.

The PR for `step-cli` is here: [github.com/smallstep/cli/pull/576](github.com/smallstep/cli/pull/576).


acme/order.go
```diff
@@ -289,14 +295,22 @@ func canonicalize(csr *x509.CertificateRequest) (canonicalized *x509.Certificate
...
-	// (ACME IP SANS)
+	// subjectAltName extension, or both. Subject Common Names that can be
+	// parsed as an IP are included as an IP address for the equality check.
+	// If these were excluded, a certificate could contain an IP as the
+	// common name without having been challenged.
 	if csr.Subject.CommonName != "" {
-		// nolint:gocritic
-		canonicalized.DNSNames = append(csr.DNSNames, csr.Subject.CommonName)
+		ip := net.ParseIP(csr.Subject.CommonName)
+		subjectIsIP := ip != nil
```
dopey, line 304: I wann have a quick discussion about syntax here. Historically we would write the line above as: 
```
if ip := net.ParseIP(blah); ip != nil {
   // do something
} else {
  // do something else
}
```

But you're breaking that first line up into three separate lines. I'm on board with breaking up code into digestible chunks, but I kinda feel like we're oversimplifying here and avoiding some useful shortcuts. Like, making `subjectIsIP` a variable seems kinda unnecessary if we're just going to consume that variable on the very next line.

Most importantly, I think we should just be consistent within the codebase. So if we'd prefer to use the slightly longer but more clear syntax (like in your PR), that's cool. We can slowly switch change the codebase over time, but maybe let's just have a discussion about it. If we're not going to use it everywhere, it feels kinda strange to only use it sometimes.

dopey (changes requested):
Everything looks good to me, but I left a comment about some golang coding convention that I think we should settle on. Let's discuss tomorrow.

dopey (approved):
lgtm

PR #774, opened 2021-12-14
Avoid doing unauthenticated requests on the SDK

maraino:
### Description

When step-ca runs with mTLS required on some endpoints, the SDK used in autocert will fail to start because the identity certificate is missing. This certificate is only required to retrieve all roots, in most cases, there's only one, and the SDK has access to it.

I've added you @hslatman and @dopey, but I only need one review.

### DIFF
--- a/ca/bootstrap.go
+++ b/ca/bootstrap.go
@@ -65,7 +65,7 @@ func BootstrapClient(ctx context.Context, token string, options ...TLSOption) (*
 		return nil, err
 	}
 
-	// Make sure the tlsConfig have all supported roots on RootCAs.
+	// Make sure the tlsConfig has all supported roots on RootCAs.
 	//
 	// The roots request is only supported if identity certificates are not
 	// required. In all cases the current root is also added after applying all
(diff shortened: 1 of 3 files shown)


hslatman (approved):
Looks OK to me, but have to admit I'm not familiar with this part of the code (yet).

I think it's possible to combine the steps in the `BootstrapXXX` into a shared function or maybe even in `Bootstrap`. No blocker, though.

Am I correct that `RequireClientAuthentication` is only set online?

ca/bootstrap.go
```diff
@@ -120,25 +118,21 @@ func BootstrapServer(ctx context.Context, token string, base *http.Server, optio
...
 
-	req, pk, err := CreateSignRequest(token)
-	if err != nil {
-		return nil, err
-	}
-
-	sign, err := client.Sign(req)
-	if err != nil {
-		return nil, err
+	// Make sure the tlsConfig have all supported roots on RootCAs.
```
dopey, line 126: Make sue the tlsConfig `has` all supported roots on RootCAs. Change in this spot and in the one above.

dopey (approved):
lgtm. Let a comments about grammar in documentation.

hslatman (approved):

PR #777, opened 2021-12-17
Implement the kms.Decrypter with PKCS#11

maraino:
### Description

This PR implements the `kms.Decrypter` interface with PKCS#11, this interface allows the use of SCEP with PKCS#11 modules.

### DIFF
--- a/kms/pkcs11/pkcs11.go
+++ b/kms/pkcs11/pkcs11.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"crypto"
 	"crypto/elliptic"
+	"crypto/rsa"
 	"crypto/x509"
 	"encoding/hex"
 	"fmt"
@@ -142,8 +143,7 @@ func (k *PKCS11) CreateKey(req *apiv1.CreateKeyRequest) (*apiv1.CreateKeyRespons
 	}, nil
 }
 
-// CreateSigner creates a signer using the key present in the PKCS#11 MODULE signature
-// slot.
+// CreateSigner creates a signer using a key present in the PKCS#11 module.
 func (k *PKCS11) CreateSigner(req *apiv1.CreateSignerRequest) (crypto.Signer, error) {
 	if req.SigningKey == "" {
 		return nil, errors.New("createSignerRequest 'signingKey' cannot be empty")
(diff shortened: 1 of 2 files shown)


kms/pkcs11/pkcs11.go
```diff
@@ -157,6 +157,27 @@ func (k *PKCS11) CreateSigner(req *apiv1.CreateSignerRequest) (crypto.Signer, er
 	return signer, nil
 }
 
+// CreateDecrypter creates a decrypter using a key present in the PKCS#11
+// module.
+func (k *PKCS11) CreateDecrypter(req *apiv1.CreateDecrypterRequest) (crypto.Decrypter, error) {
+	if req.DecryptionKey == "" {
+		return nil, errors.New("createDecrypterRequest 'decriptionKey' cannot be empty")
```
hslatman, line 164: Small typo: decriptionKey -> decryptionKey

kms/pkcs11/pkcs11.go
```diff
@@ -157,6 +157,27 @@ func (k *PKCS11) CreateSigner(req *apiv1.CreateSignerRequest) (crypto.Signer, er
...
+	}
+
+	signer, err := findSigner(k.p11, req.DecryptionKey)
```
hslatman, line 167: It may be good to introduce a `findDecrypter`, which could wrap the type assertions. Not a blocker, IMO.

kms/pkcs11/pkcs11_test.go
```diff
@@ -491,6 +492,86 @@ func TestPKCS11_CreateSigner(t *testing.T) {
...
+		name    string
+		args    args
+		wantErr bool
+	}{
+		{"RSA", args{&apiv1.CreateDecrypterRequest{
+			DecryptionKey: "pkcs11:id=7371;object=rsa-key",
+		}}, false},
+		{"RSA PSS", args{&apiv1.CreateDecrypterRequest{
+			DecryptionKey: "pkcs11:id=7372;object=rsa-pss-key",
+		}}, false},
```
hslatman, line 512: I think the test name should be `RSA OAEP`? For the key object it's probably OK to use the `rss-pss-key` object and not having to add a new one (for now).

maraino: Each test tests both pkcs#1 v1.5 and RSA OAEP, this is basically using a regular RSA key. 

On PKCS#11 there's no difference on the keys, the difference happens at signing time. But for other KMSs you create a key for RSA PSS and it always signs using PSS. This is why the API makes distinctions.

kms/pkcs11/pkcs11.go
maraino: Fixed on d5c6572

kms/pkcs11/pkcs11.go
maraino: I don't think is worthy because at the end it will duplicate findSigner or call findSigner too.

hslatman: Agree, not required.

kms/pkcs11/pkcs11_test.go
hslatman: Ah, got it. The PSS threw me off. Looks good then!

hslatman (approved):

PR #779, opened 2021-12-22
Set Step CLI User-Agent when performing ACME requests

hslatman:
This PR sets the Step CLI User-Agent generated at build time when performing ACME requests.


dopey (approved):
lgtm

PR #792, opened 2022-01-10
Improve errors related to template execution failures

### LINKED ISSUE #791: Improve surfacing certificate template errors
A user of `step-ca` reported an issue when signing a certificate using an ACME provisioner. From the error message it wasn't immediately clear what was wrong, as it only contained `error unmarshaling certificate`. Apparently this error can result from a template with bad contents, such as bad JSON. Applying the CSR data to the template when creating a new certificate then results in this error.

We can probably improve this error by being more specific about where the error comes from. Indicating that it likely has something to do with the template may be sufficient.

hslatman:
This PR (slightly) improves error messages reported by `step-ca` when an error occurs related to applying the JSON certificate template. It addresses #791. 

Before applying the changes from this PR, an error would be reported roughly like this (when using an ACME provisioner):

`
error finalizing order: error signing certificate for order REDACTED_ORDER_ID: authority.Sign: error unmarshaling certificate: invalid character '[' looking for beginning of object key string
`

With the changes from this PR, the error is shown like this (with a different syntax present):

`
error finalizing order: error signing certificate for order REDACTED_ORDER_ID: authority.Sign: failed to apply certificate template: invalid character '\\n' in string literal
`

There's definitely more possible if I change some of the internals of the `x509util` and `sshutil` packages. We could report some predefined error, which can then be type checked instead of matching based on a prefix string, as I do now. With a bit more work I can implement JSON `SyntaxError` and `UnmarshalTypeError` handling, so that we can report the actual character location at which the error occurs using our own error types. This might also be helpful for the template configurations that are read from `ca.json` (not from a filepath).

Because the template is read every time a certificate is requested (when configured as a file), checking that the template is correct is likely not going to be sufficient for templates loaded from a file, but it might still be helpful to do this and report an error when starting `step-ca`. We could also choose to make this logic consistent with the rest of the CA configuration, which is only (re)loaded at startup time or when SIGHUP'ed. That might be a bit more complex, though.

Fixes #791


authority/ssh.go
```diff
@@ -198,6 +198,14 @@ func (a *Authority) SignSSH(ctx context.Context, key ssh.PublicKey, opts provisi
...
 			)
 		}
+		// explicitly check for unmarshaling errors, which are most probably caused by JSON template syntax errors
+		if strings.HasPrefix(err.Error(), "error unmarshaling certificate") {
+			msg := strings.TrimSpace(strings.TrimPrefix(err.Error(), "error unmarshaling certificate:"))
+			return nil, errs.ApplyOptions(
+				errs.InternalServer("authority.Sign: failed to apply certificate template: %s", msg),
+				errs.WithKeyVal("signOptions", signOpts),
+			)
+		}
```
maraino, line 208: I have an unfinished PR to write better internal errors, where we send users nicer errors and write in the logs more details if available. Without my PR this internal server error will send a very opaque error to the CLI.

We can do a few changes to follow this:
* I'm removing is those `authority.Sign` headers, so remove that.
* Although in my PR is different (easier), in this code we already have a way to send a message to the CLI using the options `errs.WithMessage()` or something similar. We can send the users something saying that there was an error processing the certificate template without specifying more details.
* `errs.ApplyOptions` is something that I added on my PR to add support for this k/v details, I'm not sure if that works with "old" error types like that one.

I think we can write the return as:

```go
return nil, errs.InternalServerErr(err, 
    errs.WithKeyVal("signOptions", signOpts),
    errs.WithMessage("error processing certificate template (or something similar)"),
)
```

That will write the original error to the logs and send the user the message that we want. 

I don't think is necessary to trim the message that we write in the logs, but if you think you need you can use `errors.Cause` (in pkg/errors, returns the wrapped error). It's even possible with `errors.As` to get the `json.SyntaxError`, but there might be other types of errors, so it's not that easy.

authority/tls.go
```diff
@@ -126,6 +127,15 @@ func (a *Authority) Sign(csr *x509.CertificateRequest, signOpts provisioner.Sign
...
 		}
+		// explicitly check for unmarshaling errors, which are most probably caused by JSON template syntax errors
+		if strings.HasPrefix(err.Error(), "error unmarshaling certificate") {
+			msg := strings.TrimSpace(strings.TrimPrefix(err.Error(), "error unmarshaling certificate:"))
+			return nil, errs.ApplyOptions(
+				errs.InternalServer("authority.Sign: failed to apply certificate template: %s", msg),
+				errs.WithKeyVal("csr", csr),
+				errs.WithKeyVal("signOptions", signOpts),
+			)
+		}
```
maraino, line 138: Same comment as before.

maraino (changes requested):

authority/tls.go
```diff
@@ -549,3 +560,23 @@ func (a *Authority) GetTLSCertificate() (*tls.Certificate, error) {
...
+		typeError   *json.UnmarshalTypeError
+	)
+	if errors.As(err, &syntaxError) {
+		// offset is arguably not super clear to the user, but it's the best we can do here
+		cause = fmt.Errorf("%s at offset %d", cause.Error(), syntaxError.Offset)
+	}
+	if errors.As(err, &typeError) {
+		// slightly rewriting the default error message to include the offset
+		cause = fmt.Errorf("cannot unmarshal %s at offset %d into Go value of type %s", typeError.Value, typeError.Offset, typeError.Type)
+	}
```
maraino, line 580: It should only attempt one of these conversions. Use a switch or `else if`

maraino (approved):
LGTM but I've added a comment that we should fix.

PR #794, opened 2022-01-12
Create the db directory on step ca init

tashian:
The systemd unit has a stanza that requires the db folder to exist, or else it will not start the unit.
But `step ca init` doesn't create the db folder.
This PR makes it possible to run `step ca init` and then `systemctl start step-ca` without needing to manually mkdir the db folder first.


tashian:
@dopey I'd love a review on this one, as I believe this will unlock the possibility of distributing a systemd unit file with step-ca .deb/.rpm.

dopey (approved):
lgtm

PR #797, opened 2022-01-14
Fix macOS SCEP client issues

### LINKED ISSUE #746: macOS SCEP renewals fail over HTTPS
### Subject of the issue
The macOS SCEP client has some strange behavior. If the SCEP endpoint is served over HTTPS and accepts client certificates, it will use the certificate that it is attempting to renew, but there are two very important caveats:

1. It will only do so if it can build a certificate chain to a root that it trusts
2. It only presents the leaf certificate to the server

Item number 1 is interesting because, despite retrieving the intermediate CA certificate as part of the SCEP enrollment process, macOS does not save it to the keychain. This seems like a bug but, as a result, administrators who create configuration profiles that include a SCEP payload but not the intermediate CA cert will find their systems not being able to actually use the certificate generated by the process.

Item number 2 is a problem for us, because `step-ca` is currently written [to verify client certificates, if presented, against only the root CA certificate][verify]. As a result, when a properly-configured macOS machine attempts to renew the certificate, `step-ca` will fail to validate the presented client certificate and drop the connection.

### Your environment
* OS - macOS 12.0.1
* Version - 0.17.6

### Steps to reproduce
1. Download and extract [this demo environment](https://github.com/smallstep/certificates/files/7543568/environment.zip) on a Mac
2. From the directory into which it has been extracted, run `step-ca config/ca.json`
3. Install the configuration profile named "Sample SCEP Profile - No Chain.mobileconfig"
    1. Open the file
    2. Launch System Preferences
    3. Click "Profiles"
    4. Click on "Sample SCEP Profile - No Chain"
    5. Click "Install"
    6. Click "Install" again
4. Renew the certificate by clicking the "Update" button in the bottom-right of the window
5. Remove the profile
6. Repeat steps 3-5 with the profile named "Sample SCEP Profile - With Chain.mobileconfig"

### Expected behaviour
Renewal succeeds with either configuration profile.

### Actual behaviour
Renewal fails with the "Sample SCEP Profile - With Chain" profile. `step-ca` logs the following:
```
2021/11/15 20:17:21 /usr/local/Cellar/go/1.17.2/libexec/src/net/http/server.go:3157: http: TLS handshake error from 127.0.0.1:62298: tls: failed to verify client certificate: x509: certificate signed by unknown authority
```

[verify]: https://github.com/smallstep/certificates/blob/v0.17.6/ca/ca.go#L420-L437

hslatman:
This PR adds the intermediate CA certificate to ClientCAs for chain building and sends root certificates to clients when configured in the SCEP provisioner.

It also includes an additional knob to control the encryption algorithm for clients that don't support the default (DES-CBC). 

The SCEP provisioner has now been successfully tested with micromdm/scep, certnanny/sscep and the macOS SCEP client. The macOS SCEP client has also been tested with HTTPS requests instead of just HTTP successfully.

Fixes #746

### DIFF
--- a/authority/authority.go
+++ b/authority/authority.go
@@ -595,12 +595,6 @@ func (a *Authority) IsRevoked(sn string) (bool, error) {
 	return a.db.IsRevoked(sn)
 }
 
-// GetIntermediateCertificate returns the x509 intermediate CA
-// certificate.
-func (a *Authority) GetIntermediateCertificate() (*x509.Certificate, error) {
-	return pemutil.ReadCertificate(a.config.IntermediateCert)
-}
-
 // requiresDecrypter returns whether the Authority
 // requires a KMS that provides a crypto.Decrypter
 // Currently this is only required when SCEP is
--- a/authority/authority.go
+++ b/authority/authority.go
@@ -595,7 +595,7 @@ func (a *Authority) IsRevoked(sn string) (bool, error) {
 	return a.db.IsRevoked(sn)
 }
 
-// GetIntermediateCertificate returns the x509 CA intermediate
+// GetIntermediateCertificate returns the x509 intermediate CA
 // certificate.
 func (a *Authority) GetIntermediateCertificate() (*x509.Certificate, error) {
 	return pemutil.ReadCertificate(a.config.IntermediateCert)
--- a/authority/authority.go
+++ b/authority/authority.go
@@ -437,13 +437,6 @@ func (a *Authority) init() error {
 		}
 	}
 
-	// Check if a KMS with decryption capability is required and available
-	if a.requiresDecrypter() {
-		if _, ok := a.keyManager.(kmsapi.Decrypter); !ok {
-			return errors.New("keymanager doesn't provide crypto.Decrypter")
-		}
-	}
-
 	// TODO: decide if this is a good approach for providing the SCEP functionality
 	// It currently mirrors the logic for the x509CAService
 	if a.requiresSCEPService() && a.scepService == nil {
@@ -454,6 +447,7 @@ func (a *Authority) init() error {
 		if err != nil {
 			return err
 		}
+		options.CertificateChain = append(options.CertificateChain, a.rootX509Certs...)
 		options.Signer, err = a.keyManager.CreateSigner(&kmsapi.CreateSignerRequest{
 			SigningKey: a.config.IntermediateKey,
 			Password:   []byte(a.password),
@@ -599,6 +593,12 @@ func (a *Authority) IsRevoked(sn string) (bool, error) {
 	}
 
 	return a.db.IsRevoked(sn)
+}
+
+// GetIntermediateCertificate returns the x509 CA intermediate
+// certificate.
+func (a *Authority) GetIntermediateCertificate() (*x509.Certificate, error) {
+	return pemutil.ReadCertificate(a.config.IntermediateCert)
 }
 
 // requiresDecrypter returns whether the Authority
--- a/authority/provisioner/scep.go
+++ b/authority/provisioner/scep.go
@@ -18,13 +18,20 @@ type SCEP struct {
 	ForceCN           bool     `json:"forceCN,omitempty"`
 	ChallengePassword string   `json:"challenge,omitempty"`
 	Capabilities      []string `json:"capabilities,omitempty"`
+	// IncludeRoots makes the provisioner return the CA root(s) in the GetCACerts response
+	IncludeRoots bool `json:"includeRoots,omitempty"`
 	// MinimumPublicKeyLength is the minimum length for public keys in CSRs
-	MinimumPublicKeyLength int      `json:"minimumPublicKeyLength,omitempty"`
-	Options                *Options `json:"options,omitempty"`
-	Claims                 *Claims  `json:"claims,omitempty"`
-	claimer                *Claimer
+	MinimumPublicKeyLength int `json:"minimumPublicKeyLength,omitempty"`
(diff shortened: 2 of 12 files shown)


authority/authority.go
```diff
@@ -601,6 +595,12 @@ func (a *Authority) IsRevoked(sn string) (bool, error) {
 	return a.db.IsRevoked(sn)
 }
 
+// GetIntermediateCertificate returns the x509 intermediate CA
+// certificate.
+func (a *Authority) GetIntermediateCertificate() (*x509.Certificate, error) {
+	return pemutil.ReadCertificate(a.config.IntermediateCert)
+}
```
maraino, line 602: Not all the configurations of step-ca will have an intermediate file. For example, by default, an RA will not, and SCEP is a great candidate for an RA, although it might not be supported yet because it requires an RSA certificate and key for decrypt/encrypt things.

To be future proof, we should get the intermediate from `authority.x509CAService`, but right now it does not have a way to do it directly. The other option is to get it from the step-ca own certificate, generated in `authority.GetTLSCertificate()`.

ca/ca.go
```diff
@@ -432,6 +427,19 @@ func (ca *CA) getTLSConfig(auth *authority.Authority) (*tls.Config, error) {
...
+		certPool.AddCert(crt)
+	}
+
+	// adding the intermediate CA to the pool will allow clients that do not
+	// send the intermediate for chain building to connect to the CA successfully.
+	cert, err := auth.GetIntermediateCertificate()
+	if err != nil {
+		return nil, err
+	}
+	certPool.AddCert(cert)
```
maraino, line 441: As an easy fix, without requiring that new method or changes in the CAS services, we can get the cert from the tlsCert from above.

authority/provisioner/scep.go
```diff
@@ -18,13 +18,20 @@ type SCEP struct {
 	ForceCN           bool     `json:"forceCN,omitempty"`
 	ChallengePassword string   `json:"challenge,omitempty"`
 	Capabilities      []string `json:"capabilities,omitempty"`
+	// IncludeRoots makes the provisioner return the CA root(s) in the GetCACerts response
+	IncludeRoots bool `json:"includeRoots,omitempty"`
```
maraino, line 22: I believe this should be called `IncludeRoot`, the ca.json supports multiple roots for rotation purposes, but only one is really in use.

scep/authority.go
```diff
@@ -153,14 +151,28 @@ func (a *Authority) GetCACertificates() ([]*x509.Certificate, error) {
...
 
-	return []*x509.Certificate{a.intermediateCertificate}, nil
+	certs := []*x509.Certificate{}
+	certs = append(certs, a.caCerts[0])
+
+	// NOTE: we're adding the CA roots here, but they are (highly likely) different than what the RFC means.
+	// Clients are responsible to select the right cert(s) to use, though.
+	if p.ShouldIncludeRootsInChain() && len(a.caCerts) > 1 {
+		certs = append(certs, a.caCerts[1:]...)
+	}
```
maraino, line 173: Perhaps we should only add the actual root and not all of them.

maraino (changes requested):
Although it works for SCEP that's not the proper way to get the intermediate certificate. We don't really have one proper way, but for this case, we can also hack it and get it from the server cert.

ca/ca.go
maraino: We should also include all intermediates, as there might be more than one.

scep/authority.go
hslatman: Changed in https://github.com/smallstep/certificates/pull/797/commits/64680bb16dd076a33e3c1ad53b71b507825c6560. 

I still have my doubts if what we're doing here is in fact what is expected in SCEP, as what they refer to in the RFC as CA seems to be our intermediate CA, but this change shouldn't break things for people already using SCEP.

ca/ca.go
hslatman: Fixed in https://github.com/smallstep/certificates/pull/797/commits/64680bb16dd076a33e3c1ad53b71b507825c6560.

authority/authority.go
hslatman: There are indeed improvements to how SCEP can be configured and used. I think that should be combined in a bigger package of work related to SCEP after discussing what types of use cases to support and how.

scep/authority.go
maraino: This works in most cases, but it's not strictly right that the first root in `authority.rootX509Certs` is the root that signed the intermediate. But as this special case must be configured in the provisioner, we can force the user to configure the roots accordingly.

maraino (changes requested):
All ok, but as we have included changes in the provisioner we need to do a couple of changes more:
* Add field to linkedca https://github.com/smallstep/linkedca/blob/be128d16de4d98ef02c31f6deef571997a781e1b/provisioners.proto#L138-L143
* Add changes in ProvisionerToCertificates and ProvisionerToLinkedca in https://github.com/smallstep/certificates/blob/master/authority/provisioners.go

hslatman:
@maraino: added the additional fields and updated the implementation. Also opened a PR  for SCEP provisioner CRUD support in the CLI: https://github.com/smallstep/cli/pull/616. 

In the process I had to update the `grpc.WithInsecure` option too, which was deprecated recently.

maraino (approved):
LGTM: are you still using `LoadProvisionerByID`?

hslatman:
@maraino: no, I changed it to the name variant to make it work with LinkedCA. Should I use the ID one instead? Or are you asking because if I don't use it, I should remove it completely?

PR #798, opened 2022-01-14
Add Hashicorp Vault RA

### LINKED ISSUE #642: Write a Vault RA Backend
https://github.com/hashicorp/vault/issues/8690
Original Vault Issue Comment: https://github.com/hashicorp/vault/issues/8690#issuecomment-678442610

Old Issue that tracked the completion of the RA API which enables a plugin like architecture for new backends: https://github.com/smallstep/certificates/issues/343

ahmet2mir:
Hello,

I added a Vault Integration with RA mode, but I had some questions/remarks:

1. Vault doesn't allow a "random" when doing domain restriction on sign.
If I allow SAN  with domaine example.com, I could allow subdomain but the CN must part of it.
(I had the same issue when creating certificate on Sectigo.)

But `authority/tls/GetTLSCertificate` hardcode `Step Online CA` as CN value, I added CommonName into global config. So I could set 
```json
  "dnsNames": [ "ra.example.com", "other.example.com" ],
  "commonName": "cn.example.com"
```
And probably `ASN1DN` in future provides other attributes like country etc to use for the CA HTTPS server.

=> I could move it into dedicated MR (currently it's a separate commit)

2. Currently `options` contains a lot of parameters, some common params between different CAS, some for Google CAS, other for StepCAS.
Here I come with my vault params, tomorrow if you add another CAS you will have different params inside the same "level" of config.

I'm more conviced that each CAS could define a config struct and "register" it into options to avoid 1000 params in future.

An just keep "common" params in options, so I added a generic `map[string]interface{}` structure and let the VaultCAS process it via `loadOptions` into custom structure `VaultOptions`

=> I could move it into dedicated MR (currently it's a separate commit)

The json config become more clear for users:

Example GoogleCAS become

```json
"authority": {
    "type": "cloudcas",
    ...
    "config": {
        "credentialsFile": "xxxxxx",
        "project": "xxxxxx",
        "location": "xxxxxx",
        "caPool": "xxxxxx",
        "caPoolTier": "xxxxxx",
        "gCSBucket": "xxxxxx",
    },
  },
```

3. It's not clear how https CA manage certificate for HTTPS Server.
After each restart it generate a CSR and CAS had to sign it.
But when stopped it doesn't ask for revoke.
So when doing my dev tests, I have several certificate signed which are not marked as revoked.
I had to manually clean them, on dev it's easy to clean up but on prod I will not for sure verify if each cert are used or not (even if scripting a clean up is not that hard).
Or is there a way to persist the signed cert and avoid resign at all restart ?

4. Finally here we are, I wrote the vaultcas backend also prepared the website [documentation](https://github.com/ahmet2mir/docs/commit/8dd50841238153cbaf1ebeffc9dc8ddc4242db48)

For tests part, well, I just write a basic test before continuing. It's a smallstep :)

I'm not sure that I took the right path or the way of the cas is wrote.
If you agree I could continue, let me know.

💔Thank you very very much!

Resolve #642


maraino:
Thanks, @ahmet2mir great contribution, I'll be probably looking at this during this week.

cas/vaultcas/vaultcas.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	serialNumber := req.SerialNumber
+	if req.Certificate != nil {
+		serialNumber = req.Certificate.SerialNumber.String()
+	}
+
+	serialNumberDash := strings.ReplaceAll(serialNumber, ":", "-")
+
+	_, err := v.client.Logical().Write(v.config.PKI+"/revoke/"+serialNumberDash, nil)
```
maraino, line 313: It looks like Vault expects the serial number in hexadecimal format, you will get it as a big decimal number, you can use the big.Int to convert it to bytes and then encode it as vault expects.

cas/vaultcas/vaultcas.go
```diff
@@ -0,0 +1,322 @@
...
+		return New(ctx, opts)
+	})
+}
+
+type VaultOptions struct {
+	PKI             string `json:"pki,omitempty"`
+	PKIRole         string `json:"pkiRole,omitempty`
+	PKIRoleRSA      string `json:"pkiRoleRSA,omitempty`
+	PKIRoleEC       string `json:"pkiRoleEC,omitempty`
+	PKIRoleED25519  string `json:"PKIRoleED25519,omitempty`
```
maraino, line 31: Fix missing double quotes and lower case `d`
```suggestion
	PKIRoleRSA      string `json:"pkiRoleRSA,omitempty"`
	PKIRoleEC       string `json:"pkiRoleEC,omitempty"`
	PKIRoleEd25519  string `json:"PKIRoleEd25519,omitempty"`
```

cas/vaultcas/vaultcas.go
```diff
@@ -0,0 +1,322 @@
...
+	config := vault.DefaultConfig()
+	config.Address = opts.CertificateAuthority
+
+	client, err := vault.NewClient(config)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to initialize vault client")
+	}
+
+	var appRoleAuth *auth.AppRoleAuth
+	if vc.IsWrappingToken == true {
```
maraino, line 217: Fix linter error
```suggestion
	if vc.IsWrappingToken {
```

maraino (changes requested):
It's quite good, but I've noticed a few errors and things that can be improved.

cas/vaultcas/vaultcas.go
```diff
@@ -0,0 +1,322 @@
...
+
+// VaultCAS implements a Certificate Authority Service using Hashicorp Vault.
+type VaultCAS struct {
+	client      *vault.Client
+	config      VaultOptions
+	fingerprint string
+}
+
+func loadOptions(config map[string]interface{}) (vc VaultOptions, err error) {
+	err = mapstructure.Decode(config, &vc)
```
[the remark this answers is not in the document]
ahmet2mir: I also use mapstructure to unmarshal, so I convert to byte then unmarshal, working well :)

cas/vaultcas/vaultcas.go
```diff
@@ -0,0 +1,322 @@
...
+		sans = append(sans, ip.String())
+	}
+	for _, u := range cr.URIs {
+		sans = append(sans, u.String())
+	}
+
+	commonName := cr.Subject.CommonName
+	if commonName == "" && len(sans) > 0 {
+		commonName = sans[0]
+	}
```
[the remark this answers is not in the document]
ahmet2mir: It's a bad copy/pasta from StepCAS ^^

cas/vaultcas/vaultcas.go
ahmet2mir: Serial numbers are "dashed", see https://github.com/hashicorp/vault/blob/56e81b899bdb3ce8f3bc28a36988f6a7cd37826b/builtin/logical/pki/path_revoke.go#L20

cas/vaultcas/vaultcas.go
```diff
@@ -0,0 +1,322 @@
...
+	client, err := vault.NewClient(config)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to initialize vault client")
+	}
+
+	var appRoleAuth *auth.AppRoleAuth
+	if vc.IsWrappingToken == true {
+		appRoleAuth, err = auth.NewAppRoleAuth(
+			vc.RoleID,
+			&auth.SecretID{FromString: vc.SecretID},
```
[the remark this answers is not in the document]
ahmet2mir: I will use the same struct and configure secretid with 3 options

ahmet2mir:
@maraino thanks for the fast review, I did a new commit gathering all suggestions.

cas/vaultcas/vaultcas.go
maraino: This is not yet fixed.

maraino (commented):
Hi @ahmet2mir, except for the lower case `d` in the name of Ed25519 everything looks good.

> (Once ok, I could continue on test part ^^)

Do you mean that you're going to add more unit tests?

ahmet2mir:
> Do you mean that you're going to add more unit tests?

@maraino as I wasn't sure on the "code"design" I just wrote test on create part not the rest (lazy me). So I will add them.

ahmet2mir:
@maraino I did a new commit with more tests (on revoke, authority renew and loadoptions)

cas/vaultcas/vaultcas.go
```diff
@@ -0,0 +1,322 @@
...
+	case csr.PublicKeyAlgorithm == x509.Ed25519:
+		vaultPKIRole = v.config.PKIRoleED25519
+	default:
+		return nil, nil, errors.Errorf("createCertificate: Unsupported public key algorithm '%v'", csr.PublicKeyAlgorithm)
+	}
+
+	certPemBytes := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE REQUEST", Bytes: csr.Raw})
+	if certPemBytes == nil {
+		return nil, nil, errors.Errorf("createCertificate: Failed to encode pem '%v'", csr.Raw)
+	}
```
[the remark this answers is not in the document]
ahmet2mir: missed this one, thanks!

maraino (changes requested):
@ahmet2mir I've been testing this with vault, got it working but not as expected.

In VaultCAS `CreateCertificate` should be able to sign with an intermediate certificate, but `GetCertificateAuthority` should get the root one. As it is right now the signer and what `step-ca` thinks is the root are always the same.

I got it working more or less as expected changing the read path of `GetCertificateAuthority` using the mount path of the root I created. 

This is more or less how I configured this:

* Generate root CA
```sh
vault secrets enable pki
vault secrets tune -max-lease-ttl=87600h pki
vault write pki/root/generate/internal common_name="Vault Root CA" ttl=87600h
```
* Generate intermediate csr
```
vault secrets enable -path=pki_int pki
vault secrets tune -max-lease-ttl=43800h pki_int
vault write pki_int/intermediate/generate/internal common_name="Vault Intermediate CA" ttl=43800h
vault write pki/root/sign-intermediate csr=@intermediate.csr format=pem_bundle ttl=43800h
vault write pki_int/intermediate/set-signed certificate=@intermediate_ca.crt
vault write pki_int/config/urls issuing_certificates="http://127.0.0.1:8200/v1/pki_int/ca" crl_distribution_points="http://127.0.0.1:8200/v1/pki_int/crl"
```

I created an EC role `ec-test-role` on both `pki` and `pki_int` just in case. And I added the following policies:

```hcl
path "pki/sign/ec-test-role" {
    capabilities = ["create", "update"]
}

path "pki_int/sign/ec-test-role" {
    capabilities = ["create", "update"]
}

path "pki/revoke" {
    capabilities = ["create", "update"]
}

path "pki_int/revoke" {
    capabilities = ["create", "update"]
}
```

To get the proper root I hardcoded in `GetCertificateAuthority` my root mount patch like this:
```diff
diff --git a/cas/vaultcas/vaultcas.go b/cas/vaultcas/vaultcas.go
index 234ea82..5b54170 100644
--- a/cas/vaultcas/vaultcas.go
+++ b/cas/vaultcas/vaultcas.go
@@ -254,7 +254,7 @@ func (v *VaultCAS) CreateCertificate(req *apiv1.CreateCertificateRequest) (*apiv
 }

 func (v *VaultCAS) GetCertificateAuthority(req *apiv1.GetCertificateAuthorityRequest) (*apiv1.GetCertificateAuthorityResponse, error) {
-       secret, err := v.client.Logical().Read(v.config.PKI + "/cert/ca")
+       secret, err := v.client.Logical().Read("pki/cert/ca")
        if err != nil {
                return nil, errors.Wrap(err, "unable to read root")
        }
```

And configured the CA with this:
```
"type": "vaultcas",
"certificateAuthority": "http://127.0.0.1:8200",
"certificateAuthorityFingerprint": "c199d121abf78bd6dc7a7d59e9bf2fd4ebaf0d5b0a36b84e7ea9fa5e006cab78",
"config": {
  "pki": "pki_int",
  "pkiRoleEC": "ec-test-role",
  "appRole": "approle",
  "roleID": "REDACTED",
  "secretID": {
    "fromString": "REDACTED"	
  },
  "IsWrappingToken": false
}
```

That fingerprint was the one from the root in `pki`, not the one from the intermediate.

I haven't been able to use a common mount path for the root an intermediate, and looking at [vault docs](https://www.vaultproject.io/docs/secrets/pki#setting-up-intermediate-ca) they also use a different one. So I believe you will need to add two different `pki` mount paths or get the IssuingCA of the intermediate, and in case that you're using multiple intermediates the actual root.

devon-mar:
@maraino 

I think that you can read the full chain with `vault read pki_int/cert/ca_chain` (see [here](https://www.vaultproject.io/api-docs/secret/pki#read-ca-certificate-chain)).
This only works though if you include the root in `intermediate_ca.crt` when you write to `set-signed`.

maraino:
@ahmet2mir See @devon-mar comments, that makes sense, can you try using this here?
```
func (v *VaultCAS) GetCertificateAuthority(req *apiv1.GetCertificateAuthorityRequest) (*apiv1.GetCertificateAuthorityResponse, error) {
	secret, err := v.client.Logical().Read(v.config.PKI + "/cert/ca")
```

I'll try too, I was using vault in dev mode and have to recreate everything :(

maraino:
@ahmet2mir @devon-mar That actually works, but I had to make some modifications:
1. In `GetCertificateAuthority` I need to extract the last cert from the bundle:
```golang
bundle, err := pemutil.ParseCertificateBundle([]byte(certBundle.Certificate))
if err != nil {
	return nil, err
}
cert := bundle[len(bundle)-1]
```
2. In `CreateCertificate` or `createCertificate` I need to exclude the root from the chain. We only want final certificates to be like `leaf+int' or if there are multiple intermediates 'leaf+int1+int2+...'

Simple ways to test this:
1. `step ca roots` should only print the root
```sh
$ step ca roots | step certificate inspect  --bundle --short
X.509v3 Root CA Certificate (RSA 2048) [Serial: 1055...3576]
  Subject:     Vault Root CA
  Issuer:      Vault Root CA
  Valid from:  2022-02-03T19:51:59Z
          to:  2032-02-01T19:52:29Z
```
2. RA certificate should only contain the leaf and intermediate:
```sh
$ step certificate inspect  --bundle --short --insecure https://ra.smallstep.com:9100
X.509v3 TLS Certificate (ECDSA P-256) [Serial: 7250...5052]
  Subject:     Step Online CA
               ra.smallstep.com
  Issuer:      Vault Intermediate CA
  Valid from:  2022-02-03T20:16:11Z
          to:  2022-02-04T20:16:41Z
X.509v3 Intermediate CA Certificate (RSA 2048) [Serial: 2340...8721]
  Subject:     Vault Intermediate CA
  Issuer:      Vault Root CA
  Valid from:  2022-02-03T19:52:46Z
          to:  2027-02-02T19:53:16Z
```
3. And of course, the certificate in `step ca certificate localhost localhost.crt localhost.key` should only contain the leaf and the intermediate.
```sh
$ step ca certificate localhost localhost.crt localhost.key
✔ Provisioner: mariano@smallstep.com (JWK) [kid: L38TOXsig8h6FeBOos03nFy6iXmwusFcIBBB0ZilahY]
✔ Please enter the password to decrypt the provisioner key: █
✔ CA: https://ra.smallstep.com:9100
✔ Certificate: localhost.crt
✔ Private Key: localhost.key
➜  vaultcas step certificate inspect --short --bundle --insecure localhost.crt
X.509v3 TLS Certificate (ECDSA P-256) [Serial: 5595...6613]
  Subject:     localhost
  Issuer:      Vault Intermediate CA
  Valid from:  2022-02-03T20:26:26Z
          to:  2022-02-04T20:26:56Z
X.509v3 Intermediate CA Certificate (RSA 2048) [Serial: 2340...8721]
  Subject:     Vault Intermediate CA
  Issuer:      Vault Root CA
  Valid from:  2022-02-03T19:52:46Z
          to:  2027-02-02T19:53:16Z
```
4. You should be able to revoke, using serial and mTLS:
```sh
$ step ca revoke 177563624599296747080562250802770551645516204551
✔ Provisioner: mariano@smallstep.com (JWK) [kid: L38TOXsig8h6FeBOos03nFy6iXmwusFcIBBB0ZilahY]
✔ Please enter the password to decrypt the provisioner key:
✔ CA: https://ra.smallstep.com:9100
Certificate with Serial Number 177563624599296747080562250802770551645516204551 has been revoked.
$ step ca revoke --cert localhost.crt --key localhost.key
✔ CA: https://ra.smallstep.com:9100
Certificate with Serial Number 559505194742719878141008310471477021510860676613 has been revoked.
```

ahmet2mir:
Hello @mariano,

I take another path, my main issue when playing with table is that we must expect that certificate delivered by vault or any other cas are "consistent".

To avoid, vault sign return a certificate (which could be chained) and a cachain (which could be chained).
I merge both, iterate and identify which is root and leaf, and let other as intermediate.
Then I bubble sort intermediate to have a consistent chain in intermediate.

What do you think ?

When I run provided commands, I have the same output :)

maraino:
@ahmet2mir I've pushed some changes to your branch to solve the conflicts and fix the identification of a root certificate.

> To avoid, vault sign return a certificate (which could be chained) and a cachain (which could be chained).
I merge both, iterate and identify which is root and leaf, and let other as intermediate.
Then I bubble sort intermediate to have a consistent chain in intermediate.
>
> What do you think ?

What's the reason behind your sort, can we just assume that the PKI in Vault must be properly configured?

By the way, `isSignedBy` is not properly implemented, it might work because you don't pass the same certificate twice, but if you do, and you pass the same intermediate to both arguments the function will return true, and that's not the case. See my method `isRoot`.

And instead of implementing the sort yourself, you can use the [sort](https://pkg.go.dev/sort) package.

ahmet2mir:
Hello,

I'm still busy at work I will try to fix in 2 weeks. Keep you posted.

maraino:
This PR is now tracked on https://github.com/smallstep/certificates/pull/894. 

@ahmet2mir, thanks and let me know if you have any issues with my changes.

PR #804, opened 2022-01-19
Normalize IPv6 hostname addresses

hslatman:
This PR ensures that IPv6 addresses the CA is configured for will contain square brackets when used as a hostname. This is at least the case for token audiences.

With these changes and the ones from the CLI in https://github.com/smallstep/cli/pull/610 the CA will have a cert for an IP SAN (instead of DNS) for serving its API, will respond to requests to its IPv6 address (i.e. `https://[::1]:8443`) and allows operations that are authorized by a JWT with that same hostname in it.


dopey (approved):
lgtm

hslatman:
@dopey: I made a couple more changes to ensure IPv6 is handled consistently.

dopey (approved):
lgtm

PR #813, opened 2022-02-01
Added some example ansible configs

J-Hunter-Hawke:
We have had some customers ask and others provide ansible configuration files. These are some examples of this that we can add to our Secret Menu.


tashian (approved):
lgtm!

PR #814, opened 2022-02-02
Authority enforcer option

maraino:
### Description

Add an authority option that we can use to inject X509 enforcers. These enforcers are used to modify a cert after any validation.


dopey (approved):
lgtm

PR #817, opened 2022-02-04
Add descriptive provisioner JWK decryption error messages

### LINKED ISSUE #816: Improve Provisioner Error Messaging Surrounding JWK Decryption

More descriptive error messaging when JWK decryption is attempted but fails.  For example, an errant newline in a base64 encoded password currently results in the following message bubbling up to the logs, which isn't particularly helpful:

`square/go-jose: error in cryptographic primitive`

### Why this is needed

Assists end users in troubleshooting issues like the aforementioned one, rather than having to trace through the codebase to figure out where something it going awry.  I ran into this issue when deploying the `step-issuer` helm chart, and would have appreciated better feedback.

Will be opening a PR to address this, though `golang` is not my forte, so please feel free to modify/decline if my improvement isn't warranted.

Cpcrook:
### Description

Wrap other errors in JWK decryption process with more helpful messaging.  This should help users troubleshoot misconfiguration more easily.

Fixes #816

### DIFF
--- a/ca/provisioner.go
+++ b/ca/provisioner.go
@@ -155,11 +155,11 @@ func (p *Provisioner) SSHToken(certType, keyID string, principals []string) (str
 func decryptProvisionerJWK(encryptedKey string, password []byte) (*jose.JSONWebKey, error) {
 	enc, err := jose.ParseEncrypted(encryptedKey)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


maraino (approved):
Thanks, @Cpcrook looks good.

Issue #735, opened 2021-11-01
Azure Provisioner - Authorize based on Subscription ID


rvichery:
### What would you like to be added
Today, Azure provisioner requests are authorized based on Tenant ID and optionally Resource Groups. I think it would be good to also support filtering/authorizing base on Subscription ID. The Subscription ID is available in the Azure Token under `XMSMirID` path.

### Why this is needed
In my company's environment, Resource Groups are dynamically created/deleted (multiple times a day) so configuring Resource Group filtering at the provisioner level is not an option. Our Tenant ID contains many subscriptions (dev, staging, pre-prod, prod, ...), which will let unwanted Virtual Machines issue certificate against a given intermediate CA (dev VM to Prod Intermediate CA for example).

I would be happy to contribute this feature/enhancement.

maraino:
Good point, as you say we currently only support filtering by resource groups, but not subscription ids, but the logic should be pretty similar. We'll bring this to our next triage meeting for prioritization, if I don't fix it before that :)

maraino:
Hi @rvichery, I'm seeing now that you're willing to contribute this feature, if you do, make sure to add the functionality in all places where ResourceGroup appears:
* authority/provisioner/azure.go
* authority/provisioner/azure_test.go
* authority/provisioners.go

You will also need to change the [linkedca repo](https://github.com/smallstep/linkedca/blob/708f4c5b5a8917a2d31991ea541c19805e26a9fd/provisioners.proto#L113-L119), and we would like to add that functionality to the `step` cli too.

rvichery:
@maraino I can probably get a PR going over the weekend for this. Thank you for pointing me to the https://github.com/smallstep/linkedca/blob/708f4c5b5a8917a2d31991ea541c19805e26a9fd/provisioners.proto#L113-L119. I will also make the changes in `step` cli.

rvichery:
Sorry for the delay on this one, I have been quite busy with other projects. I looked at how to implement this and was faced with a fundamental question on how to handle cases where a list of subscription IDs and a list of resource groups are configured: Should we consider a logical disjunction between the two lists where allowing the token? 
Another option could be to enhance the way resource groups are configured today and have the following format options:
- `subscriptionA/*` to allow all resource group in the subscription A
- `subscriptionA/resourceGroupA` to only allow resource group A in subscription A
- `resourceGroupA` to only allow resource group A from all subscriptions (backward compatible with today's behavior)

Please tell me what you think and I can implement accordingly.

PR #829, opened 2022-02-22
Add subscription and object ID validation options to Azure provisioner

vijayjt:
Add support for validation of certificate requests for the Azure provisioner based on:                                                                                  
                                                                                                                                                                        
* Subscription ID.                                                                                                                                                      
* Object ID of the Azure AD identity (Service Principal, Managed Identity etc) making the request.                                                                      
                                                                                                                                                                        
I used the object ID as this uniquely identifies the object in Azure AD. As there is a 1:many relationship between Azure AD Applications and Service Principals as per the MSFT [docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/app-objects-and-service-principals).                                                                                                                                                                
                                                                                                                                                                        
* If only the subscription ID is provided then this is equivalent to saying any resource group in the subscription can be used to request a certificate                 
* If a subscription ID and a resource group is specified then the request must be from a subscription ID in the list and one of the resource groups in the list         
                                                                                                                                                                        
In other words the logic is:                                                                                                                                            
                                                                                                                                                                        
`(s1 OR s2) AND (r1 OR r2) AND (oid1 OR oid2) `                                                                                                             
                                                                                                                                                                        
This PR covers the requirement from issue #735.                                                                                                                         
                                                                                                                                                                        
I have raised the corresponding PR in the:
* Linked CA  [repo](https://github.com/smallstep/linkedca/pull/10)
* CLI [repo](https://github.com/smallstep/cli/pull/642) 

I am not a developer so apologies in advance if there are glaring mistakes / go style faux pas in the code.


maraino (commented):
The change looks good, but I have a question is the term `aadObjectID` more common than just `ObjectID` in Azure? When I configure an OIDC app in Azure I see ObjectID and not Azure Active Directory Object ID. Same in the docs for [identity tokens](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http)

I've noticed that our existing filter by resource group was only in `AuthorizeSign` and not in `AuthorizeSSHSign`, so for the moment the position of your filters looks right to me. We might move them later to authorizeToken so it works for both sign methods.

vijayjt:
> The change looks good, but I have a question is the term `aadObjectID` more common than just `ObjectID` in Azure? When I configure an OIDC app in Azure I see ObjectID and not Azure Active Directory Object ID. Same in the docs for [identity tokens](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http)
> 
> I've noticed that our existing filter by resource group was only in `AuthorizeSign` and not in `AuthorizeSSHSign`, so for the moment the position of your filters looks right to me. We might move them later to authorizeToken so it works for both sign methods.

Thanks for the feedback. You're right it makes more sense to just use `ObjectID`. To make it consistent I can update the CLI flags such as this [one](https://github.com/smallstep/cli/pull/642/files#diff-2b6361ca85e114302902cdaa83939be1a0a16cb1a44d694f7977dcc6501159f2R248) from `azureAdIdentityObjectIDFlag` to `azureObjectIDFlag`.

maraino:
> Thanks for the feedback. You're right it makes more sense to just use ObjectID. To make it consistent I can update the CLI flags such as this [one](https://github.com/smallstep/cli/pull/642/files#diff-2b6361ca85e114302902cdaa83939be1a0a16cb1a44d694f7977dcc6501159f2R248) from azureAdIdentityObjectIDFlag to azureObjectIDFlag.

Yes please, certificates, cli and linkedca accordingly. I'll review all of them.

maraino (changes requested):
I've merged your linkedca PR and tagged as `v0.10.0`.

There's a couple of this missing in this PR:
1. Update the version of linkedca.
2. The authority/provisioners.go file contains some code to convert from and to linkedca types, you should the new types here:
https://github.com/smallstep/certificates/blob/dedd1364072e8fdcbb2a89e202b70707411ac55e/authority/provisioners.go#L701-L714
https://github.com/smallstep/certificates/blob/dedd1364072e8fdcbb2a89e202b70707411ac55e/authority/provisioners.go#L854-L877

maraino:
@vijayjt I'll take a look at the cli one after merging this. After merging you will need to update the cli dependency to certificate@master

vijayjt:
> I've merged your linkedca PR and tagged as `v0.10.0`.
> 
> There's a couple of this missing in this PR:
> 
> 1. Update the version of linkedca.
> 2. The authority/provisioners.go file contains some code to convert from and to linkedca types, you should the new types here:
>    https://github.com/smallstep/certificates/blob/dedd1364072e8fdcbb2a89e202b70707411ac55e/authority/provisioners.go#L701-L714
>    
>    https://github.com/smallstep/certificates/blob/dedd1364072e8fdcbb2a89e202b70707411ac55e/authority/provisioners.go#L854-L877

Thanks for reviewing - missed those. I will update that shortly. Thanks!

maraino (approved):
@vijayjt Thanks a bunch

maraino:
@vijayjt I've changed the JSON tag for ObjectIDs to `objectIDs` here https://github.com/smallstep/certificates/pull/840

vijayjt:
Great, thank you! Once the new tag is created I will update dependency version in the cli repo for certificates (and linkedca).

PR #831, opened 2022-02-23
bump nosql for postgres support

dopey:


hslatman:
🎉

maraino (approved):
LGTM

I was wondering why pgx depends on pq, and it turns out that github.com/jackc/pgtype has a package used in tests that allows running the same tests on both pgx and pq.

imZack:
This is awesome! Can we have a release for enabling pg?

dopey:
👍 released

PR #836, opened 2022-02-28
Add ACME configuration prerequisites check

hslatman:


acme/api/middleware.go
```diff
@@ -309,6 +308,24 @@ func (h *Handler) lookupProvisioner(next nextHTTP) nextHTTP {
...
+func (h *Handler) checkPrerequisites(next nextHTTP) nextHTTP {
+	return func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		ok, err := h.prerequisitesChecker(ctx)
+		if err != nil {
+			api.WriteError(w, acme.WrapErrorISE(err, "error checking acme provisioner prerequisites"))
+			return
+		}
+		if !ok {
+			api.WriteError(w, acme.NewErrorISE("acme provisioner configuration lacks prerequisites"))
```
dopey, line 322: this should maybe be un Unimplemented error, rather than 500

hslatman: Sounds OK. I had it at first, but changed it later, as I saw that it would report a `RejectedIdentifier` ACME error when doing a 501. I don't think it'll be too big of a problem, so I changed it to 501 now.

hslatman:
Merging it; I need it in Landlord

PR #838, opened 2022-02-28
Validate provisioner configuration before storing in DB

dopey:


authority/provisioners.go
```diff
@@ -133,9 +133,18 @@ func (a *Authority) StoreProvisioner(ctx context.Context, prov *linkedca.Provisi
...
 	}
 
+	provisionerConfig, err := a.generateProvisionerConfig(ctx)
+	if err != nil {
+		return admin.WrapErrorISE(err, "error generating provisioner config")
+	}
+
+	if err := certProv.Init(*provisionerConfig); err != nil {
+		return admin.WrapError(admin.ErrorBadRequestType, err, "error validating configuration for provisioner %s", prov.Name)
+	}
```
maraino, line 143: We probably can change the return type of `generateProvisionerConfig` if we always use it as a value.

maraino (commented):
This is ok, but I don't think it will solve the issue completely. At the start of the CA, we can disable those provisioners that are not in a valid state, e.g. an OIDC server is not responding.

maraino (approved):
lgtm

PR #844, opened 2022-03-02
Enable use of `--context` in `step-ca` command

dopey:


tashian:
Related docs PR: https://github.com/smallstep/docs/pull/58

cmd/step-ca/main.go
```diff
@@ -119,7 +119,8 @@ func main() {
 	app.Usage = "an online certificate authority for secure automated certificate management"
 	app.UsageText = `**step-ca** <config> [**--password-file**=<file>]
 [**--ssh-host-password-file**=<file>] [**--ssh-user-password-file**=<file>]
-[**--issuer-password-file**=<file>] [**--resolver**=<addr>] [**--help**] [**--version**]`
+[**--issuer-password-file**=<file>] [**--resolver**=<addr>] [**--help**]
+[**--version**] [**--context**=<name>]`
```
maraino, line 123: We need an example on running step-ca with context, and just `step-ca` with the default context.

maraino (changes requested):

PR #845, opened 2022-03-07
WIP: Support Azure tokens generated by managed identities

vijayjt:
I  know that its expected that users utilise either the ACME provisioner or I guess the K8sSA provisioner on Kubernetes. However, I think it would still be nice if it were possible to use step cli in container on Kubernetes, specifically on Azure Kubernetes Service (AKS).                                                                 
                                                                                                                                                                        
At the moment it's not possible because the value in the `xms_mirid` claim is expected to match the regular expression:                                                 
                                                                                                                                                                        
```go                                                                                                                                                                   
var azureXMSMirIDRegExp = regexp.MustCompile(`(?i)^/subscriptions/([^/]+)/resourceGroups/([^/]+)/providers/Microsoft.Compute/virtualMachines/([^/]+)$`)                 
```                                                                                                                                                                     
                                                                                                                                                                        
However, on AKS you can use [Azure AD Pod Identity](https://docs.microsoft.com/en-us/azure/aks/use-azure-ad-pod-identity) the purpose of this is to allow you to use a managed identity to access Azure resources from within workloads running on AKS. When you use this the token you obtain from the IMDS has an `xms_mirid` claim that looks like this:                                                                                                                                                             
                                                                                                                                                                        
```                                                                                                                                                                     
var azureXMSMirIDRegExp = regexp.MustCompile(`(?i)^/subscriptions/([^/]+)/resourceGroups/([^/]+)/providers/Microsoft.ManagedIdentity/userAssignedIdentities/([^/]+)$`)  
```                                                                                                                                                                     
                                                                                                                                                                        
Note the `xms_mirid` claim looks like this for VM Scale Sets as well.                                                                                                   
                                                                                                                                                                        
The net result is that today if you try to run                                                                                                                          
                                                                                                                                                                        
```                                                                                                                                                                     
step ca certificate <somedomain> mycert.crt mykey.key --provisioner "Azure"                                                                                             
```                                                                                                                                                                     
                                                                                                                                                                        
You receive the error `token is not supported`.                                                                                                                         
                                                                                                                                                                        
The regular expression is used in `authority/provisioner/azure.go` in [authorizeToken](https://github.com/smallstep/certificates/blob/master/authority/provisioner/azure.go#L265) and in [parseResponse](https://github.com/smallstep/cli/blob/master/token/parse.go#L174) in the step cli. Furthermore, in `parseResponse` the virtual machine name is extracted from this `xms_mirid` claim as it is used as a default SAN in [CreateSignRequest](https://github.com/smallstep/cli/blob/master/utils/cautils/certificate_flow.go#L274). 

This PR changes the regular expression used for the `xms_mirid` claim so it can accept both types of resource identifiers. However, there is a side effect that the virtual machine name is used as the default SAN - in particular when DisableCustomSANs is true, the IP address and the VM name are added as SANs. But when the resource type is Microsoft.ManagedIdentity we won't have a real VM name so I have assumed in this case that the virtual machine name should just be the empty string.                
                                                                                                                                                                        
I think it also means that if custom SANs are disabled then it won't issue certs/authorize the token when the resource type is Managed.Identity.

Please let me know if people think we should support this scenario in a different way (or indeed if the feeling is either the ACME or K8sSA provisioner should be used instead). I've marked it as WIP because I think I'd probably need to add new tests but I wanted to see if the project is open to making a change like this. 

This is the associated [PR](https://github.com/smallstep/cli/pull/646) for the cli repo.

### DIFF
--- a/authority/provisioner/azure.go
+++ b/authority/provisioner/azure.go
@@ -30,7 +30,7 @@ const azureDefaultAudience = "https://management.azure.com/"
 
 // azureXMSMirIDRegExp is the regular expression used to parse the xms_mirid claim.
 // Using case insensitive as resourceGroups appears as resourcegroups.
-var azureXMSMirIDRegExp = regexp.MustCompile(`(?i)^/subscriptions/([^/]+)/resourceGroups/([^/]+)/providers/Microsoft.Compute/virtualMachines/([^/]+)$`)
+var azureXMSMirIDRegExp = regexp.MustCompile(`(?i)^/subscriptions/([^/]+)/resourceGroups/([^/]+)/providers/Microsoft.(Compute/virtualMachines|ManagedIdentity/userAssignedIdentities)/([^/]+)$`)
 
 type azureConfig struct {
 	oidcDiscoveryURL string
@@ -263,11 +263,19 @@ func (p *Azure) authorizeToken(token string) (*azurePayload, string, string, str
 	}
 
 	re := azureXMSMirIDRegExp.FindStringSubmatch(claims.XMSMirID)
-	if len(re) != 4 {
+	if len(re) != 5 {
 		return nil, "", "", "", "", errs.Unauthorized("azure.authorizeToken; error parsing xms_mirid claim - %s", claims.XMSMirID)
 	}
+
+	var subscription, group, name string
 	identityObjectID := claims.ObjectID
-	subscription, group, name := re[1], re[2], re[3]
+
+	if strings.Contains(claims.XMSMirID, "virtualMachines") {
+		subscription, group, name = re[1], re[2], re[4]
+	} else {
+		// This is not a VM resource ID so we don't have the VM name so set that to the empty string
+		subscription, group, name = re[1], re[2], ""
+	}
 	return &claims, name, group, subscription, identityObjectID, nil
 }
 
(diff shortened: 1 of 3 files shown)


authority/provisioner/azure.go
```diff
@@ -263,11 +263,19 @@ func (p *Azure) authorizeToken(token string) (*azurePayload, string, string, str
...
+
+	var subscription, group, name string
 	identityObjectID := claims.ObjectID
-	subscription, group, name := re[1], re[2], re[3]
+
+	if strings.Contains(claims.XMSMirID, "virtualMachines") {
+		subscription, group, name = re[1], re[2], re[4]
+	} else {
+		// This is not a VM resource ID so we don't have the VM name so set that to the empty string
+		subscription, group, name = re[1], re[2], ""
```
maraino, line 277: I don't see the point of setting `name` to empty string here. That `name` will be set and validated as `commonName` only if `disableCustomSANs` is set to true. 

In fact, you mention that with disableCustomSANs enabled, it won't issue certs. I don't know what are the values on userAssignedIdentities, but after looking at examples it looks like its something that you can define. So I think that just removing that if condition and just using 
```
subscription, group, name = re[1], re[2], re[4]
```

Will work "ok" for both cases, when disableCustomSANs are disabled (default) or when they are enabled.

maraino (changes requested):
Hi @vijayjt, 

I never used k8s on Azure, but I added a comment about the empty name, I think it's safe to assign re[4] for it.

maraino (approved):
Thanks @vijayjt, looks good.

Issue #665, opened 2021-08-03
When starting up in a terminal, step-ca should print some useful information


tashian:
When started from the terminal, and when the logger format is not JSON, `step-ca` should always print:

- The root fingerprint & CA endpoints
- Configuration paths in use
- Link to docs (& Discord?)
- Client bootstrapping command?

PR #851, opened 2022-03-14
add --context flag to step-ca command

### LINKED ISSUE #802: Add `--context` option to `step-ca` command.
Would be nice to not have to enter the ca.json path as this can change if when using multiple CAs. 

Something like: 

`step-ca --context foo` where the ca.json file is selected from the default location for that context (if not provided as an arg).

azazeal:
This PR adds the `context` flag to `step-ca`.

Closes #802.

### DIFF
--- a/cmd/step-ca/main.go
+++ b/cmd/step-ca/main.go
@@ -117,7 +117,7 @@ func main() {
 	app.HelpName = "step-ca"
 	app.Version = step.Version()
 	app.Usage = "an online certificate authority for secure automated certificate management"
-	app.UsageText = `**step-ca** <config> [**--password-file**=<file>]
+	app.UsageText = `**step-ca** [config] [**--context**=name] [**--password-file**=<file>]
 [**--ssh-host-password-file**=<file>] [**--ssh-user-password-file**=<file>]
 [**--issuer-password-file**=<file>] [**--resolver**=<addr>] [**--help**] [**--version**]`
 	app.Description = `**step-ca** runs the Step Online Certificate Authority
@@ -133,6 +133,7 @@ This command will run indefinitely on success and return \>0 if any error occurs
 These examples assume that you have already initialized your PKI by running
 'step ca init'. If you have not completed this step please see the 'Getting Started'
 section of the README.
+
 Run the Step CA and prompt for password:
 '''
 $ step-ca $STEPPATH/config/ca.json
@@ -141,6 +142,15 @@ Run the Step CA and read the password from a file - this is useful for
 automating deployment:
 '''
 $ step-ca $STEPPATH/config/ca.json --password-file ./password.txt
+'''
+Run the Step CA for the context named _mybiz_ and prompt for password:
+'''
+$ step-ca --context=mybiz
+'''
+Run the Step CA for the context named _mybiz_ and read the password from a file - this is useful for
+automating deployment:
+'''
+$ step-ca --context=mybiz --password-file ./password.txt
 '''`
 	app.Flags = append(app.Flags, commands.AppCommand.Flags...)
 	app.Flags = append(app.Flags, cli.HelpFlag)
--- a/cmd/step-ca/main.go
+++ b/cmd/step-ca/main.go
@@ -117,7 +117,7 @@ func main() {
 	app.HelpName = "step-ca"
 	app.Version = step.Version()
 	app.Usage = "an online certificate authority for secure automated certificate management"
-	app.UsageText = `**step-ca** [config] [**--context**=name] [**--password-file**=<file>]
+	app.UsageText = `**step-ca** [config] [**--context**=<name>] [**--password-file**=<file>]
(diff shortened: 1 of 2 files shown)


cmd/step-ca/main.go
```diff
@@ -117,7 +117,7 @@ func main() {
 	app.HelpName = "step-ca"
 	app.Version = step.Version()
 	app.Usage = "an online certificate authority for secure automated certificate management"
-	app.UsageText = `**step-ca** <config> [**--password-file**=<file>]
+	app.UsageText = `**step-ca** [config] [**--context**=name] [**--password-file**=<file>]
```
maraino, line 120: Add `<x>` to be consistent:
```suggestion
	app.UsageText = `**step-ca** [config] [**--context**=<name>] [**--password-file**=<file>]
```

commands/app.go
```diff
@@ -153,6 +159,23 @@ To get a linked authority token:
...
+		if ctx.NArg() == 0 {
+			return "", cli.ShowAppHelp(ctx)
+		}
+
+		return ctx.Args().Get(0), nil
+	}
+
+	if ctx.NArg() > 0 {
+		return "", errors.New("either the context flag or a config file may be specified but not both")
+	}
```
maraino, line 174: I'm not sure about this, although we don't have full support for some scenarios yet, I think it makes sense to allow to specify a ca.json too. 

Some files in the ca.json support relative paths, and they are relative to your current context or STEPPATH if contexts are not being used. If we add support for it for "root", "crt", "key", and I'm perhaps others, it will make total sense to have two different ca.json in the context to try different configurations.

maraino (changes requested):
@azazeal I've added some comments, @dopey should look if the context-related methods are properly used.

commands/app.go
```diff
@@ -153,6 +159,23 @@ To get a linked authority token:
 	return nil
 }
 
+func pathToConfigFile(ctx *cli.Context) (string, error) {
+	if !ctx.IsSet("context") {
```
maraino, line 163: This is probably better:
```suggestion
	if ctx.String("context") == "" {
```

commands/app.go
dopey: I like the idea of being able to select a non-default `ca.json` file for the context. I've had situations where I've wanted to do this.

hslatman: Wouldn't it make it more complex for users, though? Like yet another option to make the configuration more flexible? I create a new CA in a new context if I need to quickly test something, so that I don't have to think about all of the paths involved. I do see some utility in being able to use the same root/intermediate and such, though.

Showing current configuration information at startup time becomes more of a necessity if we were to support this.

hslatman (commented):
Combined usage of `--context` and a different `ca.json` makes the CA run with the `ca.json` that is provided explicitly, so it overrides the one that's in the directory pointed to by `--context`. I think this behavior is OK, although it could be a bit surprising to users, as the `--context` does not seem to have any effect whatsoever. 

In order for the `--context` an `ca.json` to have an actual combined effect, we'd need to support relative paths in the `ca.json` and ensure that they'd be made absolute based on the `--context` provided. As mentioned before, I think that may make things more confusing for end users, so that's why I'm OK with the current state of the PR (at least for now).

tashian:
Added "current paths in use" to #665

cmd/step-ca/main.go
```diff
@@ -141,7 +142,21 @@ Run the Step CA and read the password from a file - this is useful for
...
+'''
+Run the Step CA for the context named _mybiz_ and an alternate ca.json file:
+'''
+$ step-ca --context=mybiz other-ca.json
+'''
+Run the Step CA for the context named _mybiz_ and read the password from a file - this is useful for
+automating deployment:
+'''
+$ step-ca --context=mybiz --password-file ./password.txt
+'''
```
maraino, line 158: We should add an example for running just `step-ca`, with context enabled, running just `step-ca` will use the default context.

For example, this will run the ssh context.
```
step context select ssh
step-ca --password-file password.txt
```

maraino (approved):
We should add another example, other than that it looks good. We can add it later if you want to merge this.

cmd/step-ca/main.go
azazeal: Sure, will get right on that.

cmd/step-ca/main.go
```diff
@@ -141,7 +142,26 @@ Run the Step CA and read the password from a file - this is useful for
...
+Run the Step CA for the context that's been selected with step and with a custom password file:
+'''
+$ step context select ssh
+$ step-ca --password-file ./password.txt
+'''
```
maraino, line 163: This should be the first example on the context part, it makes sense to be the default command when contexts are enabled.

You can also remove the passive, just removing `that's been`.

maraino (approved):
lgtm

PR #858, opened 2022-03-18
api/read: initial implementation of the package

azazeal:
This PR introduces the `api/read` package and refactors `api`, `authority/admin` & `ca` accordingly so that they use it. It's the first PR, in a series of PRs not yet written, which will allow for less boilerplate in API as well as help reduce dependency on intermediary `errs`-like packages.


ca/client_test.go
```diff
@@ -17,13 +17,15 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"go.step.sm/crypto/x509util"
```
hslatman, line 20: Minor nit: this should go down.

hslatman (commented):
Looking good. I like what's happening.

ca/client_test.go
azazeal: good catch - will amend shortly.

api/read/read.go
```diff
@@ -0,0 +1,31 @@
...
+	"google.golang.org/protobuf/proto"
+
+	"github.com/smallstep/certificates/errs"
+)
+
+// JSON reads JSON from the request body and stores it in the value
+// pointed by v.
+func JSON(r io.Reader, v interface{}) error {
+	if err := json.NewDecoder(r).Decode(v); err != nil {
+		return errs.BadRequestErr(err, "error decoding json")
```
maraino, line 18: If this is always used with the request body, it's probably better to show
```suggestion
		return errs.BadRequestErr(err, "error reading request body")
```

api/read/read.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+	return nil
+}
```
maraino, line 21: Are you planning to add the validator later?

maraino (approved):
lgtm, but I added a small suggestion.

api/read/read.go
azazeal: Right - so, the plan is this: this is the first PR in a series of PRs which, in t he end, will also include the validator.

For now, I'm trying to implement these changes incrementally so what's happening may be reviewed easier.

azazeal: Final result should also render helpers like `errs.BadRequestErr` obsolete, not quite sure how I'll get there right now - but the rest should be straightforward.

api/read/read.go
azazeal: I, too, think this should be changed but it'd break the test suite of other packages. For now, I'll leave it as it is!

PR #859, opened 2022-03-18
api/log: initial implementation of the package

azazeal:
This PR moves `api.LogError` in its own package (`api/log`) for now. It'll be going away soon enough, but for now it's being moved in its own package.


maraino (approved):
lgtm, I didn't know about that `EnabledResponse`, I guess is something acme related only.

PR #861, opened 2022-03-21
Change go version to 1.17 and 1.18

maraino:
### Description
This PR changes the supported versions of Go to 1.17 and 1.18, using Go 1.18 for releases.


hslatman (approved):
LGTM!

It's a bit surprising that it's currently hard to compare the actual contents of the `x509.CertPool`. Even just checking the `Subjects` is not entirely correct, but if that's what we did before, it's probably good to continue doing so. Let's hope https://github.com/golang/go/issues/46057 lands soon 🙂

PR #863, opened 2022-03-22
Linked RA improvements

maraino:
### Description

This PR allows reporting the provisioner used to sign a certificate to the linkedca service. This is especially useful if an RA is used because the provisioner extension might not be there, or might not match the provisioner used in the RA.

At the same time, linked RA configuration can be returned by the linkedca client, this allows to be able to configure the hosted CA for a new RA using the web interface.

LinkedCA/RA does not support SSH yet, so this change doesn't include any improvement on that side.


authority/authority.go
```diff
@@ -253,13 +253,44 @@ func (a *Authority) init() error {
 		}
 	}
 
+	// Initialize linkedca client if necessary. On a linked RA, the issuer
+	// configuration might come from majordomo.
```
dopey, line 257: kinda weird to mention majordomo (not open source) in the documentation, no?

maraino: Well the linkedca client is called Majordomo.

dopey (approved):
lgtm

PR #866, opened 2022-03-22
Add /roots.pem handler

### LINKED ISSUE #832: Provide roots as PEM instead of (just) JSON
On Discord @nicoulaj asked for an option to download the roots in PEM format instead of JSON, so that it's not necessary to extract the certificate(s) from the JSON using `jq`, for example.

Some options are to add a query parameter like `/roots?format=pem`, use a file type like `/roots.pem` or the ability to provide `Accept: application/x-pem-file` (non-standard, but used in practice) as a header.

areed:
### Description

Add an endpoint that returns all root certificates in PEM format. This avoids the need to parse the JSON response from the current /roots endpoint if you only want the certificate.

Fixes #832

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -1344,6 +1344,53 @@ func Test_caHandler_Roots(t *testing.T) {
 	}
 }
 
+func Test_caHandler_RootsPEM(t *testing.T) {
+	cs := &tls.ConnectionState{
+		PeerCertificates: []*x509.Certificate{parseCertificate(certPEM)},
+	}
+	parsedRoot := parseCertificate(rootPEM)
+	tests := []struct {
+		name       string
+		tls        *tls.ConnectionState
+		cert       *x509.Certificate
+		roots      []*x509.Certificate
+		err        error
+		statusCode int
+		expect     string
+	}{
+		{"one root", cs, parseCertificate(certPEM), []*x509.Certificate{parsedRoot}, nil, http.StatusOK, rootPEM},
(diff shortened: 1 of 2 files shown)


api/api_test.go
```diff
@@ -1344,6 +1344,53 @@ func Test_caHandler_Roots(t *testing.T) {
...
+		cert       *x509.Certificate
+		roots      []*x509.Certificate
+		err        error
+		statusCode int
+		expect     string
+	}{
+		{"one root", cs, parseCertificate(certPEM), []*x509.Certificate{parsedRoot}, nil, http.StatusOK, rootPEM},
+		{"two roots", cs, parseCertificate(certPEM), []*x509.Certificate{parsedRoot, parsedRoot}, nil, http.StatusOK, rootPEM + "\n" + rootPEM},
+		{"no peer certificates", &tls.ConnectionState{}, parseCertificate(certPEM), []*x509.Certificate{parsedRoot}, nil, http.StatusOK, rootPEM},
+		{"fail", cs, nil, nil, fmt.Errorf("an error"), http.StatusForbidden, ""},
```
hslatman, line 1364:
```suggestion
		{"fail", cs, nil, nil, errors.New("an error"), http.StatusForbidden, ""},
```

hslatman: Nothing to format

hslatman (approved):
lgtm!

The `InternalServerErr` makes sense; at least for now. The API is currently being revised and this one will probably be changed too then. The bigger issue is the difference in authentication between open source and product, but that's a different and bigger story.

PR #871, opened 2022-03-24
Attempt to fix TestBootstrapClientServerRotation

maraino:
### Description

This change attempts to fix the test TestBootstrapClientServerRotation.Due to the backdate, the renewal options get too large, causing continuous renewals, and random errors. After experimenting with different options, truncating durations to seconds has shown better results than rounding or just using the plain time.


azazeal (approved):
LGTM if it LGTY

PR #888, opened 2022-04-06
Fix/adminra

maraino:
### Description

This PR adds a new table with the provisioner used to store a certificate and uses that table to load the provisioner instead of looking at the provisioner extensions.

With a linked CA or a linked RA it will use the linkedca client to get the certificate and the provisioner who authorized the certificate.

Depends on: 
- [x] #863 
- [x] https://github.com/smallstep/web/pull/806


authority/provisioners.go
```diff
@@ -46,13 +47,43 @@ func (a *Authority) GetProvisioners(cursor string, limit int) (provisioner.List,
...
 	}
 	return p, nil
 }
 
+func (a *Authority) unsafeLoadProvisionerFromDatabase(crt *x509.Certificate) (provisioner.Interface, error) {
+	// certificateDataGetter is an interface that can be use to retrieve the
+	// provisioner from a db or a linked ca.
+	type certificateDataGetter interface {
+		GetCertificateData(string) (*db.CertificateData, error)
+	}
```
hslatman, line 69: `certificateDataGetter` implies there could be more "data" related to certificates than just the provisioner metadata. Is that something to keep in mind for the future, e.g. would we extend the `CertificateData` to include other metadata if needed?

hslatman, line 69: `certificateDataGetter` implies there could be more "data" related to certificates than just the provisioner metadata. Is that something to keep in mind for the future, e.g. would we extend the `CertificateData` to include other metadata if needed?

authority/authorize.go
```diff
@@ -285,9 +282,15 @@ func (a *Authority) authorizeRenew(cert *x509.Certificate) error {
...
 	}
-	p, ok := a.provisioners.LoadByCertificate(cert)
-	if !ok {
-		return errs.Unauthorized("authority.authorizeRenew: provisioner not found", opts...)
+	p, err := a.LoadProvisionerByCertificate(cert)
+	if err != nil {
+		var ok bool
+		// For backward compatibility this method will also succeed if the
+		// provisioner does not have an extension. LoadByCertificate returns the
+		// noop provisioner if this happens, and it allow certificate renewals.
```
hslatman, line 290:
```suggestion
		// noop provisioner if this happens, and it allows certificate renewals.
```

hslatman, line 290:
```suggestion
		// noop provisioner if this happens, and it allows certificate renewals.
```

authority/authorize.go
```diff
@@ -130,22 +130,19 @@ func (a *Authority) AuthorizeAdminToken(r *http.Request, token string) (*linkedc
 	// According to "rfc7519 JSON Web Token" acceptable skew should be no
 	// more than a few minutes.
 	if err := claims.ValidateWithLeeway(jose.Expected{
-		Issuer: prov.GetName(),
+		Issuer: "step-admin-client/1.0",
```
hslatman, line 133: So, if I understand correctly, any client wishing to integrate with the admin API needs to create a JWT with `step-admin-client/1.0` as the `iss` claim after this change? I'm not sure if it's a hard requirement to restrict it to this, but it makes sense to do so for validation purposes. I'm thinking about use cases in which someone doesn't use our CLI to interact with the admin API, which might not happen all too often, but is not unexpected.

hslatman, line 133: So, if I understand correctly, any client wishing to integrate with the admin API needs to create a JWT with `step-admin-client/1.0` as the `iss` claim after this change? I'm not sure if it's a hard requirement to restrict it to this, but it makes sense to do so for validation purposes. I'm thinking about use cases in which someone doesn't use our CLI to interact with the admin API, which might not happen all too often, but is not unexpected.

authority/provisioners.go
```diff
@@ -46,13 +47,43 @@ func (a *Authority) GetProvisioners(cursor string, limit int) (provisioner.List,
...
 	p, ok := a.provisioners.LoadByCertificate(crt)
-	if !ok {
+	if !ok || p.GetType() == 0 {
 		return nil, admin.NewError(admin.ErrorNotFoundType, "unable to load provisioner from certificate")
 	}
 	return p, nil
 }
 
+func (a *Authority) unsafeLoadProvisionerFromDatabase(crt *x509.Certificate) (provisioner.Interface, error) {
+	// certificateDataGetter is an interface that can be use to retrieve the
```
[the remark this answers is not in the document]
maraino: Duplicated suggestion.

authority/authorize.go
maraino: Duplicated suggestion.

authority/provisioners.go
maraino: I wasn't able to add more metadata to the certificate table because the structure is `key -> cert.Raw` so I created a new table that can contain more metadata, right now it only contains the provisioner used, but we can add more to it.

maraino: Duplicated comment.

authority/authorize.go
maraino: Duplicated comment.

authority/linkedca.go
```diff
@@ -235,6 +235,28 @@ func (c *linkedCaClient) DeleteAdmin(ctx context.Context, id string) error {
 	return errors.Wrap(err, "error deleting admin")
 }
 
+func (c *linkedCaClient) GetCertificateData(serial string) (*db.CertificateData, error) {
```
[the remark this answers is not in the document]
maraino: linkedCAClient is an implementation of the `admin.DB` but it also implements a few methods from the `db.DB` interface, the ones used to store a certificate.

Currently, the only "non-standard" type used by the `db.DB` interface is also in the `db` package, so it made sense to add it there too. What would you suggest to do here?

authority/authorize.go
```diff
@@ -130,22 +130,19 @@ func (a *Authority) AuthorizeAdminToken(r *http.Request, token string) (*linkedc
...
 		return nil, admin.WrapError(admin.ErrorUnauthorizedType, err, "x5c.authorizeToken; invalid x5c claims")
 	}
 
 	// validate audience: path matches the current path
-	if r.URL.Path != claims.Audience[0] {
-		return nil, admin.NewError(admin.ErrorUnauthorizedType,
-			"x5c.authorizeToken; x5c token has invalid audience "+
-				"claim (aud); expected %s, but got %s", r.URL.Path, claims.Audience)
+	if !matchesAudience(claims.Audience, a.config.Audience(r.URL.Path)) {
+		return nil, admin.NewError(admin.ErrorUnauthorizedType, "x5c.authorizeToken; x5c token has invalid audience claim (aud)")
```
dopey, line 141: Feels like the error in the CA logs could have more information - like the expected and got values as before.

authority/authorize.go
```diff
@@ -285,9 +282,15 @@ func (a *Authority) authorizeRenew(cert *x509.Certificate) error {
...
 		return errs.Unauthorized("authority.authorizeRenew: certificate has been revoked", opts...)
 	}
-	p, ok := a.provisioners.LoadByCertificate(cert)
-	if !ok {
-		return errs.Unauthorized("authority.authorizeRenew: provisioner not found", opts...)
+	p, err := a.LoadProvisionerByCertificate(cert)
+	if err != nil {
+		var ok bool
+		// For backward compatibility this method will also succeed if the
+		// provisioner does not have an extension. LoadByCertificate returns the
```
dopey, line 289: if the certificate does not have a provisioner extension.

authority/authorize.go
```diff
@@ -386,8 +389,8 @@ func (a *Authority) AuthorizeRenewToken(ctx context.Context, ott string) (*x509.
 		return nil, errs.InternalServerErr(err, errs.WithMessage("error validating renew token"))
 	}
 
-	p, ok := a.provisioners.LoadByCertificate(leaf)
-	if !ok {
+	p, err := a.LoadProvisionerByCertificate(leaf)
```
dopey, line 392: How will this work for certificates without the provisioner extension?

authority/provisioners.go
```diff
@@ -46,13 +47,43 @@ func (a *Authority) GetProvisioners(cursor string, limit int) (provisioner.List,
...
+	type certificateDataGetter interface {
+		GetCertificateData(string) (*db.CertificateData, error)
+	}
+
+	var err error
+	var data *db.CertificateData
+
+	if cdg, ok := a.adminDB.(certificateDataGetter); ok {
+		data, err = cdg.GetCertificateData(crt.SerialNumber.String())
+	} else if cdg, ok := a.db.(certificateDataGetter); ok {
```
dopey, line 76: this can just be `else if cdg, ok = a.db.(certificateDataGetter); ok {`, right?

db/db.go
```diff
@@ -210,6 +225,46 @@ func (db *DB) StoreCertificate(crt *x509.Certificate) error {
...
+	ID   string `json:"id"`
+	Name string `json:"name"`
+	Type string `json:"type"`
+}
+
+// StoreCertificateChain stores the leaf certificate and the provisioner that
+// authorized the certificate.
+func (db *DB) StoreCertificateChain(p provisioner.Interface, chain ...*x509.Certificate) error {
+	leaf := chain[0]
+	if err := db.StoreCertificate(leaf); err != nil {
```
dopey, line 246: Should we do this in a transaction so we don't get stuck in some intermediate state?

dopey (approved):
I left a few comments, but in general I approve.

authority/authorize.go
maraino: We will need to change the issuer in `step` to use a fixed string similar to what I've done in `AuthorizeAdminToken`.

authority/provisioners.go
maraino: It can, both ways are perfectly ok, and the compiler will create the same code on both cases.

db/db.go
maraino: Yes, fixed with 3694ba30

authority/authorize.go
maraino: With ad5aedfa I fixed the keep backward compatibility issue.

authority/authorize.go
maraino: I feel like it was too verbose. And this won't happen unless you use your own implementations. The list of audiences accepted is the number of configured dnsNames + 1, so it can be too verbose.

dopey (approved):

PR #892, opened 2022-04-07
authority/status: removed the package

azazeal:
This PR removes the `authority/status` package, which seems to be unused.


hslatman (approved):
lgtm!

Don't know what it's intended usage is right now, nor if other systems rely on it. Should be simple to revert, if necessary.

PR #901, opened 2022-04-18
Drop any query string from the admin tokens

maraino:
### Description

This PR makes sure the admin token audience is passed without a query string (or any fragment). The audience is checked against a string like `https://host/path`, so URLs like `https://host/path?limit=100` were failing on master if we replace the version of the certificates in the cli with master.


dopey (approved):
lgtm

PR #910, opened 2022-04-25
Dockerfile for HSM-enabled Step CA

jschlyter:
### Description

- Update existing Dockerfile to include GCP and AWS KMS (does not require CGO)
- Add new Dockerfile to build Step CA with support for PKCS#11 and Yubikey (with CGO and PCSC), not yet included in the build process 
- Start pcscd in entrypoint if installed

### DIFF
--- a/docker/Dockerfile.step-ca
+++ b/docker/Dockerfile.step-ca
@@ -3,15 +3,15 @@ FROM golang:alpine AS builder
 WORKDIR /src
 COPY . .
 
-RUN apk add --no-cache \
-   curl \
-   git \
-   make && \
-   make V=1 bin/step-ca
+RUN apk add --no-cache curl git make
+RUN make V=1 bin/step-ca bin/step-awskms-init bin/step-cloudkms-init
+
 
 FROM smallstep/step-cli:latest
 
 COPY --from=builder /src/bin/step-ca /usr/local/bin/step-ca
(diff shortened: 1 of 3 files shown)


tashian:
Hi Jakob,
Thanks for the PR!
I was able to get both images to build in my local environment.
I'm wondering if there's a way we can achieve this CGO-enabled build without duplicating the Dockerfile?
For example, using a build arg and some sort of conditional.

maraino:
@jschlyter Nice, have you tested pkcs#11 or yubikey with musl?

jschlyter:
> @jschlyter Nice, have you tested pkcs#11 or yubikey with musl?

I've tested it with SoftHSM, but not with Yubikey PIV.

jschlyter:
> I'm wondering if there's a way we can achieve this CGO-enabled build without duplicating the Dockerfile and adding 3-4 lines? It would be nice if we didn't have two Dockerfiles. Maybe we could use a build arg and some sort of conditional.

It's not pretty, but it could be done like this: https://gist.github.com/jschlyter/ed4196eb4b907b0ab3448cdca0a6d939

The problem is that Docker doesn't really have conditionals and every command must succeed.

tashian:
Ok, I'm comfortable with having a second Dockerfile, in that case.

tashian (approved):
lgtm!

PR #913, opened 2022-04-26
Vault Kubernetes Auth

delamart:
Support for the vault kubernetes authentication method when using the Vault CAS by using the token available in a kubernetes pod.
https://www.vaultproject.io/docs/auth/kubernetes


maraino:
Hi @delamart, thanks for sending this, we should definitively add something like this, but after looking at Vault's [Auth.Login](https://pkg.go.dev/github.com/hashicorp/vault/api#Auth.Login) I think we can do something a little more generic and make it work for k8s, gcp, azure, aws, ...

Perhaps something like this:
```
"authMethod": "kubernetes"
"roleName": "myRole"
```
or like this to differentiate custom auth options that we can add in the future:
```
"auth": {
  "method": "kubernetes",
  "role": "myRole",
  "futureOption": "foo"
}
```

I'm not asking you to add support for all of them, but I think we need to be able to change the method without adding `gcpRole`, `azureRole`, ... And if the method is `""` or `"approle"` we should use the method already implemented.

You can add something like this:

```golang
func getAuthMethod(ops VaultOptions) (vault.AuthMethod, error) {
    switch strings.ToLower(opts.Auth.Method) {
    case "", "approle":
       ...
    case "kubernetes":
       ...
    default:
       return nil, fmt.Errorf("unsupported authMethod %q", opts.AuthMethod)
    }
}
```

maraino:
> Vault already implements this in the vault agent auto-auth (https://www.vaultproject.io/docs/agent/autoauth).
I can try and implement something like that based on: https://github.com/hashicorp/vault/blob/main/command/agent.go#L402

Yes, that is what I was suggesting.

> Any idea when you would do another release? I'd like to write an article on how to run this with Vault in Kubernetes.

Not sure at the moment, we're working on a couple of PRs that we would like to release soon, but we can do a minor release at any time. 

By the way, we will release also some docs on how to configure it https://github.com/smallstep/docs/pull/92, take a look and let us know how can we improve them.

maraino:
@delamart Here are our docs from the PR I linked you before
https://smallstep.com/docs/step-ca/registration-authority-ra-mode/#hashicorp-vault-ra-mode

maraino (changes requested):
See my comment https://github.com/smallstep/certificates/pull/913#issuecomment-1110439060 or your suggestion about the agent after.

delamart:
Started working on it, started thinking about refactoring and adding many options (like all the vault connection options like mTLS) but will keep it simple. I believe connection options are already available via the VAULT_ env vars btw.
Will try to have a commit by next week.

maraino:
> Started working on it, started thinking about refactoring and adding many options (like all the vault connection options like mTLS) but will keep it simple. I believe connection options are already available via the VAULT_ env vars btw.
> Will try to have a commit by next week.

Transport options like ca-certificates or client certificates are available using the same environment variables as vault cli. See 
https://github.com/hashicorp/vault/blob/b6826e88300922acd770e32130a72d88aeefa495/api/client.go#L327
and 
https://github.com/hashicorp/vault/blob/b6826e88300922acd770e32130a72d88aeefa495/api/client.go#L36-L53

maraino:
@delamart Thanks, once this is ready, can you re-request my review.

delamart:
@maraino ready for review :) there are changes to the VaultOptions that require to update the documentation, let me know if they are ok and if you want me to submit à PR on the documentation.

delamart:
rewrote the if-else chain to switch as required by linter

delamart:
@maraino have you had time to have a look ?

maraino:
> @maraino have you had time to have a look ?

Not yet, perhaps tomorrow.

cas/vaultcas/vaultcas.go
```diff
@@ -77,28 +77,22 @@ func New(ctx context.Context, opts apiv1.Options) (*VaultCAS, error) {
...
 	if err != nil {
-		return nil, fmt.Errorf("unable to initialize AppRole auth method: %w", err)
+		return nil, fmt.Errorf("unable to configure auth method: %w", err)
 	}
 
-	authInfo, err := client.Auth().Login(ctx, appRoleAuth)
+	authInfo, err := client.Auth().Login(ctx, method)
 	if err != nil {
-		return nil, fmt.Errorf("unable to login to AppRole auth method: %w", err)
+		return nil, fmt.Errorf("unable to login to Kubernetes auth method: %w", err)
```
maraino, line 95:
```suggestion
		return nil, fmt.Errorf("unable to login to %s auth method: %w", vc.AuthType, err)
```

cas/vaultcas/vaultcas.go
```diff
@@ -77,28 +77,22 @@ func New(ctx context.Context, opts apiv1.Options) (*VaultCAS, error) {
...
-			auth.WithMountPath(vc.AppRole),
-		)
+	var method vault.AuthMethod
+	switch vc.AuthType {
+	case "kubernetes":
+		method, err = kubernetes.NewKubernetesAuthMethod(vc.AuthMountPath, vc.AuthOptions)
+	case "approle":
+		method, err = approle.NewApproleAuthMethod(vc.AuthMountPath, vc.AuthOptions)
+	default:
+		return nil, fmt.Errorf("unknown auth type: %v", vc.AuthType)
```
maraino, line 87: Can we add a different message when AuthType is not set?

cas/vaultcas/vaultcas.go
```diff
@@ -29,15 +30,14 @@ func init() {
...
-	AppRole         string        `json:"appRole,omitempty"`
-	IsWrappingToken bool          `json:"isWrappingToken,omitempty"`
+	PKIMountPath   string          `json:"pkiMountPath,omitempty"`
+	PKIRoleDefault string          `json:"pkiRoleDefault,omitempty"`
+	PKIRoleRSA     string          `json:"pkiRoleRSA,omitempty"`
+	PKIRoleEC      string          `json:"pkiRoleEC,omitempty"`
+	PKIRoleEd25519 string          `json:"pkiRoleEd25519,omitempty"`
+	AuthType       string          `json:"authType,omitempty"`
+	AuthMountPath  string          `json:"authMountPath,omitempty"`
+	AuthOptions    json.RawMessage `json:"authOptions,omitempty"`
```
maraino, line 40: @tashian The way to configure Vault has changed a little bit, it looks better this way, and it will be more flexible if we want to add more authentication. 

What do you think, is it ok to make this change, or should still attempt to support the old method?

maraino (changes requested):
Hi @delamart, I've added a couple of small changes, and asked @tashian about the configuration changes.

cas/vaultcas/vaultcas.go
tashian: @maraino I think because the feature is so new, we should just change it.

maraino: 👍

cas/vaultcas/vaultcas.go
delamart: I updated the error message, check latest commit, hope that was what you had in mind. Otherwise let me know.

maraino:
@ahmet2mir, just letting you know that we're going to update the way you need to configure vault, we think this will provide us more flexibility for supporting new authentication methods in the future. Docs will be udated too (https://github.com/smallstep/docs/pull/100)

maraino (approved):
Thanks again @delamart

PR #914, opened 2022-04-26
Retrieve authority from the context

maraino:
### Description

This PR changes the HTTP handlers to retrieve the authority or any necessary object from the context, instead of using handlers on methods on a type.

The following packages contains HTTP handlers:

- [x] `api`
- [x] `acme`
- [x] `scep`
- [x] `authority/admin/api`


maraino:
@azazeal If you want you can start looking at this, there will be changes because I will need to merge/rebase @hslatman and @dopey changes.

maraino:
@azazeal this is ready for an initial review, we might need to make some other changes if other PRs are merged. 

This PR only works on the HTTP handlers, we can convert those to private methods, but they won't appear in docs, so I kind of prefer to have them public. 

So bad that `db.AuthDB` as well as `Renew` and `Rekey` do not use context, if we want to change them it will be a breaking change, but I don't think that would be very problematic.

AFAIK the only package that we shouldn't break should be the `acme` one, I think we're ok breaking `api`, `admin` and `scep`.

azazeal (approved):
LGTM

PR #915, opened 2022-04-30
exposing authority configuration for provisioner cli commands

dopey:
Need to expose some more inner workings of the Authority type for access from the cli.

Accompanying cli PR: https://github.com/smallstep/cli/pull/673.

### DIFF
--- a/authority/admin/db.go
+++ b/authority/admin/db.go
@@ -71,52 +71,6 @@ type DB interface {
 	DeleteAdmin(ctx context.Context, id string) error
 }
 
-type NoDB struct{}
-
-func NewNoDB() *NoDB {
-	return &NoDB{}
-}
-
-func (n *NoDB) CreateProvisioner(ctx context.Context, prov *linkedca.Provisioner) error {
-	return nil
-}
-
-func (n *NoDB) GetProvisioner(ctx context.Context, id string) (*linkedca.Provisioner, error) {
-	return nil, nil
-}
-
-func (n *NoDB) GetProvisioners(ctx context.Context) ([]*linkedca.Provisioner, error) {
-	return nil, nil
-}
-
-func (n *NoDB) UpdateProvisioner(ctx context.Context, prov *linkedca.Provisioner) error {
-	return nil
-}
-
-func (n *NoDB) DeleteProvisioner(ctx context.Context, id string) error {
-	return nil
-}
-
-func (n *NoDB) CreateAdmin(ctx context.Context, admin *linkedca.Admin) error {
-	return nil
-}
-
-func (n *NoDB) GetAdmin(ctx context.Context, id string) (*linkedca.Admin, error) {
-	return nil, nil
-}
-
-func (n *NoDB) GetAdmins(ctx context.Context) ([]*linkedca.Admin, error) {
-	return nil, nil
-}
-
-func (n *NoDB) UpdateAdmin(ctx context.Context, prov *linkedca.Admin) error {
-	return nil
-}
-
-func (n *NoDB) DeleteAdmin(ctx context.Context, id string) error {
-	return nil
-}
-
 // MockDB is an implementation of the DB interface that should only be used as
 // a mock in tests.
 type MockDB struct {
--- a/authority/admin/db.go
+++ b/authority/admin/db.go
@@ -71,6 +71,52 @@ type DB interface {
 	DeleteAdmin(ctx context.Context, id string) error
 }
 
+type NoDB struct{}
+
+func NewNoDB() *NoDB {
+	return &NoDB{}
+}
+
+func (n *NoDB) CreateProvisioner(ctx context.Context, prov *linkedca.Provisioner) error {
+	return nil
+}
+
+func (n *NoDB) GetProvisioner(ctx context.Context, id string) (*linkedca.Provisioner, error) {
+	return nil, nil
+}
+
+func (n *NoDB) GetProvisioners(ctx context.Context) ([]*linkedca.Provisioner, error) {
+	return nil, nil
+}
+
+func (n *NoDB) UpdateProvisioner(ctx context.Context, prov *linkedca.Provisioner) error {
+	return nil
+}
+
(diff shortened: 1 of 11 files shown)


authority/authority.go
```diff
@@ -115,6 +115,20 @@ func New(cfg *config.Config, opts ...Option) (*Authority, error) {
...
+	var a = new(Authority)
+
+	// Apply options.
+	for _, fn := range opts {
+		if err := fn(a); err != nil {
+			return nil, err
+		}
+	}
+	return a, nil
+}
```
maraino, line 130: I don't think we should expose this method, you can just copy this code in the cli and use an internal with this code.

authority/admin/db.go
```diff
@@ -71,6 +71,52 @@ type DB interface {
 	DeleteAdmin(ctx context.Context, id string) error
 }
 
+type NoDB struct{}
```
maraino, line 74: You're using this only on the cli, why don't just have this there as a private or internal type.

ca/adminClient.go
```diff
@@ -363,19 +363,19 @@ retry:
...
-	o := new(provisionerOptions)
-	if err := o.apply(opts); err != nil {
+	o := new(ProvisionerOptions)
+	if err := o.Apply(opts); err != nil {
 		return nil, err
 	}
 	var u *url.URL
 	switch {
-	case len(o.id) > 0:
+	case len(o.ID) > 0:
```
maraino, line 372: Linters sometimes suggest to use this:
```suggestion
	case o.ID != "":
```

ca/adminClient.go
```diff
@@ -363,19 +363,19 @@ retry:
...
 		u = c.endpoint.ResolveReference(&url.URL{
 			Path:     "/admin/provisioners/id",
 			RawQuery: o.rawQuery(),
 		})
-	case len(o.name) > 0:
-		u = c.endpoint.ResolveReference(&url.URL{Path: path.Join(adminURLPrefix, "provisioners", o.name)})
+	case len(o.Name) > 0:
```
maraino, line 377:
```suggestion
	case o.Name != "":
```

ca/client.go
```diff
@@ -425,16 +425,17 @@ func parseEndpoint(endpoint string) (*url.URL, error) {
...
-type ProvisionerOption func(o *provisionerOptions) error
+type ProvisionerOption func(o *ProvisionerOptions) error
 
-type provisionerOptions struct {
-	cursor string
-	limit  int
-	id     string
-	name   string
+// ProvisionerOptions stores options for the provisioner CRUD API.
+type ProvisionerOptions struct {
```
maraino, line 431: If we're going to use the `With*` methods from bellow, there's no need to change the visibility of this. Exporting new things is most costly if we want to change the implementation because we need to take into account backward compatibility.

maraino (changes requested):
Mostly ok, but I don't think there's need to expose some things here.

ca/client.go
dopey: I use the ca.ProvisionerOptions in the following places:
https://github.com/smallstep/cli/pull/673/files#diff-88a143d78812b71589b94b33a441ea5b91e1e52a2ab19d6c2306112e8fc9ef26R142
https://github.com/smallstep/cli/pull/673/files#diff-88a143d78812b71589b94b33a441ea5b91e1e52a2ab19d6c2306112e8fc9ef26R177
https://github.com/smallstep/cli/pull/673/files#diff-88a143d78812b71589b94b33a441ea5b91e1e52a2ab19d6c2306112e8fc9ef26R98

maraino: Well ok, it can be done like that or creating its own options in the command. One not though, ID should have more precedence than Name.

authority/admin/db.go
maraino: I think we want to remove this. Or as I suggested on the cli, we might prefer an admin.DB implementation that works directly on the ca.json

ca/adminClient.go
```diff
@@ -472,19 +472,19 @@ func (c *AdminClient) RemoveProvisioner(opts ...ProvisionerOption) error {
...
-	o := new(provisionerOptions)
-	if err := o.apply(opts); err != nil {
+	o := new(ProvisionerOptions)
+	if err := o.Apply(opts); err != nil {
 		return err
 	}
 
 	switch {
-	case len(o.id) > 0:
+	case len(o.ID) > 0:
```
maraino, line 481:
```suggestion
	case len(o.ID) != "":
```

ca/adminClient.go
```diff
@@ -472,19 +472,19 @@ func (c *AdminClient) RemoveProvisioner(opts ...ProvisionerOption) error {
...
 		u = c.endpoint.ResolveReference(&url.URL{
 			Path:     path.Join(adminURLPrefix, "provisioners/id"),
 			RawQuery: o.rawQuery(),
 		})
-	case len(o.name) > 0:
-		u = c.endpoint.ResolveReference(&url.URL{Path: path.Join(adminURLPrefix, "provisioners", o.name)})
+	case len(o.Name) > 0:
```
maraino, line 486:
```suggestion
	case len(o.Name) != "":
```

maraino (commented):
Looks ok, but there're some pending suggestions.

authority/options.go
```diff
@@ -266,6 +266,14 @@ func WithAdminDB(d admin.DB) Option {
 	}
 }
 
+// WithProvisioners is an option to set the provisioner collection.
+func WithProvisioners(ps *provisioner.Collection) Option {
```
maraino, line 270: Can we mark this function as deprecated and add a nolint in the cli?

```suggestion
// WithProvisioners is an option to set the provisioner collection.
// 
// Deprecated: provisioner collections will likely change
func WithProvisioners(ps *provisioner.Collection) Option {
```

maraino (approved):
Approved, but I would mark the WithProvisioners as deprecated or Experimental.

PR #920, opened 2022-05-03
add HSM container build

jschlyter:
#### Name of feature:

Add HSM container build to normal build process.

Building your own HSM container from source and keeping it up to date.

Less pain for users == more joy for developers

No, but I'd be happy to add if I know where to put it

Environments needing support for HSM via PKCS#11, Yubikey or AWS/GCP KMS.

Dunno.

None

### DIFF
--- a/make/docker.mk
+++ b/make/docker.mk
@@ -14,7 +14,6 @@ endif
 
 DOCKER_PLATFORMS = linux/amd64,linux/386,linux/arm,linux/arm64
 DOCKER_IMAGE_NAME = smallstep/step-ca
-DOCKER_IMAGE_NAME_HSM = smallstep/step-ca-hsm
 
 docker-prepare:
 	# Ensure, we can build for ARM architecture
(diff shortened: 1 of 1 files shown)


jschlyter:
I should note that I have no way of testing this, but I hope for the best.

maraino:
@jschlyter I'm not sure if we want to add a new binary to the same image or just another image, a possible way to test this would be to add softhsm2 to a test image or use something like https://github.com/SUNET/pkcs11-proxy to test it against another image.

jschlyter:
@maraino I'm thinking one can either ship another container or ship a single container with HSM support. I understand that the a single container with HSM support is not wanted, so I suggest a separate one (now changed so it is a specific tag, not new image basename).

I agree anyone using this with PKCS#11 will most likely need to build a new container based on this one. Using Softhsm2 for testing might be a good idea, but I'm not sure we want it in the general container. Adding [pkcs11-proxy](https://github.com/SUNET/pkcs11-proxy) is a better choice, but I'm not sure it is maintained anymore - perhaps @leifj knows more. There are no packages for alpine though, so that needs to be build as well.

tashian:
This makes me wonder if we should revisit, for Docker in particular, the idea of building a cgo-enabled image as the default?

maraino:
> This makes me wonder if we should revisit, for Docker in particular, the idea of building a cgo-enabled image as the default?

@tashian that's an option, we can talk about this in this week's meeting. Options are:
1. One docker image with non-cgo and cgo binaries.
2. Two docker images, one with non-cgo (current) binary and the second with cgo.
3. One docker image using only cgo.

In my opinion, separate images would be better because for pkcs#11 you need more than just the CGO binary, you need also the modules you want to support. In our HSM image, we can support one or more, and users can easily inherit our image to support others.

Even if we decide with going to 3, I think a second image with supported modules would be better, you don't need those if you're using Google/AWS/Azure's KMS.

tashian:
> > This makes me wonder if we should revisit, for Docker in particular, the idea of building a cgo-enabled image as the default?
> 
> @tashian that's an option, we can talk about this in this week's meeting. Options are:
> 
>     1. One docker image with non-cgo and cgo binaries.
> 
>     2. Two docker images, one with non-cgo (current) binary and the second with cgo.
> 
>     3. One docker image using only cgo.
> 
> 
> In my opinion, separate images would be better because for pkcs#11 you need more than just the CGO binary, you need also the modules you want to support. In our HSM image, we can support one or more, and users can easily inherit our image to support others.

Oh man I forgot about the PKCS#11 vendor modules. That's so annoying. And, I'm not sure if we'd have permission to redistribute those modules. Plus, if the module accesses HSM / PKCS11 hardware directly, that may not be possible or easy in a containerized environment.

jschlyter:
> Oh man I forgot about the PKCS#11 vendor modules. That's so annoying. And, I'm not sure if we'd have permission to redistribute those modules. Plus, if the module accesses HSM / PKCS11 hardware directly, that may not be possible or easy in a containerized environment.

Most PKCS#11 providers to talk over the network or via a mountable device, so that should not be a problem for containers. The generic user could either:

1. Use the HSM container as is with AWS/GCP KMS or Yubikey (via PCSC), or with the PKCS#11 provider mounted as a volume on startup
2. Build a new container based on the HSM container, adding the PKCS#11 provider of their choice

maraino:
Hi @jschlyter, I've talked with the team, and I think that instead of building an HSM specific image like `Dockerfile.step-ca.hsm`, we need to create a CGO specific one that we release, so users can inherit to add any necessary dependencies, there are some details on this issue I've just created https://github.com/smallstep/certificates/issues/921.

We will probably use something like your PR to create the binaries, but we should do other changes, I'm gonna close this issue for now, but we want to address this properly.

jschlyter:
Thank you @maraino - feel free to ask me for review when you have a new PR.

PR #929, opened 2022-05-10
optimized render.JSON

azazeal:
This PR optimizes `render.JSON` so that it doesn't use an intermediate buffer.

### DIFF
--- a/api/render/render.go
+++ b/api/render/render.go
@@ -2,7 +2,6 @@
 package render
 
 import (
-	"bytes"
 	"encoding/json"
 	"errors"
 	"net/http"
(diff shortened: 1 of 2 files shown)


azazeal:
I don't know whether to go ahead with this or not @maraino & @hslatman. Existing behavior would panic in case there' something wrong with what we're trying to marshal (which is nice for helping debug).

On the other hand, tests are there to take care of this too.

I'm asking for review to see what the team feels like but I'm happy to either close or merge so let me know your thoughts.

api/render/render.go
```diff
@@ -24,14 +23,9 @@ func JSON(w http.ResponseWriter, v interface{}) {
...
 func JSONStatus(w http.ResponseWriter, v interface{}, status int) {
-	var b bytes.Buffer
-	if err := json.NewEncoder(&b).Encode(v); err != nil {
-		panic(err)
-	}
-
 	setContentTypeUnlessPresent(w, "application/json")
 	w.WriteHeader(status)
-	_, _ = b.WriteTo(w)
+	_ = json.NewEncoder(w).Encode(v)
```
hslatman, line 28: You could panic or log here if an error was returned?

azazeal: I've pushed a change that maintains the old behavior (i.e. panics) in case of a `json.UnsupportedValueError` and `json.UnsupportedTypeError` errors.

I think this last version is the one we should go with.

azazeal: The reason I want to keep the panic is that even if something's written in the log the response will still arrive as 200 to the caller.

If either of these error types bubble up, I think, it's an indicator that something unintended has happened in our code (like a channel being marshaled) so I'd like these cases to remain as panics.

hslatman (approved):

api/render/render.go
```diff
@@ -24,14 +23,20 @@ func JSON(w http.ResponseWriter, v interface{}) {
...
+	if err := json.NewEncoder(w).Encode(v); err != nil {
+		var errUnsupportedType *json.UnsupportedTypeError
+		if errors.As(err, &errUnsupportedType) {
+			panic(err)
+		}
+
+		var errUnsupportedValue *json.UnsupportedValueError
+		if errors.As(err, &errUnsupportedValue) {
+			panic(err)
+		}
```
maraino, line 38: A different type of error is not managed.

maraino (changes requested):

azazeal:
The only errors that could panic earlier are addressed, no? Old code panics
only in case of a marshaling error.

api/render/render.go
maraino: json.MarshalerError is missing https://go.dev/play/p/bxYjIBeRTvg

maraino (approved):
lgtm

PR #935, opened 2022-05-14
feat: add health checks to ensure DB availability

fritterhoff:
#### Name of feature: 

Extend health to ensure database availability

Lost connections to the database are only recognized on user/client requests and not before. Using the health endpoint and the liveness checks provided by docker/k8s this detected earlier.

Ensure availability of service. 

Enable health checks in docker containers/configure liveness probes in k8s

All. Especially k8s and docker environments.

None

https://github.com/smallstep/nosql/pull/23

(Side note: At the moment I replace the nosql repo ... that should be fixed as soon as the PR is merged 😉)

💔Thank you!

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -50,6 +50,7 @@ type Authority interface {
 	GetRoots() ([]*x509.Certificate, error)
 	GetFederation() ([]*x509.Certificate, error)
 	Version() authority.Version
+	Health() error
 }
 
 // TimeDuration is an alias of provisioner.TimeDuration
@@ -293,7 +294,12 @@ func (h *caHandler) Version(w http.ResponseWriter, r *http.Request) {
 
 // Health is an HTTP handler that returns the status of the server.
 func (h *caHandler) Health(w http.ResponseWriter, r *http.Request) {
-	render.JSON(w, HealthResponse{Status: "ok"})
+	err := h.Authority.Health()
+	if err == nil {
+		render.JSON(w, HealthResponse{Status: "ok"})
+	} else {
+		render.JSONStatus(w, HealthResponse{Status: "error"}, http.StatusServiceUnavailable)
+	}
 }
 
 // Root is an HTTP handler that using the SHA256 from the URL, returns the root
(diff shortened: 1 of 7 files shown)


dopey:
Hey @fritterhoff 👋 , thanks for opening the PR!

Philosophically, our team agrees with @delamart 's comment here: https://github.com/smallstep/nosql/pull/23#issuecomment-1129297117. Set up a different health endpoint for the database health check and monitor that endpoint individually.

However, we're curious to know more about your intended use case. Specifically, what action are you planning to take if the authority health check fails?

fritterhoff:
Ok, than I'll close both PR's. 
Regarding the action. We actually already switched to the readiness endpoint provided by k8s so incoming requests are rejected on k8s layer and the monitoring detects this error.

It is certainly a question of philosophy what is monitored or equipped with explicit health endpoints. In our case, however, it makes more sense to let the application itself monitor availability.

hslatman:
Noting here for the future: I think having a way to check the health of dependencies like the DB, the signing keys, but potentially also system resources like RAM, disk space, etc still makes sense, even though philosophically it should be monitored directly. 

In this case, the emphasis seems to be on K8s, but not everyone runs `step-ca` on K8s or a similar operating environment. I think there's value in providing a simpler method to also check the dependencies. 

A health check performed from the app itself is more contextual: an external monitoring system may have a different view of the attached resources than the application itself. Say, the access control to a DB changes and the app gets the new configuration, but the monitoring does not (or the other way around). This would result in one of the systems reporting the system is healthy and the other says it's not. Maybe a bit of a contrived example, but that's the idea. 

The current health endpoint is more like a liveness endpoint. All it does now is saying that the port is open, can be reached and that the application can write an HTTP response. It doesn't mean it can serve actual API traffic nor sign certs. I think it makes sense to have different endpoints (or maybe query params, but at least within the same app) through which the states of the application environment and backing services is communicated, so that there's a choice for the user regarding what a healthy CA means to them. 

Implementing a system in which the states of the components are surfaced would require some more thinking and engineering. Also we would have to think if a call to the health endpoint requires the check to be performed synchronously, or if we would kick off an internal lightweight monitoring routine that performs the checks asynchronously and then serves the last result.

PR #941, opened 2022-05-19
Report SSH provisioner

maraino:
### Description

This PR allows reporting the provisioner used to sign an SSH certificate to linkedca, on a previous PR, the reporting of provisioners for X509 certificates was implemented.

As a pending task, we can add a new table in the nosql to store this.

### DIFF
--- a/authority/linkedca.go
+++ b/authority/linkedca.go
@@ -289,7 +289,7 @@ func (c *linkedCaClient) StoreRenewedCertificate(parent *x509.Certificate, fullc
 		PemCertificateChain:  serializeCertificateChain(fullchain[1:]...),
 		PemParentCertificate: serializeCertificateChain(parent),
 	})
-	return errors.Wrap(err, "error posting certificate")
+	return errors.Wrap(err, "error posting renewed certificate")
 }
 
 func (c *linkedCaClient) StoreSSHCertificate(prov provisioner.Interface, crt *ssh.Certificate) error {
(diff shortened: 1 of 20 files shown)


authority/linkedca.go
```diff
@@ -292,11 +292,22 @@ func (c *linkedCaClient) StoreRenewedCertificate(parent *x509.Certificate, fullc
...
+}
+
+func (c *linkedCaClient) StoreRenewedSSHCertificate(parent, crt *ssh.Certificate) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+	_, err := c.client.PostSSHCertificate(ctx, &linkedca.SSHCertificateRequest{
+		Certificate:       string(ssh.MarshalAuthorizedKey(crt)),
+		ParentCertificate: string(ssh.MarshalAuthorizedKey(parent)),
 	})
 	return errors.Wrap(err, "error posting ssh certificate")
```
hslatman, line 312: Maybe this one should state `error posting renewed ssh certificate` to be able to more easily distinguish from the other case?

authority/linkedca.go
```diff
@@ -292,11 +292,22 @@ func (c *linkedCaClient) StoreRenewedCertificate(parent *x509.Certificate, fullc
...
 	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
 	defer cancel()
 	_, err := c.client.PostSSHCertificate(ctx, &linkedca.SSHCertificateRequest{
 		Certificate: string(ssh.MarshalAuthorizedKey(crt)),
+		Provisioner: createProvisionerIdentity(prov),
+	})
+	return errors.Wrap(err, "error posting ssh certificate")
+}
+
+func (c *linkedCaClient) StoreRenewedSSHCertificate(parent, crt *ssh.Certificate) error {
```
hslatman, line 305: Does it make sense to add the `provisioner.Interface` here too? Since the same `linkedca.SSHCertificateRequest` is used to post, I think the provisioner might be a useful piece of data to include? Or is the absence of that property meaningful somewhere different?

maraino: That's something that I asked in #eng, right now we don't have access to it, and there are a few options of what we want to store, do we want to show SSHPOP? do we want to show the original provisioner?

Currently, I'm matching the X509 behavior, but as the renewal is different I might address this in a different PR.

authority/linkedca.go
maraino: Fixed with dd985ce1

hslatman (approved):

PR #944, opened 2022-05-25
Set nil dial context for js/wasm runtime

hslatman:
#### Name of feature:

Set nil dial context for js/wasm runtime

API requests performed by a `js/wasm` program currently fail with a "Protocol not supported" error. With this change HTTP(S) requests performed by a `js/wasm` build will be handled by the browser's Fetch API, instead.

For `js/wasm` targets.

### DIFF
--- a/ca/tls.go
+++ b/ca/tls.go
@@ -12,6 +12,7 @@ import (
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


maraino (approved):
lgtm

PR #950, opened 2022-06-09
step-pkcs11-init pin-file support

gdbelvin:
#### Name of feature:
pin-file support for step-pkcs11-init.

This allows the pin to be provided with bash redirection `<(curl path to secret)`

Pin passwords are private.

💔Thank you!


gdbelvin:
@maraino FYI

PR #951, opened 2022-06-10
Support for key type in step-pkcs11-init

gdbelvin:
This PR enables support for RSA keys which are required for scep

The flags

All environments supported by the key API

#583 

💔Thank you!

### DIFF
--- a/cmd/step-pkcs11-init/main.go
+++ b/cmd/step-pkcs11-init/main.go
@@ -37,11 +37,13 @@ type Config struct {
 	RootKeyObject    string
 	RootSubject      string
 	RootPath         string
+	RootKty          string
 	CrtObject        string
 	CrtPath          string
 	CrtKeyObject     string
 	CrtSubject       string
 	CrtKeyPath       string
+	CrtKty           string
 	SSHHostKeyObject string
 	SSHUserKeyObject string
 	RootFile         string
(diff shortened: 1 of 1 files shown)


maraino (changes requested):
Can you add at least --size for the number of bits in the RSA key, and optionally `--crv` to pick the EC curve to use?

gdbelvin:
sure thing

gdbelvin:
crv doesn't look like it's part of the API?
https://github.com/smallstep/certificates/blob/8d7c3d2f085bc4474ac022f102cd3de423aa6004/kms/softkms/softkms.go#L113

cmd/step-pkcs11-init/main.go
```diff
@@ -297,6 +315,12 @@ func checkObject(k kms.KeyManager, rawuri string) {
 	}
 }
 
+var ktyMapping = map[string]apiv1.SignatureAlgorithm{
+	"EC":  apiv1.ECDSAWithSHA256,
+	"RSA": apiv1.SHA256WithRSAPSS,
```
maraino, line 320: Some PKCS#11 modules like YubiHSM2 can sign using `RSA-PSS`, but the same key can sign using `RSA-PKCS#1 v1.5` and it will do that by default. For example, running this code as it is will sign both certificates using PKCS#1 instead of PSS. If you want to use PSS you need to add to the templates `SignatureAlgorithm: x509.SHA256WithRSAPSS`. 

Then as YubiHSM2 can sign with both, if you want to force RSA-PSS you will need to set templates on each provisioner and manually specify the signature algorithm too.

The behavior is different for other KMSs, for example, in Google CloudKMS, if you create a key for SHA256+RSA-PSS signing, it can only sign with it, `step-ca` is able to detect this and sets the appropriate SignatureAlgorithm.

It might make sense for other KMSs to inherit the signature algorithm from the certificate if an RSA key is used, but right now it uses Go's default, in this case, RSA PKCS#1.

In summary, right now you have a couple of options, change this to be `apiv1.SHA256WithRSA` or add `SignatureAlgorithm: x509.SHA256WithRSAPSS` to the templates in this file and be aware of the known issues.

maraino (changes requested):
See my comment about RSA-PSS.

cmd/step-pkcs11-init/main.go
maraino: Related https://github.com/smallstep/certificates/issues/952

maraino:
Hi @gdbelvin, have you been able to use [step-kms-plugin](https://github.com/smallstep/step-kms-plugin), can I close this as now these scripts are deprecated?

maraino:
`step-pkcs11-init` will be removed after the v0.23.0 release.
`step` + `step-kms-plugin` can be used to create customized certificates using PKCS#11

PR #954, opened 2022-06-15
all: reformat all go files with the next gofmt (Go 1.19)

shuLhan:
There are some changes that manually edited, for example using '-' as
default list and grouping imports.


maraino (commented):
Thanks @shuLhan, I'm merging this right away.

PR #955, opened 2022-06-15
cas/cloudcas: update test on createPublicKey for the next Go release

shuLhan:
The next Go release call panic on elliptic.Marshal [1][2], which
affect the test case fail_ec_marshal on createPublicKey.

This changes fix this by initializing the P and B in test case
PublicKey CurveParams to prevent panic.

[1] https://github.com/golang/go/issues/50975
[2] https://github.com/golang/go/commit/a218b3520a500254cc82b996b79ad6f5a355021c


maraino (approved):
Thanks @shuLhan, and this works on 1.17 and 1.18 too. Merging it to master.

PR #956, opened 2022-06-15
kms/uri: fix test on Parse for the next Go release

shuLhan:
The next Go release add field OmitHost to url.URL [1] which cause the
TestParse fail.
Since the CI supports two consecutive Go versions at the same times, we
copy the uri_test.go to uri_119_test.go for testing with Go 1.19.
    
While at it, print the got and want object using the same format
(%#v) and type (*URL) for consistency.
    
[1] https://go-review.googlesource.com/c/go/+/391294

### DIFF
--- a/kms/uri/uri_119_test.go
+++ b/kms/uri/uri_119_test.go
@@ -1,6 +1,4 @@
-//TODO: Remove this file when go version in go.mod set to 1.19.
-//
-//go:build !go1.19
+//go:build go1.19
 
 package uri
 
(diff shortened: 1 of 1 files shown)


maraino (changes requested):
Oooooh, for Go 1.19+, our policy is to be able to compile and pass the tests with the last two versions of Go, hopefully, in August 1.18 and 1.19, we will need to split this test into two files, one for 1.19 and above and one for the rest. Can you move this test two to files, if I'm not mistaken, one with `//go:build go1.19` and one with `//go:build !go1.19`.

And thanks @shuLhan

shuLhan:
@maraino Done. Thanks for the review. Now I can continue setting up step-ca ;)

maraino (approved):
@shuLhan I meant to just copy the problematic script, but I can do that after.

PR #957, opened 2022-06-15
docs: set environment CI=true when building from source

shuLhan:
Without setting CI=true, the make command will always fail on
TestNew_real/fail_default_credentials and TestNew/cloudkms.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -33,7 +33,7 @@ The only prerequisites are [`go`](https://golang.org/) and make.
 
 To build from source:
 
-    make bootstrap && make
+    make bootstrap && make CI=true
 
 Find your binaries in `bin/`.
 


maraino:
For context, yes, this happens if your environment has gcloud credentials, cloudkms APIs will try to use them if they exist, causing the test to fail.

shuLhan:
Oh, I think I found the cause.  My host connect to GCP through VPN and use Google internal DNS. When cloudcas.New() called, client can resolve and connect to `metadata.google.internal` (which should not exist on public network).

When I switch the VPN all tests success. Sorry for the noise.

maraino:
@shuLhan I haven't looked into what are all the conditions, it has always failed in my local machine, but in a clean environment it does not, perhaps we should set fake credentials or a test option to force the always the error, or just skip the test completely:

```
$ go test ./cas/cloudcas
ok  	github.com/smallstep/certificates/cas/cloudcas	0.128s
```

In my local machine, it reads `$HOME/.config/gcloud/application_default_credentials.json`, and if I change the name of that directory it works.

PR #958, opened 2022-06-16
Sign certificates with the issuer signature algorithm

### LINKED ISSUE #952: RSA X509 signers should inherit the signature algorithm from the intermediate
### Description

An RSA key can sign X509 certificates using RSA-PKCS#1 v1.5 or using RSA-PSS (plus the selected hash algorithm). The default behavior of Go, if the template to be signed does not have a SignatureAlgorithm is to use SHA256+RSA-PKCS#1, the only way to sign with RSA-PSS or with a different hash algorithm is to edit the provisioner template to enforce it.

The online `step-ca` certificate doesn't have a template so we cannot fix it in the same way. If we want RSA-PSS for this we should get it from the parent, the intermediate. It might also make sense to inherit the Key type too or at least make it configurable in the ca.json.

The behavior is different for Google CloudKMS, because when you define a key you also define the signature algorithm and it cannot sign with a different one, in our implementation when we load the signer we also set the appropriate signature algorithm if you don't set any other in a template.

maraino:
### Description

An RSA key can sign another certificate using the RSA PKCS#1 and the RSA-PSS scheme, this PR will keep the signature algorithm used in the issuer in the signed certificates instead of using PKCS#1 by default.

Fixes #952

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Changed
+- Certificates signed by an issuer using an RSA key will use the same algorithm
+  and not default to PKCS #1. For example, if the issuer certificate uses
+  x509.SHA256WithRSAPSS, the signed certificate will also be signed using
+  x509.SHA256WithRSAPSS.
+
 ## [0.20.0] - 2022-05-26
 ### Added
 - Added Kubernetes auth method for Vault RAs.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Changed
+- Certificates signed by an issuer using an RSA key will use the same algorithm
+  and not default to PKCS #1. For example, if the issuer certificate uses
+  x509.SHA256WithRSAPSS, the signed certificate will also be signed using
+  x509.SHA256WithRSAPSS.
```
hslatman, line 12:
```suggestion
- Certificates signed by an issuer using an RSA key will be signed using the same algorithm as the issuer certificate was signed with. The signature will no longer default to PKCS #1. For example, if the issuer certificate was signed using RSA-PSS with SHA-256, a new certificate will also be signed using RSA-PSS with SHA-256.
```

hslatman (approved):

Issue #962, opened 2022-06-20
[Bug]: using admin commands fails if root ca has name constraints


georgmu:
### Steps to Reproduce

1. create a root CA with name constraints
permitted: DNS:example.com
2. run `step ca init --root=CA.pem --key=CA.key`
3. open ~/.step/config/ca.json, delete provisioners and add `"enableAdmin": true` to `authority`
4. start `step-ca`
   * enter a password for first provisioner
6. on a second terminal, run `step ca provisioner add acme --type ACME`
   * when asked for admin name/subject, provide `step` and the password entered while starting `step-ca`

### Your Environment

* OS - ubuntu 20.04
* `step-ca` Version - Smallstep CA/0.20.0

### Expected Behavior

add the provisioner

### Actual Behavior

```
adminHandler.authorizeToken; error verifying x5c certificate chain in token: x509: a root or intermediate certificate is not authorized to sign for this name: DNS name "step" is not permitted by any constraint
```

### Additional Context

A solution for the problem would be to ask for the subject in the CreateAdmin command in the "Create first admin" section (`authority/authority.go:584`)
When I manually change it in the source code to `admin@example.com` (which apparently fulfills the name constraint), the command succeeds.

### Contributing

Vote on this issue by adding a 👍 reaction.
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

dopey:
A temporary workaround here would be to use the default PKI when starting up the CA. Then create a new admin using `step ca admin add`, and use a name permitted by the desired name constraints. Then replace the PKI with the desired one.

PR #964, opened 2022-06-20
Ask for admin name/subject when creating creating first provisioner

georgmu:
In enableAdmin=true mode, the first provisioner is created on startup.

This patch asks for the admin name/subject when creating the provisioner.
This is necessary if the root CA has nameConstraints and only permits
certain subjects.

Otherwise, there will be an error when using the provisioner with an
admin command (e.g. `step ca admin list`). See #962

Fixes issue described in #962 if the root CA has name constraints

💔Thank you!

### DIFF
--- a/authority/authority.go
+++ b/authority/authority.go
@@ -573,7 +573,7 @@ func (a *Authority) init() error {
 		}
 		if len(provs) == 0 && !strings.EqualFold(a.config.AuthorityConfig.DeploymentType, "linked") {
 			// Create First Provisioner
-			prov, err := CreateFirstProvisioner(ctx, a.adminDB, string(a.password))
+			prov, subject, err := CreateFirstProvisioner(ctx, a.adminDB, string(a.password))
 			if err != nil {
 				return admin.WrapErrorISE(err, "error creating first provisioner")
 			}
@@ -581,7 +581,7 @@ func (a *Authority) init() error {
 			// Create first admin
 			if err := a.adminDB.CreateAdmin(ctx, &linkedca.Admin{
 				ProvisionerId: prov.Id,
-				Subject:       "step",
+				Subject:       subject,
 				Type:          linkedca.Admin_SUPER_ADMIN,
 			}); err != nil {
 				return admin.WrapErrorISE(err, "error creating first admin")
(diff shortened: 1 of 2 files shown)


dopey:
Hey @georgmu thanks for opening the issue and associated PR. Totally agree that we need some way to change the first admin name to address this use case.

I don't love the idea of prompting for it, but I'm also not sure that there are better options.

@maraino of all the possible options I can think of, prompting may be the least bad.

Options:
* prompt in `step-ca` startup if first admin does not exist (this PR).
* allow option in ca.json for first admin name
* allow flag to `step-ca` for first admin name
* allow to set when calling `step ca init --enable-admin --first-admin xxx`

The problem with most of these is that if a value is not provided, we would default to `step` (assuming we didn't want to prompt) which would leave you stuck if that name wasn't allowed. Basically, we'd still need to prompt.

maraino:
@dopey @georgmu: The proper fix is to add support for it in `step ca init`, but before that, we can have an `--admin-subject` flag in `step-ca` that sets the initial subject if we need to. We can also have an `--admin-password-file` to set a different password.

hslatman:
One of the things I added in https://github.com/smallstep/certificates/pull/1075 and https://github.com/smallstep/cli/pull/758 is the ability to provide `--admin-subject` when doing `step ca init`. It'll work when initializing a CA with Remote Management enabled (`--remote-management`), because as part of that flow the first provisioner and admin are created in the DB immediately. It won't work when migrating to DB backed provisioners from config backed provisioners, as that's performed at CA startup. We would need to add the flag to that too, but that's not really nice.

maraino:
@hslatman, if you think this PR is covered by #1075, feel free to close it.

hslatman:
@maraino it partially addresses the issue, as it only works when initializing with `--remote-management` enabled. 

When migrating from `ca.json` to DB backed provisioners, there's currently no option to provide the admin subject. Adding `--admin-subject` to the CA startup works, but is not really nice to do, IMO. Alternatively we could add a helper command that enables Remote Management, performs the migration and takes `--admin-subject`, but I doubt how often that will be used.

Besides one of the solutions above, we should probably add an (early) additional checks for the name constraints and policy configured to see if it would allows `step` or not and prevent from migrating if it isn't allowed.

hslatman:
@dopey did you intentionally close this?

dopey:
I did. There hasn't been any movement and I'm trynna clean up open pull requests that are just hanging around. If you think it should be reopened that's totally fine - I'll leave that up to you.

hslatman:
Fine with it being closed. I think the migration is not done very often.

PR #1001, opened 2022-08-05
Provisioner webhooks

areed:
#### Name of feature:
Provisioner webhooks

Augment template contexts with dynamically changing data.

https://github.com/smallstep/docs/pull/141

These can be added to any provisioner that uses templates, but the admin API must be enabled to support secrets, including the signing secrets and authorization headers.

https://github.com/smallstep/cli/pull/726
https://github.com/smallstep/crypto/pull/58
https://github.com/smallstep/linkedca/pull/24

💔Thank you!


authority/admin/api/handler.go
```diff
@@ -12,24 +12,26 @@ import (
...
 func (h *Handler) Route(r api.Router) {
-	Route(r, h.acmeResponder, h.policyResponder)
+	Route(r, h.acmeResponder, h.policyResponder, h.webhookResponder)
 }
 
 // NewHandler returns a new Authority Config Handler.
 //
 // Deprecated: use Route(r api.Router, acmeResponder ACMEAdminResponder, policyResponder PolicyAdminResponder)
-func NewHandler(auth adminAuthority, adminDB admin.DB, acmeDB acme.DB, acmeResponder ACMEAdminResponder, policyResponder PolicyAdminResponder) api.RouterHandler {
+func NewHandler(auth adminAuthority, adminDB admin.DB, acmeDB acme.DB, acmeResponder ACMEAdminResponder, policyResponder PolicyAdminResponder, webhookResponder WebhookAdminResponder) api.RouterHandler {
```
maraino, line 30: I never liked adding the responders like this. Now that we have the need to change this, we might want to do:

```go
type HandlerOption func(*Handler)

func WithACMEResponder(r ACMEAdminResponder) HandlerOption {
    return func(*Handler) {
        h.acmeResponser = r
    }
}

func WithPolicyResponder(r PolicyAdminResponder) HandlerOption {
    return func(*Handler) {
        h.policyResponser = r
    }
}

func WithWebhookResponder(r WebhookAdminResponder) HandlerOption {
    return func(*Handler) {
        h.webhookResponser = r
    }
}

func NewHandler(auth adminAuthority, adminDB admin.DB, acmeDB acme.DB, opts ...HandlerOption) api.RouterHandler {
    h := new(Handler)
    for _, fn := range opts {
        fn(h)
    }
    return h
}
```

maraino: This method is deprecated, and it should not be used. You should remove this change.

authority/admin/api/handler.go
```diff
@@ -38,7 +40,7 @@ var mustAuthority = func(ctx context.Context) adminAuthority {
 }
 
 // Route traffic and implement the Router interface.
-func Route(r api.Router, acmeResponder ACMEAdminResponder, policyResponder PolicyAdminResponder) {
+func Route(r api.Router, acmeResponder ACMEAdminResponder, policyResponder PolicyAdminResponder, webhookResponder WebhookAdminResponder) {
```
maraino, line 43: This is ok, but a `ops... RouteOption` alternative similar to the above might be nice.

authority/admin/api/webhook.go
```diff
@@ -0,0 +1,200 @@
...
+			return
+		}
+	}
+
+	secret, err := randutil.Bytes(64)
+	if err != nil {
+		render.Error(w, admin.WrapErrorISE(err, "error generating webhook secret"))
+		return
+	}
+	newWebhook.Secret = base64.StdEncoding.EncodeToString(secret)
```
maraino, line 103: Should we validate in the function above that the secret must be empty?

authority/admin/api/webhook.go
```diff
@@ -0,0 +1,200 @@
...
+		if isBadRequest(err) {
+			render.Error(w, admin.WrapError(admin.ErrorBadRequestType, err, "error creating provisioner webhook"))
+			return
+		}
+
+		render.Error(w, admin.WrapErrorISE(err, "error creating provisioner webhook"))
+		return
+	}
+
+	render.ProtoJSONStatus(w, newWebhook, http.StatusCreated)
```
maraino, line 118: Should we return the secret?

authority/admin/api/webhook.go
```diff
@@ -0,0 +1,200 @@
...
+		if isBadRequest(err) {
+			render.Error(w, admin.WrapError(admin.ErrorBadRequestType, err, "error updating provisioner webhook"))
+			return
+		}
+
+		render.Error(w, admin.WrapErrorISE(err, "error updating provisioner webhook"))
+		return
+	}
+
+	render.ProtoJSONStatus(w, newWebhook, http.StatusCreated)
```
maraino, line 199: Should we return the secret?

authority/admin/db/nosql/provisioner.go
```diff
@@ -24,6 +24,22 @@ type dbProvisioner struct {
...
+}
+
+type dbWebhook struct {
+	Name          string      `json:"name"`
+	ID            string      `json:"id"`
+	URL           string      `json:"url"`
+	Kind          string      `json:"kind"`
+	SigningSecret string      `json:"signingSecret"`
+	BearerToken   string      `json:"bearerToken"`
+	BasicAuth     dbBasicAuth `json:"basicAuth"`
```
maraino, line 42: We can get more compact json like this:
```suggestion
	BearerToken   string      `json:"bearerToken,omitempty"`
	BasicAuth     *dbBasicAuth `json:"basicAuth,omitempty"`
```

authority/provisioner/options.go
```diff
@@ -69,6 +72,9 @@ type X509Options struct {
 	// AllowWildcardNames indicates if literal wildcard names
 	// like *.example.com are allowed. Defaults to false.
 	AllowWildcardNames bool `json:"-"`
+
+	// Webhooks is a list of webhooks that can augment template data
+	Webhooks []*Webhook `json:"webhooks,omitempty"`
```
maraino, line 77: Have you added this to SSHOptions?

authority/provisioner/options.go
```diff
@@ -129,10 +135,28 @@ func CustomTemplateOptions(o *Options, data x509util.TemplateData, defaultTempla
...
 
 	return certificateOptionsFunc(func(so SignOptions) []x509util.Option {
+		var enrich = func(fn func(string, x509util.TemplateData) x509util.Option, arg1 string) x509util.Option {
+			return func(cr *x509.CertificateRequest, xOpts *x509util.Options) error {
+				if opts != nil {
+					for _, wh := range opts.Webhooks {
+						if wh.Kind != linkedca.Webhook_ENRICHING.String() {
+							continue
+						}
+						d, err := wh.Do(context.TODO(), so.WebhookClient, cr, data)
```
maraino, line 145: Use a context with a timeout

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,108 @@
...
+		Timestamp: time.Now().Format(time.RFC3339Nano),
+		CSR:       csr.Raw,
+	}
+
+	reqBytes, err := json.Marshal(reqBody)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequest("POST", url, bytes.NewReader(reqBytes))
```
maraino, line 66:
```suggestion
	req, err := http.NewRequestWithContext(ctx, "POST", url, bytes.NewReader(reqBytes))
```

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,108 @@
...
+
+	resp, err := client.Do(req)
+	if err != nil {
+		if retries > 0 {
+			retries--
+			time.Sleep(time.Second)
+			goto retry
+		}
+		return nil, err
+	}
```
maraino, line 93: We should not retry if the error is a deadline/timeout caused by the context.

authority/provisioners.go
```diff
@@ -494,9 +494,39 @@ func optionsToCertificates(p *linkedca.Provisioner) *provisioner.Options {
...
+		for _, wh := range p.Webhooks {
+			whCert := webhookToCertificates(wh)
+			ops.SSH.Webhooks = append(ops.SSH.Webhooks, whCert)
+		}
+	} else {
+		for _, wh := range p.Webhooks {
+			whCert := webhookToCertificates(wh)
+			ops.X509.Webhooks = append(ops.X509.Webhooks, whCert)
+		}
+	}
```
maraino, line 507: We should add SSH support for webhooks, not just to the SSHPOP provisioner

ca/ca.go
```diff
@@ -534,3 +542,14 @@ func dumpRoutes(mux chi.Routes) {
...
+
+func getWebhookTransport() *http.Transport {
+	return &http.Transport{
+		Proxy:                 http.ProxyFromEnvironment,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+	}
```
maraino, line 554: You might want to .Clone() the DefaultTransport:
```go
http.DefaultTransport.(*http.Transport).Clone()
```

ca/ca.go
```diff
@@ -165,6 +169,9 @@ func (ca *CA) Init(cfg *config.Config) (*CA, error) {
 		return nil, err
 	}
 
+	// TODO does this config have a renewing cert with client auth?
+	webhookTransport.TLSClientConfig = tlsConfig
```
maraino, line 173: No, this config does not have a client certificate, only a server one.

maraino: We should have a custom tlsConfig for the webhook client, using `GetClientCertificate` and `RootCAs` with the system roots and the server ones. The system roots should not be part of the server ClientCAs.

authority/admin/api/handler.go
areed: Can I delete it?

maraino: Some of them, not all, in this case, because this is our own admin you can.

ca/ca.go
```diff
@@ -154,6 +155,9 @@ func (ca *CA) Init(cfg *config.Config) (*CA, error) {
 		opts = append(opts, authority.WithDatabase(ca.opts.database))
 	}
 
+	webhookTransport := getWebhookTransport()
+	opts = append(opts, authority.WithWebhookClient(&http.Client{Transport: webhookTransport}))
```
hslatman, line 159: The client must not follow redirects. For hosted, we must ensure it never resolves to a private network range. We may need to think about how to do this when deployed in own infrastructure, because generally we can probably allow resolving to private network ranges, but there may be cases where it's not OK. We may also need to think about malicious servers exhausting resources.

authority/admin/api/handler.go
areed: d48f6740

authority/admin/api/handler.go
areed: d48f6740

authority/admin/api/webhook.go
areed: a5ef6d05
I validated inline since the behavior is different for create vs update.

authority/admin/api/webhook.go
areed: How else would they get the secret if we don't return it here?

authority/admin/api/webhook.go
areed: 1656506e

authority/admin/db/nosql/provisioner.go
areed: 113d0f95

authority/provisioners.go
areed: @maraino is there ever a case when a single provisioner can sign both x509 and ssh certificates or will it always be one or the other?

areed: actually I moved options up out of the x509 and ssh fields so it doesn't matter 0bb0a2b3

authority/provisioner/options.go
areed: 118f151

authority/provisioner/options.go
areed: 118f15105

authority/provisioner/webhook.go
areed: 118f15105

authority/provisioner/webhook.go
areed: 8348af5b

authority/tls.go
```diff
@@ -645,6 +645,85 @@ func (a *Authority) GetTLSCertificate() (*tls.Certificate, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	signer, ok := priv.(crypto.Signer)
+	if !ok {
+		return nil, errors.New("private key is not a crypto.Signer")
+
+	}
+
+	// prepare the sans: IPv6 DNS hostname representations are converted to their IP representation
```
areed, line 662: Does the client cert need SANs?

authority/tls.go
```diff
@@ -645,6 +645,85 @@ func (a *Authority) GetTLSCertificate() (*tls.Certificate, error) {
 	return &tlsCrt, nil
 }
 
+// GetTLSClientCertificate creates a new leaf certificate to be used for webhook
```
areed, line 648: Mostly copied from https://github.com/smallstep/certificates/pull/1001/commits/a30444b0d4c3f0af514b428f59d2989cf9b5b358#diff-664ff05b55cabc1ecdc81445fa819245e11fc531aa6b25a0fa6b2d737176d558R569

authority/tls.go
```diff
@@ -645,6 +645,85 @@ func (a *Authority) GetTLSCertificate() (*tls.Certificate, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Get x509 certificate template, set validity and sign it.
+	now := time.Now()
+	certTpl := template.GetCertificate()
+	certTpl.NotBefore = now.Add(-1 * time.Minute)
+	certTpl.NotAfter = now.Add(24 * time.Hour)
+	certTpl.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth}
```
areed, line 690: I added this

ca/ca.go
```diff
@@ -528,6 +531,51 @@ func (ca *CA) getTLSConfig(auth *authority.Authority) (*tls.Config, error) {
 	return tlsConfig, nil
 }
 
+// getTLSClientConfig returns a TLSConfig for the CA with a self-renewing client
+// certificate.
+func (ca *CA) getTLSClientConfig(auth *authority.Authority) (*tls.Config, error) {
```
areed, line 536: Mostly copied from https://github.com/smallstep/certificates/pull/1001/commits/a30444b0d4c3f0af514b428f59d2989cf9b5b358#diff-9033031dd64a1f3c17bc6f97c5e0c3a0375eafa6e81bbbbbaccedc6446d94c8dR471

ca/ca.go
```diff
@@ -528,6 +531,51 @@ func (ca *CA) getTLSConfig(auth *authority.Authority) (*tls.Config, error) {
...
+	} else {
+		tlsConfig = &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		}
+	}
+
+	tlsConfig.GetClientCertificate = ca.clientRenewer.GetClientCertificateForCA
+
+	// Web PKI plus this authority's roots
+	certPool, err := x509.SystemCertPool()
```
areed, line 567: In design discussion we said only web PKI. Should I leave off the authority root?

ca/ca.go
areed: a30444b

ca/ca.go
areed: 798652a08ce9882c321c16f5d168e5c94f71a7d7
With this commit you can set a policy in ca.json such as
```
"webhookClient": {
  "allow": {
    "public": true,
    "hostnames": ["customer.smallstep.svc.cluster.local"],
  }
}
```
The policy will be checked before every dial so I think it's safe to follow redirects.

Do you have anything in mind for protecting against malicious servers other than setting a timeout?

authority/tls.go
maraino: I think we can reuse the server tls.Config for this.

authority/tls.go
maraino: This is an interesting question because it can send internal names to the server. Some SAN might be useful but not required.

If we end up using the same tls.Config as the server, we might be sending internal names. It really makes this easier, but I'm not sure if this is what we should do.

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,204 @@
...
+	Signature       *ssh.Signature
+}
+
+type webhookRequestBody struct {
+	Timestamp string `json:"timestamp"`
+	//nolint:revive // acronyms can use all caps
+	X509_CSR []byte `json:"csr,omitempty"`
+	//nolint:revive // acronyms can use all caps
+	SSH_CR         *sshRequest       `json:"ssh_cr,omitempty"`
+	Certificate    *x509.Certificate `json:"certificate,omitempty"`
```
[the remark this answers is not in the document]
maraino: I can add something like that in crypto, another option is using the template representation.

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,204 @@
...
+			Key:             r.Key.Marshal(),
+			Serial:          r.Serial,
+			CertType:        r.CertType,
+			KeyId:           r.KeyId,
+			ValidPrincipals: r.ValidPrincipals,
+			ValidAfter:      r.ValidAfter,
+			ValidBefore:     r.ValidBefore,
+			Permissions:     r.Permissions,
+			Reserved:        r.Reserved,
+			Signature:       r.Signature,
```
maraino, line 123: A To-Be-Signed certificate won't have a signature.

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,204 @@
...
+	//nolint:revive // field name matches ssh.Certificate
+	KeyId           string
+	ValidPrincipals []string
+	ValidAfter      uint64
+	ValidBefore     uint64
+	Permissions     ssh.Permissions
+	Reserved        []byte
+	SignatureKey    []byte
+	Signature       *ssh.Signature
+}
```
maraino, line 60: you probably want JSON tags here. And you can just fix that revive error; at the end of the day, it doesn't matter if they don't perfectly match.

ca/ca.go
maraino: This tlsConfig is intended for servers, it doesn't include a valid RootCAs. We should allow RootCAs for Web PKI (`x509.SystemCertPool()`) and our CA root.

maraino (changes requested):
I'm still looking into this. And I don't see how device attestation will be integrated here.

authority/admin/db/nosql/provisioner.go
```diff
@@ -24,6 +24,23 @@ type dbProvisioner struct {
 	SSHTemplate  *linkedca.Template        `json:"sshTemplate"`
 	CreatedAt    time.Time                 `json:"createdAt"`
 	DeletedAt    time.Time                 `json:"deletedAt"`
+	Webhooks     []dbWebhook               `json:"webhooks,omitempty"`
```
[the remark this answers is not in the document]
hslatman: Unsure if it'll happen, but I've seen some iffy behavior when relying on the types in the `linkedca` package (or maybe a different, generated one) for persistence. It's rather easy to introduce changes to the data types without being aware where they end up. I think _not_ using the JSON representation of the `linkedca` package is safer. It is more work and code, but I think the benefits outweigh those.

ca/ca.go
hslatman: Was indeed thinking about a timeout. Another is to try to detect a really slow server response, like what https://github.com/skeeto/endlessh does to SSH clients, but maybe that's a stretch.

authority/admin/api/handler.go
```diff
@@ -4,41 +4,47 @@ import (
...
+var mustAuthority = func(ctx context.Context) adminAuthority {
+	return authority.MustFromContext(ctx)
 }
 
-// Route traffic and implement the Router interface.
-//
-// Deprecated: use Route(r api.Router, acmeResponder ACMEAdminResponder, policyResponder PolicyAdminResponder)
-func (h *Handler) Route(r api.Router) {
-	Route(r, h.acmeResponder, h.policyResponder)
+type Router struct {
```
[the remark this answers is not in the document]
areed: acff4f19

authority/admin/api/webhook.go
```diff
@@ -0,0 +1,224 @@
...
+	// auth secrets since those may have been updated in this request and we
+	// should show in the response that they changed
+	whResponse := &linkedca.Webhook{
+		Id:                   newWebhook.Id,
+		Name:                 newWebhook.Name,
+		Url:                  newWebhook.Url,
+		Kind:                 newWebhook.Kind,
+		Auth:                 newWebhook.Auth,
+		DisableTlsClientAuth: newWebhook.DisableTlsClientAuth,
+	}
```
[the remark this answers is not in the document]
areed: The provisioner passed to `authority.UpdateProvisioner` has a reference to `newWebhook`.

authority/admin/api/webhook.go
```diff
@@ -0,0 +1,224 @@
...
+			return
+		}
+		newWebhook.Secret = wh.Secret
+		newWebhook.Id = wh.Id
+		prov.Webhooks[i] = newWebhook
+		found = true
+		break
+	}
+	if !found {
+		msg := fmt.Sprintf("provisioner %s has no webhook with the name %s", prov.Name, newWebhook.Name)
```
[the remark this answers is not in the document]
areed: 782ae172

authority/admin/api/webhook.go
```diff
@@ -0,0 +1,224 @@
...
+		if wh.Name != newWebhook.Name {
+			continue
+		}
+		if newWebhook.Secret != "" && newWebhook.Secret != wh.Secret {
+			err := admin.NewError(admin.ErrorBadRequestType, "webhook secret cannot be updated")
+			render.Error(w, err)
+			return
+		}
+		newWebhook.Secret = wh.Secret
+		newWebhook.Id = wh.Id
```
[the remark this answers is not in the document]
areed: a938904
I enforced blank ID or secret. I don't want to add that to validate because I want clients to be able to use the response from create in an update request without having to delete the secret and id.

authority/admin/db/nosql/provisioner.go
```diff
@@ -209,3 +229,60 @@ func (db *DB) DeleteProvisioner(ctx context.Context, id string) error {
 
 	return db.save(ctx, old.ID, nu, old, "provisioner", provisionersTable)
 }
+
+func dbWebhooksToLinkedca(dbwhs []dbWebhook) []*linkedca.Webhook {
+	var lwhs []*linkedca.Webhook
```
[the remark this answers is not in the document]
areed: d3b6ae67

authority/admin/db/nosql/provisioner.go
```diff
@@ -209,3 +229,60 @@ func (db *DB) DeleteProvisioner(ctx context.Context, id string) error {
...
+			}
+		}
+		lwhs = append(lwhs, lwh)
+	}
+
+	return lwhs
+}
+
+func linkedcaWebhooksToDB(lwhs []*linkedca.Webhook) []dbWebhook {
+	var dbwhs []dbWebhook
```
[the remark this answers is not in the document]
areed: d3b6ae67

Makefile
```diff
@@ -29,7 +29,7 @@ ci: testcgo build
 
 bootstra%:
 	# Using a released version of golangci-lint to take into account custom replacements in their go.mod
-	$Q curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell go env GOPATH)/bin v1.42.0
+	$Q curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell go env GOPATH)/bin v1.45.2
```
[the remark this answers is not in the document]
areed: 4dd563fc

authority/config/config.go
```diff
@@ -104,6 +104,7 @@ type AuthConfig struct {
 	Backdate             *provisioner.Duration `json:"backdate,omitempty"`
 	EnableAdmin          bool                  `json:"enableAdmin,omitempty"`
 	DisableGetSSHHosts   bool                  `json:"disableGetSSHHosts,omitempty"`
+	WebhookClient        *WebhookClient        `json:"webhookClient,omitempty"`
```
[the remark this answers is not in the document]
areed: I agree it's only useful in hosted environments where untrusted users can set the webhook URL, but I don't see how certificates prevent all vulnerabilities. It's possible an internal endpoint could present a trusted certificate.

authority/admin/api/webhook.go
```diff
@@ -0,0 +1,224 @@
...
+		render.Error(w, err)
+		return
+	}
+
+	id, err := randutil.UUIDv4()
+	if err != nil {
+		render.Error(w, admin.WrapErrorISE(err, "error generating webhook id"))
+		return
+	}
+	newWebhook.Id = id
```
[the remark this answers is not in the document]
areed: The ID allows the webhook server to verify multiple webhooks on the same endpoint by looking up the signing secret for the ID.

authority/provisioner/webhook.go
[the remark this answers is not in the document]
areed: I think the template representation is a good option since users expect that to be the json format of a certificate. There's already a function to [get the x509 cert from the template](https://github.com/smallstep/crypto/blob/master/x509util/certificate.go#L88). I'll add a function in crypto to get the template from the x509.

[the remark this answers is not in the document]
areed: 2ee14b7d

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,204 @@
...
+
+type webhookRequestBody struct {
+	Timestamp string `json:"timestamp"`
```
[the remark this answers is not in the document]
areed: 8bafa507

authority/provisioner/webhook.go
areed: d976e88

authority/provisioner/webhook.go
areed: d976e88

areed:
> I'm still looking into this. And I don't see how device attestation will be integrated here.

@maraino the acme provisioner will need to be configured use an authorizing webhook since it doesn't use templates. https://github.com/smallstep/certificates/pull/1001/files#diff-7ebfe74c94aa45510ec8aea86e5b7fff5305fb8488e0a9b3b671013ff66dc75bR186

The webhook server will get a copy of the certificate in json template format. Will it be able to find the device ID in that cert and make an authorization decision?

authority/provisioners.go
maraino: Most provisioners can issue both ssh and x509 provisioners. Moving the Webhooks to the top of the options can work, but for most enriching webhooks would be very type-specific. For example, an SSH webhook might add a GitHub username, which doesn't make sense in X509. And an X509 webhook, might add specific URI SAN.

At the same time, if we use the same webhooks, we will need that all webhooks know how to handle the given data and at least return `{}` if that webhook doesn't work on an SSH certificate.

ca/ca.go
maraino: I prefer a different `tls.Config` for clients. They should share the certificate renewer, though.

ca/ca.go
maraino: This client policy should not be necessary, the OS and TLS address most concerts. If I want to restrict an IP, I will use a firewall instead of this.

authority/tls.go
maraino: We will use the same certificate, but we should not use the same `tls.Client`

ca/ca.go
maraino: We should add this back, passing the server tlsConfig or the renewer.

ca/ca.go
maraino: Authority root + WebPKI

authority/config/config.go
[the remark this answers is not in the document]
maraino: Only admins can create/update new provisioners. There's no reason to add a second level of admin that can edit the ca.json. Kill it.

authority/provisioner/webhook.go
maraino: We will use the x509util.Certificate

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,203 @@
...
+	Serial          uint64          `json:"serial"`
+	CertType        uint32          `json:"certType"`
+	KeyID           string          `json:"keyID"`
+	ValidPrincipals []string        `json:"validPrincipals"`
+	ValidAfter      uint64          `json:"validAfter"`
+	ValidBefore     uint64          `json:"validBefore"`
+	Permissions     ssh.Permissions `json:"permissions"`
+	Reserved        []byte          `json:"reserved"`
+	SignatureKey    []byte          `json:"signatureKey"`
+	Signature       *ssh.Signature  `json:"Signature"`
```
maraino, line 59: This won't generally exists, but it should be `signature` and `omitempty` as it will be nil.
```suggestion
	Signature       *ssh.Signature  `json:"signature,omitempty"`
```

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,203 @@
...
+		reqBody.SSH_CR = &sshRequest{
+			Type:       r.Type,
+			KeyID:      r.KeyID,
+			Principals: r.Principals,
+		}
+		if r.Key != nil {
+			reqBody.SSH_CR.Key = r.Key.Marshal()
+		}
+	case *x509.Certificate:
+		reqBody.Certificate = x509util.FromX509Certificate(r)
```
maraino, line 110: I think this will change with a new versions of crypto.

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,203 @@
...
+	case *ssh.Certificate:
+		reqBody.SSHCertificate = &sshCert{
```
maraino, line 112: It would make sense to use `sshutil.Certificate` here. If we cannot, it's ok. But at least make sure that the types and json tags match.

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "POST", url, bytes.NewReader(reqBytes))
+	if err != nil {
+		return nil, err
+	}
+
+	secret, err := base64.StdEncoding.DecodeString(w.SigningSecret)
```
maraino, line 142: The name `SigningSecret` is weird in this context. Just `Secret` or if we want to be verbose `MACSecret` would make more sense.

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,203 @@
...
+		tlsConfig.GetClientCertificate = nil
+		tlsConfig.Certificates = nil
+		transport.TLSClientConfig = tlsConfig
+		client = &http.Client{
+			Transport: transport,
+		}
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		if err == context.DeadlineExceeded {
```
maraino, line 172: The linter might complain unless you do `errors.Is(err, context.DeadlineExceeded)`

authority/provisioners.go
```diff
@@ -494,9 +494,34 @@ func optionsToCertificates(p *linkedca.Provisioner) *provisioner.Options {
...
 		}
 	}
+	for _, wh := range p.Webhooks {
+		whCert := webhookToCertificates(wh)
+		ops.Webhooks = append(ops.Webhooks, whCert)
+	}
 	return ops
 }
 
+func webhookToCertificates(wh *linkedca.Webhook) *provisioner.Webhook {
```
maraino, line 504: We probably need a webhookToLinkedca too.

authority/ssh.go
```diff
@@ -161,6 +162,8 @@ func (a *Authority) SignSSH(ctx context.Context, key ssh.PublicKey, opts provisi
 	// Set backdate with the configured value
 	opts.Backdate = a.config.AuthorityConfig.Backdate.Duration
 
+	opts.WebhookClient = a.webhookClient
```
maraino, line 165: we don't need this if we pass the client to the authorizer function, it might look cleaner.

authority/tls.go
```diff
@@ -92,6 +93,8 @@ func (a *Authority) Sign(csr *x509.CertificateRequest, signOpts provisioner.Sign
 	// Set backdate with the configured value
 	signOpts.Backdate = a.config.AuthorityConfig.Backdate.Duration
 
+	signOpts.WebhookClient = a.webhookClient
```
maraino, line 96: we don't need this if we pass the client to the authorizer function.

authority/ssh.go
```diff
@@ -268,6 +275,13 @@ func (a *Authority) SignSSH(ctx context.Context, key ssh.PublicKey, opts provisi
 		)
 	}
 
+	// Certificate final authorizers
+	for _, a := range authorizers {
+		if err := a.Authorize(certTpl, opts); err != nil {
+			return nil, errs.ForbiddenErr(err, "request not allowed by webhook server")
+		}
+	}
```
maraino, line 283: Just noting that this might change if we use the template instead.

authority/tls.go
```diff
@@ -223,6 +230,16 @@ func (a *Authority) Sign(csr *x509.CertificateRequest, signOpts provisioner.Sign
...
 
+	// Certificate final authorizers
+	for _, a := range certAuthorizers {
+		if err := a.Authorize(leaf, signOpts); err != nil {
+			return nil, errs.ApplyOptions(
+				errs.ForbiddenErr(err, "error creating certificate"),
+				opts...,
+			)
+		}
+	}
```
maraino, line 241: Just noting that this might change if we use the template instead.

maraino (changes requested):

ca/ca.go
areed: 627ee201

ca/ca.go
areed: a78aa3c2

ca/ca.go
areed: 627ee201aaef760bbcf7a1ff26daf6caae9ed34a

ca/ca.go
areed: 627ee201aaef760bbcf7a1ff26daf6caae9ed34a

authority/provisioner/acme.go
```diff
@@ -183,6 +183,7 @@ func (p *ACME) AuthorizeSign(ctx context.Context, token string) ([]SignOption, e
 		defaultPublicKeyValidator{},
 		newValidityValidator(p.ctl.Claimer.MinTLSCertDuration(), p.ctl.Claimer.MaxTLSCertDuration()),
 		newX509NamePolicyValidator(p.ctl.getPolicy().getX509()),
+		newWebhooksAuthorizer(p.Options.GetWebhooks(), nil),
```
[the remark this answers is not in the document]
areed: c3ab816fa2329cce8625bd71d932206d93614a68

authority/provisioner/options.go
```diff
@@ -129,10 +144,31 @@ func CustomTemplateOptions(o *Options, data x509util.TemplateData, defaultTempla
...
 
 	return certificateOptionsFunc(func(so SignOptions) []x509util.Option {
+		var enrich = func(fn func(string, x509util.TemplateData) x509util.Option, arg1 string) x509util.Option {
+			return func(cr *x509.CertificateRequest, xOpts *x509util.Options) error {
+				if opts != nil {
+					for _, wh := range o.Webhooks {
+						if wh.Kind != linkedca.Webhook_ENRICHING.String() {
+							continue
+						}
+						resp, err := wh.Do(context.Background(), so.WebhookClient, cr, data)
```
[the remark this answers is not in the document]
areed: Timeout is set in webhook.Do so it only has to be in one place and could be made configurable later if needed.

authority/provisioner/options.go
```diff
@@ -129,10 +144,31 @@ func CustomTemplateOptions(o *Options, data x509util.TemplateData, defaultTempla
 	}
 
 	return certificateOptionsFunc(func(so SignOptions) []x509util.Option {
+		var enrich = func(fn func(string, x509util.TemplateData) x509util.Option, arg1 string) x509util.Option {
```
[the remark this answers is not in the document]
areed: c3ab816fa2329cce8625bd71d932206d93614a68

authority/provisioner/sign_options.go
```diff
@@ -40,6 +44,11 @@ type CertificateValidator interface {
 	Valid(cert *x509.Certificate, opts SignOptions) error
 }
 
+// CertificateAuthorizer is an interface used to authorize a given X.509 certificate.
+type CertificateAuthorizer interface {
+	Authorize(req *x509.Certificate, opts SignOptions) error
+}
```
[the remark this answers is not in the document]
areed: c3ab816fa2329cce8625bd71d932206d93614a68

authority/provisioner/sign_ssh_options.go
```diff
@@ -37,6 +41,11 @@ type SSHCertValidator interface {
 	Valid(cert *ssh.Certificate, opts SignSSHOptions) error
 }
 
+// SSHCertAuthorizer is an interface used to authorize an SSH certificate.
+type SSHCertAuthorizer interface {
+	Authorize(req *ssh.Certificate, opts SignSSHOptions) error
+}
```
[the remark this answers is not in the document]
areed: c3ab816fa2329cce8625bd71d932206d93614a68

authority/provisioner/ssh_options.go
```diff
@@ -124,10 +127,31 @@ func CustomSSHTemplateOptions(o *Options, data sshutil.TemplateData, defaultTemp
 	}
 
 	return sshCertificateOptionsFunc(func(so SignSSHOptions) []sshutil.Option {
+		var enrich = func(fn func(string, sshutil.TemplateData) sshutil.Option, arg1 string) sshutil.Option {
```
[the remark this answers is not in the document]
areed: c3ab816fa2329cce8625bd71d932206d93614a68

authority/provisioner/webhook.go
areed: 6c6c618ca866b4b1b266497380e3f0dde8f2c7cc

authority/provisioner/webhook.go
areed: 6c6c618ca866b4b1b266497380e3f0dde8f2c7cc

authority/provisioners.go
areed: Where would that be used?

authority/provisioner/webhook.go
areed: acf0965c

authority/provisioner/webhook.go
areed: f6205018d8bcc271b71f82dd37ceef17c001f84d

authority/provisioner/options.go
```diff
@@ -29,6 +33,9 @@ func (fn certificateOptionsFunc) Options(so SignOptions) []x509util.Option {
 type Options struct {
 	X509 *X509Options `json:"x509,omitempty"`
 	SSH  *SSHOptions  `json:"ssh,omitempty"`
+
+	// Webhooks is a list of webhooks that can augment template data
+	Webhooks []*Webhook `json:"webhooks,omitempty"`
```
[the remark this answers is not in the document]
areed: 32ede5c7

authority/config/config.go
```diff
@@ -131,6 +132,10 @@ func (c *AuthConfig) Validate(audiences provisioner.Audiences) error {
 	// Initialize required fields.
 	c.init()
 
+	if err := c.WebhookClient.init(); err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
areed: 14b1f4602cfbcd36c6dda5317f38b33a519fb2db

authority/ssh.go
```diff
@@ -199,6 +205,26 @@ func (a *Authority) SignSSH(ctx context.Context, key ssh.PublicKey, opts provisi
...
+		whEnrichReq, err := webhook.NewRequestBody(
+			webhook.WithSSHCertificateRequest(cr),
+		)
+		if err != nil {
+			return nil, errs.ApplyOptions(
+				errs.InternalServerErr(err),
+				errs.WithKeyVal("signOptions", signOpts),
+			)
+		}
+		err = webhookCtl.Enrich(whEnrichReq)
```
maraino, line 219: This doesn't look like the natural place for this call. The template data is always defined in `CustomSSHTemplateOptions` (or `CustomTemplateOptions` for X509) in the provisioner package.

You have managed to solve it with a setter. But I think in a previous commit you had that embedded in the above methods, which feels more natural. It was a little complex because there were too many composing functions in the same function, but it was easy to solve by extracting that code to a new method.

authority/ssh.go
```diff
@@ -268,6 +294,24 @@ func (a *Authority) SignSSH(ctx context.Context, key ssh.PublicKey, opts provisi
...
+			)
+		}
+		err = webhookCtl.Authorize(whAuthBody)
+		if err != nil {
+			return nil, errs.ApplyOptions(
+				errs.ForbiddenErr(err, "authority.SignSSH: error signing certificate"),
+			)
+		}
+	}
+
```
maraino, line 314: This feels natural here, but you might want to extract this to a new function.

authority/tls.go
```diff
@@ -132,14 +134,46 @@ func (a *Authority) Sign(csr *x509.CertificateRequest, signOpts provisioner.Sign
...
+		err = webhookCtl.Enrich(whEnrichReq)
+		if err != nil {
+			return nil, errs.ApplyOptions(
+				errs.ForbiddenErr(err, err.Error()),
+				errs.WithKeyVal("csr", csr),
+				errs.WithKeyVal("signOptions", signOpts),
+			)
+		}
+	}
+
```
maraino, line 176: Ditto, this seems more appropriate in `CustomTemplateOptions`.

authority/tls.go
```diff
@@ -132,14 +134,46 @@ func (a *Authority) Sign(csr *x509.CertificateRequest, signOpts provisioner.Sign
...
+		var attested *webhook.AttestationData
+		if attData != nil {
+			attested = &webhook.AttestationData{
+				PermanentIdentifier: attData.PermanentIdentifier,
+			}
+		}
+		whEnrichReq, err := webhook.NewRequestBody(
+			webhook.WithX509CertificateRequest(csr),
+			webhook.WithAttestationData(attested),
+		)
```
maraino, line 159: The token should be part of the body too.

authority/tls.go
```diff
@@ -229,6 +263,26 @@ func (a *Authority) Sign(csr *x509.CertificateRequest, signOpts provisioner.Sign
...
+		}
+		err = webhookCtl.Authorize(whAuthBody)
+		if err != nil {
+			return nil, errs.ApplyOptions(
+				errs.ForbiddenErr(err, "error creating certificate"),
+				opts...,
+			)
+		}
+	}
+
```
maraino, line 285: I would extract this to a new function.

maraino (changes requested):

acme/order.go
```diff
@@ -194,6 +194,11 @@ func (o *Order) Finalize(ctx context.Context, db DB, csr *x509.CertificateReques
 	if err != nil {
 		return WrapErrorISE(err, "error retrieving authorization options from ACME provisioner")
 	}
+	for _, signOp := range signOps {
+		if wc, ok := signOp.(*provisioner.WebhookController); ok {
+			wc.TemplateData = data
+		}
+	}
```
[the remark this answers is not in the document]
areed: 10399d92

authority/admin/api/webhook.go
```diff
@@ -0,0 +1,235 @@
...
+	secret, err := randutil.Bytes(64)
+	if err != nil {
+		render.Error(w, admin.WrapErrorISE(err, "error generating webhook secret"))
+		return
+	}
+	newWebhook.Secret = base64.StdEncoding.EncodeToString(secret)
+
+	prov.Webhooks = append(prov.Webhooks, newWebhook)
+
+	auth := mustAuthority(ctx)
```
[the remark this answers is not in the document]
areed: 10399d92

authority/admin/api/webhook.go
```diff
@@ -0,0 +1,235 @@
...
+			found = true
+			break
+		}
+	}
+	if !found {
+		render.JSONStatus(w, DeleteResponse{Status: "ok"}, http.StatusOK)
+		return
+	}
+
+	auth := mustAuthority(ctx)
```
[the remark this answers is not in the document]
areed: 10399d92

authority/admin/api/webhook.go
```diff
@@ -0,0 +1,235 @@
...
+		break
+	}
+	if !found {
+		msg := fmt.Sprintf("provisioner %q has no webhook with the name %q", prov.Name, newWebhook.Name)
+		err := admin.NewError(admin.ErrorNotFoundType, msg)
+		render.Error(w, err)
+		return
+	}
+
+	auth := mustAuthority(ctx)
```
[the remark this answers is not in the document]
areed: 10399d92

authority/admin/db/nosql/provisioner.go
```diff
@@ -24,6 +24,24 @@ type dbProvisioner struct {
...
+	Username string `json:"username"`
+	Password string `json:"password"`
+}
+
+type dbWebhook struct {
+	Name                 string      `json:"name"`
+	ID                   string      `json:"id"`
+	URL                  string      `json:"url"`
+	Kind                 string      `json:"kind"`
+	SigningSecret        string      `json:"signingSecret"`
```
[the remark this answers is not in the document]
areed: 10399d92

authority/admin/db/nosql/provisioner.go
```diff
@@ -24,6 +24,24 @@ type dbProvisioner struct {
...
+	BearerToken          string      `json:"bearerToken,omitempty"`
+	BasicAuth            dbBasicAuth `json:"basicAuth,omitempty"`
```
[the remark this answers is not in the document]
areed: 10399d92

authority/admin/db/nosql/provisioner.go
```diff
@@ -209,3 +230,68 @@ func (db *DB) DeleteProvisioner(ctx context.Context, id string) error {
...
+			DisableTlsClientAuth: dbwh.DisableTLSClientAuth,
+			CertType:             linkedca.Webhook_CertType(linkedca.Webhook_CertType_value[dbwh.CertType]),
+		}
+		if dbwh.BearerToken != "" {
+			lwh.Auth = &linkedca.Webhook_BearerToken{
+				BearerToken: &linkedca.BearerToken{
+					BearerToken: dbwh.BearerToken,
+				},
+			}
+		} else if dbwh.BasicAuth.Username != "" || dbwh.BasicAuth.Password != "" {
```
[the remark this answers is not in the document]
areed: 10399d92

authority/provisioner/jwk.go
```diff
@@ -278,6 +280,8 @@ func (p *JWK) AuthorizeSSHSign(ctx context.Context, token string) ([]SignOption,
 		&sshCertDefaultValidator{},
 		// Ensure that all principal names are allowed
 		newSSHNamePolicyValidator(p.ctl.getPolicy().getSSHHost(), p.ctl.getPolicy().getSSHUser()),
+		// Call webhooks
+		p.ctl.newWebhookController(data, linkedca.Webhook_X509),
```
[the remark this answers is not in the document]
areed: 10399d92

authority/ssh.go
```diff
@@ -199,6 +205,26 @@ func (a *Authority) SignSSH(ctx context.Context, key ssh.PublicKey, opts provisi
 		Key:        key,
 	}
 
+	// Call enriching webhooks
+	if webhookCtl != nil {
+		whEnrichReq, err := webhook.NewRequestBody(
+			webhook.WithSSHCertificateRequest(cr),
```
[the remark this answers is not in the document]
areed: 10399d92

authority/ssh.go
areed: I think it works well here now that all enriching webhooks are also authorizing webhooks. I think the most common error will be the webhook does not allow the request to continue. It will also be easier to add in authority-level webhooks in the future.

authority/ssh.go
areed: 10399d92

authority/tls.go
areed: 10399d92

authority/tls.go
areed: 10399d92

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,204 @@
...
+		return nil, err
+	}
+	url := buf.String()
+
+	switch tmpl := data.(type) {
+	case x509util.TemplateData:
+		reqBody.Token = tmpl[x509util.TokenKey]
+	case sshutil.TemplateData:
+		reqBody.Token = tmpl[sshutil.TokenKey]
+	}
```
maraino, line 127: Although I insisted on this before, I later realized that these tokens might be problematic for a couple of reasons:
1. A K8sSA token can be reused multiple times. The webhook can misuse it to get fake certificates.
2. A webhook can misuse any other token to get its own certificate before responding.

I would comment those lines for the moment, with a note about the reasons, and we can discuss what we want to do with this.

authority/provisioner/webhook.go
```diff
@@ -0,0 +1,204 @@
...
+	defer cancel()
+
+	retries := 1
+retry:
+	reqBody.Timestamp = time.Now()
+
+	reqBytes, err := json.Marshal(reqBody)
+	if err != nil {
+		return nil, err
+	}
```
maraino, line 138: I believe the retry should have the same body, so I would move the retry to later.

I'm not sure right now if you will need to recreate the Request or not because the body reader might be consumed if we try to use it twice.

go.mod
```diff
@@ -146,9 +146,10 @@ require (
 )
 
 // replace github.com/smallstep/nosql => ../nosql
-// replace go.step.sm/crypto => ../crypto
+replace go.step.sm/crypto => ../crypto
+
 // replace go.step.sm/cli-utils => ../cli-utils
-// replace go.step.sm/linkedca => ../linkedca
+replace go.step.sm/linkedca => ../linkedca
```
maraino, line 152: Make sure to fix this before merging.

ca/adminClient.go
```diff
@@ -1101,6 +1101,103 @@ retry:
...
+	tok, err := c.generateAdminToken(u)
+	if err != nil {
+		return nil, fmt.Errorf("error generating admin token: %w", err)
+	}
+	req, err := http.NewRequest(http.MethodPost, u.String(), bytes.NewReader(body))
+	if err != nil {
+		return nil, fmt.Errorf("creating POST %s request failed: %w", u, err)
+	}
+	req.Header.Add("Authorization", tok)
+retry:
```
maraino, line 1120: Make sure the `bytes.NewReader(body)` doesn't get consumed if there's a retry. We should move this retry and the ones below if it gets consumed.

I commented on something similar before, in that method, the retry can also go before the `Do(req)` if we can reuse the same request.

maraino (changes requested):
@areed if you just fix the tokens (see my current reasoning), retries (fix one, and check the others), and go.mod. This looks good.

authority/provisioner/webhook.go
areed: 58be425

authority/provisioner/webhook.go
areed: 58be425

go.mod
areed: 58be425

ca/adminClient.go
areed: 58be425

webhook/types.go
```diff
@@ -0,0 +1,71 @@
...
+// AttestationData is data validated by acme device-attest-01 challenge
+type AttestationData struct {
+	PermanentIdentifier string `json:"permanentIdentifier"`
+}
+
+// RequestBody is the body sent to webhook servers.
+type RequestBody struct {
+	Timestamp time.Time `json:"timestamp"`
+	// Only set after successfully completing acme device-attest-01 challenge
+	AttestationData *AttestationData `json:"attestationData"`
```
maraino, line 63:
```suggestion
	AttestationData *AttestationData `json:"attestationData,omitempty"`
```

maraino (approved):
lgtm, but there's a missing `omitempty`, I've added a suggestion for it

hslatman:
🎉 🎉

PR #1006, opened 2022-08-09
Validate revocation serial number

### LINKED ISSUE #972: [Bug]: inconsistent nkey format in revoked_x509_certs table
### Steps to Reproduce

When revoking a certificarte using step client with mTLS (e.g. `step ca revoke --cert cert.pem --key cert.key ...`), the serial number of the revoked cert is stored in decimal format in the `nkey` column in `revoked_x509_certs` table. It's the same format used in `x509_certs` table and seems to be the standard/default one.

When the certificate is revoked using step client and the certificates serial number (e.g. `step ca revoke 9de5d852dda5c2869fe45d8336df008e ...`), the serial number of the revoked cert is stored in hexadecimal format in the `nkey` column in `revoked_x509_certs` table. 

It seems that the serial number is taken as-is / as it was given during the certificate revoke operation and that it is not normalized before beeing stored in the database. (https://github.com/smallstep/certificates/issues/718)
```
$ psql -qtnA -d "${dbConn}" -c "select encode(nkey, 'escape')::text from x509_certs;"
209882218884616561056554940555404968078
236816592456780696721919387019098694597
141597777458191518252926419788696651342
164251378795451772745182356178908542026
123632492247848079217495460536869063423
191988272827305694745870852859674624533

$ psql -qtnA -d "${dbConn}" -c "select encode(nkey, 'escape')::text from revoked_x509_certs;"
7B91AA35C41D2DA56E2A4E3799F4F44A
191988272827305694745870852859674624533
123632492247848079217495460536869063423
6A86BD6E727E27742AEB1D7D5F58064E
f28b0ea8b52648a920729f56a40e5395
9de5d852dda5c2869fe45d8336df008e
```

### Your Environment

* docker image smallstep/step-ca:0.20.0
* docker image smallstep/step-cli:0.20.0

### Expected Behavior

The same serial number format in all tables would be appreciated. Decimal format seems to be the preferred/default/standard one...

### Actual Behavior

`revoked_x509_certs` table uses mixed serial number formats (as is / as provided during revoke operation)

### Additional Context

_No response_

### Contributing

dopey:
Fixes #972

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Certificates signed by an issuer using an RSA key will be signed using the same algorithm as the issuer certificate was signed with. The signature will no longer default to PKCS #1. For example, if the issuer certificate was signed using RSA-PSS with SHA-256, a new certificate will also be signed using RSA-PSS with SHA-256.
 - Support two latest versions of Go (1.18, 1.19)
+- Vadlidate revocation serial number (either base 10 or prefixed with an appropriate base)
 
 ## [0.20.0] - 2022-05-26
 ### Added
--- a/api/revoke.go
+++ b/api/revoke.go
@@ -36,7 +36,7 @@ func (r *RevokeRequest) Validate() (err error) {
 	}
 	sn, ok := new(big.Int).SetString(r.Serial, 0)
 	if !ok {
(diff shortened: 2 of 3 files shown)


maraino (approved):
Nobody is going to send the serial number in binary or octal format, everybody is going to use decimal or hexadecimal. Even though it is supported, you can return an error saying to use base 10 or 16 with the 0x prefix.

PR #1008, opened 2022-08-11
Endpoint ID

maraino:
### Description

When an RA is used, this PR sends an endpoint ID to the CA so it can group the certs for the RA itself.

### DIFF
--- a/cas/apiv1/requests.go
+++ b/cas/apiv1/requests.go
@@ -52,12 +52,13 @@ const (
 
 // CreateCertificateRequest is the request used to sign a new certificate.
 type CreateCertificateRequest struct {
-	Template    *x509.Certificate
-	CSR         *x509.CertificateRequest
-	Lifetime    time.Duration
-	Backdate    time.Duration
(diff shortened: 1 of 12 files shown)


cas/apiv1/requests.go
```diff
@@ -52,20 +52,21 @@ const (
...
-	Backdate    time.Duration
-	RequestID   string
-	Provisioner *ProvisionerInfo
+	Template     *x509.Certificate
+	CSR          *x509.CertificateRequest
+	Lifetime     time.Duration
+	Backdate     time.Duration
+	RequestID    string
+	Provisioner  *ProvisionerInfo
+	IsServerCert bool
```
dopey, line 61: I don't love this name. I feel like it can be easily confused with the existing X509 concepts of client and server certificates.

What about `IsCAServerCert`

maraino: Sure 8bd01742

dopey (approved):
lgtm

PR #1014, opened 2022-08-16
Check for DNS name validity

dopey:


maraino (approved):
lgtm but we might want to add some tests. If you need a name that will fail the sanitizename you can use `xn--bücher.example.com`

maraino (approved):
lgtm

PR #1040, opened 2022-09-02
Add `tpm` attestation with configurable roots (WIP)

hslatman:
Basic implementation of configurable roots for TPM attestation verification.

Also been trying out https://github.com/ryboe/q for debugging. It's sweet! Will delete it before merging.

@maraino I noticed the verification part is fairly similar to the current `step` method. That's a good sign, I think 🙂 There's still some work to be done, like being able to provide the intermediates (albeit these ideally come from the AK chain), doing the proper thing with unhandled critical extensions, etc.


maraino:
@hslatman we should probably create a PR for roots/intermediates before merging this, so I can adapt the current implementation.

hslatman:
@maraino do you mean a new PR that adds known roots/intermediates as embedded resources in some form, so that it becomes easier to configure those? Or do you want support for intermediates added to this PR (as it only has roots now)?

maraino:
> @maraino do you mean a new PR that adds known roots/intermediates as embedded resources in some form, so that it becomes easier to configure those? Or do you want support for intermediates added to this PR (as it only has roots now)?

As a new smaller PR similar to the challenges and attestation formats, they can be easily integrated into master without more complex stuff and add cli support for it.

hslatman:
Superseded by #1063

Issue #1039, opened 2022-09-02
Support Gitlab CI Generated JWT within JWK provisioner


unreality:
## Support Gitlab CI Generated JWT within JWK provisioner

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Gitlab currently publishes a JWKS endpoint which details the keys used when creating an authentication JWT inside its CI/CD pipelines: 
* V1 Token https://docs.gitlab.com/ee/ci/examples/authenticating-with-hashicorp-vault/index.html#how-it-works
* V2 Token https://gitlab.com/gitlab-org/gitlab/-/merge_requests/72555

step-ca's JWK provisioner could be enhanced to accept these JWT as authorisation to interact with the CA (for example to issue new certs, or obtain SSH certificates for deployment jobs).

This would require two changes:
* Add a mechanism to the JWK provisioner to grab key data from a JWKS endpoint
* Enhance https://github.com/smallstep/certificates/blob/d718c69ad33e0a8f88f3f2348a910189c9a7365e/authority/provisioner/collection.go#L85 to find the provisioner by provisioner name, issuer, or some other mechanism. The JWT published by Gitlab does not contain enough information to find the provisioner via the existing methods.

## Why is this needed?

While a somewhat niche use-case (Gitlab pipeline support), allowing JWKS endpoints will allow greater flexibility in the JWK provisioner. I believe Github also publishes a similar token during its pipeline executions, so these changes could enable support for that as well.

mmalone:
My hope is that this does *not* become a niche use case. I think this is a really useful feature. I'm hopeful that stronger TPM-attestation based authentication is available in CI in the future, but this is a nice step forward.

Do you know if the JWTs issued by Gitlab conform to the OIDC "identity token" spec? There's a good chance that they do (Github has a similar feature, and their tokens are conforming OIDC identity tokens). If so, a better solution would be to use an OIDC provisioner. For this to work, Gitlab would also need to publish an "OIDC well-known endpoint" (e.g., [here's Google's](https://accounts.google.com/.well-known/openid-configuration)). The advantage here is that certain cryptographic parameters are available via discovery, and don't need to be manually configured by end users.

[...]

PR #1043, opened 2022-09-05
Allow missing Email claim in OIDC tokens

unreality:
#### Name of feature:
Remove requirement of needing `Email` claim on OIDC tokens for SSH Signing. This pull removes the requirement for an `Email` claim in the token, and if it is missing will pass a minimal data set to the template renderer with no Principals defined, and using the `Subject` claim as KeyID.

This way a user may setup an OIDC provisioner that accepts JWTs from the CI/CD pipeline, and use templates to strictly control what principals are added to the issued Certificates (in the case of SSH certificates)

JWTs issued by several CI/CD pipelines don't include `Email` claims [Gitlab](https://docs.gitlab.com/ee/ci/cloud_services/#how-it-works), Github, etc

Removing the requirement for an `Email` claim allows a token to be resolved and SSH certs to be issued when using these JWTs.

* [Gitlab JWT documentation](https://docs.gitlab.com/ee/ci/cloud_services/#how-it-works)
* [Github JWT documentation](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers)
* Resolves https://github.com/smallstep/certificates/issues/1039

### DIFF
--- a/authority/provisioner/oidc.go
+++ b/authority/provisioner/oidc.go
@@ -376,31 +376,46 @@ func (o *OIDC) AuthorizeSSHSign(ctx context.Context, token string) ([]SignOption
 	if err != nil {
 		return nil, errs.Wrap(http.StatusInternalServerError, err, "oidc.AuthorizeSSHSign")
 	}
-	// Enforce an email claim
+
+	if claims.Subject == "" {
+		return nil, errs.Unauthorized("oidc.AuthorizeSSHSign: failed to validate oidc token payload: subject not found")
+	}
+
+	var data sshutil.TemplateData
+	var principals []string
+
 	if claims.Email == "" {
-		return nil, errs.Unauthorized("oidc.AuthorizeSSHSign: failed to validate oidc token payload: email not found")
-	}
+		// If email is empty, use the Subject claim instead to create minimal data for the template to use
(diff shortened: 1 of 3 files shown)


maraino (commented):
Hi @unreality, before merging this are you planning to create just host, or also user certificates?

If you plan to sign user certificates, how are you going to authenticate those users into the servers? Or are you setting the principals using templates?

AFAIK, even though the spec allows certificates without principals, sshd's `TrustedUserCAKeys` doesn't allow user certificates without them.

unreality:
Hi @maraino, yes the intention was to utilise templates to fill in the principals name. Users can craft a template to extract other information out of the token (which may contain principal data in another form)

A future enhancement may be to alter the way `GetIdentity()` resolves its identities so that the `sub` claim may be used to lookup an identity in addition/instead of `email` which may not be present.

maraino:
> A future enhancement may be to alter the way GetIdentity() resolves its identities so that the sub claim may be used to lookup an identity in addition/instead of email which may not be present.

We are adding webhooks that can be used to add new data to the templates, so, for example, you can send the token to the webhook and get back the principals.

maraino (approved):
Thanks @unreality, looks good

PR #1050, opened 2022-09-10
WIP: Add Apple AppAttest support

arianvp:
This is still extremely WIP but implements attestation as defined https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server

Apple AppAttest is a service for attesting App installs of your consumer application.
It is available outside of enterprise settings.   It differs slightly from Apple Enterprise Attestation
in that there is no permanent identifier that survives your app install (For privacy reasons). Instead your app generates
a unique key that is used as the identifier.

Next to that (not implemented yet) Apple allows you to detect how many attestation keys were generated over the device's lifetime to assess fraud risk https://developer.apple.com/documentation/devicecheck/assessing_fraud_risk

https://developer.apple.com/documentation/devicecheck/establishing_your_app_s_integrity

iOS apps

### DIFF
--- a/acme/challenge.go
+++ b/acme/challenge.go
@@ -1,6 +1,7 @@
 package acme
 
 import (
+	"bytes"
 	"context"
 	"crypto"
 	"crypto/ecdsa"
@@ -13,6 +14,7 @@ import (
 	"crypto/x509"
 	"encoding/asn1"
 	"encoding/base64"
+	"encoding/binary"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


arianvp:
FYI: I don't actually have an iOS device to test this.  I'm merely following Apple's docs. Hence WIP and not ready for review.

maraino:
Hi @arianvp, thanks for this, but does this attestation works as an ACME extension? 

The new MDA [ACMECertificate](https://developer.apple.com/documentation/devicemanagement/acmecertificate) for enterprises follows more or less [this draft](https://datatracker.ietf.org/doc/html/draft-bweeks-acme-device-attest-01) to perform attestation using a new ACME challenge. Related [wwdc link](https://developer.apple.com/videos/play/wwdc2022/10143/).

arianvp:
The draft will need some rewordings for this to be 'compliant' . Apart from that it's a slight variation of the existing `apple` attestation Format.  

https://github.com/brandonweeks/draft-bweeks-acme-device-attest/issues/4

maraino:
I have a branch in the CLI that implements the draft with yubikeys.

PR #1058, opened 2022-09-19
Return the internal error instead of the ACME error

### LINKED ISSUE #1057: Make ACME errors more useful
### Description

Most ACME errors have some specific details associated, but those details do not appear in the logs, instead what we have in the `error` field of the logs is just the message returned to the user, that in most cases is not very helpful.

We should show the detail message in the logs instead of the message returned to the user.

maraino:
### Description

For ACME errors, return the internal error string instead of the ACME one on the "Error() string" function. This way, the logs will have more information about the cause of an error.

Fixes #1057

### DIFF
--- a/acme/errors.go
+++ b/acme/errors.go
@@ -337,7 +337,10 @@ func (e *Error) StatusCode() int {
 
 // Error allows AError to implement the error interface.
 func (e *Error) Error() string {
-	return e.Detail
+	if e.Err == nil {
+		return e.Detail
+	}
+	return e.Err.Error()
 }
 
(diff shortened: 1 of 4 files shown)


acme/errors.go
```diff
@@ -337,7 +337,10 @@ func (e *Error) StatusCode() int {
 
 // Error allows AError to implement the error interface.
```
dopey, line 338:
```suggestion
// Error implements the error interface.
```

dopey (approved):
I made a suggestion to change the documentation.

PR #1062, opened 2022-09-21
Linter Update

dopey:


acme/common.go
```diff
@@ -112,9 +112,13 @@ type MockProvisioner struct {
 	MauthorizeOrderIdentifier func(ctx context.Context, identifier provisioner.ACMEIdentifier) error
 	MauthorizeSign            func(ctx context.Context, ott string) ([]provisioner.SignOption, error)
 	MauthorizeRevoke          func(ctx context.Context, token string) error
+<<<<<<< HEAD
```
maraino, line 115: Still looking, but this file has conflicts that needs to be resolved

maraino (changes requested):

api/render/render.go
```diff
@@ -72,12 +72,22 @@ type RenderableError interface {
...
 
+// AsRenderableError attempts to return an error type that implements the
+// RenderableError interface.
+func AsRenderableError(err error) (RenderableError, bool) {
+	//nolint:errorlint // ignore type assertion warning. casting to interface is hard.
+	if r, ok := err.(RenderableError); ok {
+		return r, true
+	}
+	return nil, false
+}
```
maraino, line 83: This linter is pita, is there any configuration to avoid linter error on interfaces? If not, we should add an issue to allow the skip of these cases.

authority/provisioner/aws.go
```diff
@@ -35,19 +35,19 @@ const awsIdentityURL = "http://169.254.169.254/latest/dynamic/instance-identity/
 const awsSignatureURL = "http://169.254.169.254/latest/dynamic/instance-identity/signature"
 
 // awsAPITokenURL is the url used to get the IMDSv2 API token
-// nolint:gosec // no credentials here
+//nolint:gosec,goimports // no credentials here
```
maraino, line 38: goimports? Here and in a couple of other places.

authority/tls.go
```diff
@@ -142,7 +142,8 @@ func (a *Authority) Sign(csr *x509.CertificateRequest, signOpts provisioner.Sign
 
 	cert, err := x509util.NewCertificate(csr, certOptions...)
 	if err != nil {
-		if _, ok := err.(*x509util.TemplateError); ok {
+		var te *x509util.TemplateError
+		if errors.As(err, &te) {
 			return nil, errs.ApplyOptions(
 				errs.BadRequestErr(err, err.Error()),
 				errs.WithKeyVal("csr", csr),
```
maraino, line 149: Ok, but in a similar conde in ssh.go you've used the template error in `errs.BadRequestErr`.

ca/ca.go
```diff
@@ -529,9 +529,9 @@ func (ca *CA) shouldServeSCEPEndpoints() bool {
 	return ca.auth.GetSCEPService() != nil
 }
 
-// nolint // ignore linters to allow keeping this function around for debugging
+//nolint:unused,deadcode // useful for debugging
```
maraino, line 532: I believe deadcode was deprecated in favor of unused. We should remove it too.

ca/client.go
```diff
@@ -1138,8 +1135,7 @@ retry:
 	}
 	var check api.SSHCheckPrincipalResponse
 	if err := readJSON(resp.Body, &check); err != nil {
-		return nil, errs.Wrapf(http.StatusInternalServerError, err, "error reading %s response",
-			[]interface{}{u, errs.WithMessage("Failed to parse response from /ssh/check-host endpoint")})
+		return nil, errs.Wrapf(http.StatusInternalServerError, err, "error reading %s response", u)
```
maraino, line 1138: You are changing the behavior here, I believe the err.WithMessage sets the message in the JSON, with your changes you will use the default one.

cmd/step-ca/main.go
```diff
@@ -176,7 +175,11 @@ $ step-ca --context=mybiz --password-file ./password.txt
 	debugProfAddr := os.Getenv("STEP_PROF_ADDR")
 	if debugProfAddr != "" {
 		go func() {
-			log.Println(http.ListenAndServe(debugProfAddr, nil))
+			srv := http.Server{
+				Addr:              debugProfAddr,
+				ReadHeaderTimeout: 15 * time.Second,
+			}
+			srv.ListenAndServe()
```
maraino, line 182: And the `log.Println`?

errs/error.go
```diff
@@ -92,7 +92,8 @@ func Wrap(status int, e error, m string, args ...interface{}) error {
 		return nil
 	}
 	_, opts := splitOptionArgs(args)
-	if err, ok := e.(*Error); ok {
+	var err *Error
+	if ok := errors.As(e, &err); ok {
```
maraino, line 96: The ok here is not necessary.

errs/error.go
```diff
@@ -108,7 +109,8 @@ func Wrapf(status int, e error, format string, args ...interface{}) error {
 		return nil
 	}
 	as, opts := splitOptionArgs(args)
-	if err, ok := e.(*Error); ok {
+	var err *Error
+	if ok := errors.As(e, &err); ok {
```
maraino, line 113: The ok here is not necessary.

errs/error.go
```diff
@@ -299,7 +300,8 @@ func Errorf(code int, format string, args ...interface{}) error {
 // ApplyOptions applies the given options to the error if is the type *Error.
 // TODO(mariano): try to get rid of this.
 func ApplyOptions(err error, opts ...interface{}) error {
-	if e, ok := err.(*Error); ok {
+	var e *Error
+	if ok := errors.As(err, &e); ok {
```
maraino, line 304: The ok here is not necessary.
```suggestion
	if errors.As(err, &e) {
```

maraino (changes requested):

errs/error.go
maraino:
```suggestion
	if errors.As(e, &err) {
```

errs/error.go
maraino:
```suggestion
	if errors.As(e, &err) {
```

api/render/render.go
maraino: We've disabled that error, but it looks like we can enable it later. 
See https://github.com/polyfloyd/go-errorlint/issues/28#issuecomment-1254080200

maraino (approved):
lgtm

PR #1063, opened 2022-09-21
Add ACME DA TPM attestation

hslatman:
Replaces https://github.com/smallstep/certificates/pull/1040, because there were some issues after merging master back into that.


acme/challenge.go
```diff
@@ -463,6 +501,333 @@ func deviceAttest01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose
...
+	if !ok {
+		return nil, NewError(ErrorBadAttestationStatementType, "x5c is malformed")
+	}
+	akCert, err := x509.ParseCertificate(akCertBytes)
+	if err != nil {
+		return nil, WrapError(ErrorBadAttestationStatementType, err, "x5c is malformed")
+	}
+
+	intermediates := x509.NewCertPool()
+	if len(x5c[1:]) > 0 {
```
maraino, line 542: There is no need to add check for the length, the for loop won't do anything if length is 0

acme/challenge.go
```diff
@@ -463,6 +501,333 @@ func deviceAttest01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose
...
+
+func parsePermanentIdentifier(der []byte) (permanentIdentifier, error) {
+	var permID permanentIdentifier
+	if _, err := asn1.UnmarshalWithParams(der, &permID, "explicit,tag:0"); err != nil {
+		return permanentIdentifier{}, err
+	}
+	return permID, nil
+}
+
+func parseSANs(ext pkix.Extension) (sans []x509util.SubjectAlternativeName, err error) {
```
maraino, line 744: This and related methods should be in x509util. And there is no need of use template fields like x509util.SubjectAlternativeName, a new struct like this might be enough:

```go
type SubjectAlternativeNames struct {
   DNSNames []string
   EmailAddresses []string
   IPAddresses []net.IP
   URIs []*url.URL
   PermanentIdentifier []string // Or a struct with the assigner, should it be a slice?
   HardwareModuleName []Whatever // Should it be a slice?
   OtherNames []OtherName
}
```

You can also skip OtherNames or any other attribute that we don't need to handle at the moment.

maraino: If DNSNames and standard SANs are in the *x509.Certificate, you don't need to add them there.

go.mod
```diff
@@ -38,19 +37,24 @@ require (
...
 	google.golang.org/grpc v1.54.0
 	google.golang.org/protobuf v1.30.0
 	gopkg.in/square/go-jose.v2 v2.6.0
 )
 
+require (
+	github.com/google/go-attestation v0.4.4-0.20220404204839-8820d49b18d9
+	golang.org/x/exp v0.0.0-20230310171629-522b1b587ee0
+)
+
```
maraino, line 57: Merge require with the ones above, and move the indirect ones below.

acme/challenge.go
```diff
@@ -463,6 +501,333 @@ func deviceAttest01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose
...
+				unhandledCriticalExtensions = append(unhandledCriticalExtensions, extOID)
+			}
+		}
+		akCert.UnhandledCriticalExtensions = unhandledCriticalExtensions
+	}
+
+	roots, ok := prov.GetAttestationRoots()
+	if !ok {
+		return nil, NewErrorISE("failed getting TPM attestation root CAs")
+	}
```
maraino, line 576: I think `ok` is false if the attestation roots are empty. If that's the case, we should change the error message.

acme/challenge.go
```diff
@@ -463,6 +501,333 @@ func deviceAttest01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose
...
+	if err != nil {
+		return nil, WrapError(ErrorBadAttestationStatementType, err, "failed parsing AK certificate SAN extension")
+	}
+
+	var permanentIdentifiers []string
+	for _, san := range sans {
+		if san.Type == x509util.PermanentIdentifierType {
+			permanentIdentifiers = append(permanentIdentifiers, san.Value)
+		}
+	}
```
maraino, line 613: All this will be something like:
```go
sans, err := x509util.ParseSubjectAlternativeNames(cert)
if err != nil {
   return nil, WrapError(Foo, err, "whatever")
}
permanentIdentifiers := sans.PermanentIdentifiers
```

acme/challenge.go
```diff
@@ -463,6 +501,333 @@ func deviceAttest01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose
...
+	if err != nil {
+		return nil, WrapError(ErrorBadAttestationStatementType, err, "failed creating key auth digest")
+	}
+	hashedKeyAuth := sha256.Sum256([]byte(keyAuth))
+
+	// verify the WebAuthn object contains the expect key authorization digest, which is carried
+	// within the encoded `certInfo` property of the attestation statement.
+	if subtle.ConstantTimeCompare(hashedKeyAuth[:], []byte(tpmCertInfo.ExtraData)) == 0 {
+		return nil, NewError(ErrorBadAttestationStatementType, "key authorization does not match")
+	}
```
maraino, line 688: Is this how this is verified in a TPM? With a YubiKey we verify with the public key in the certificate that `sig` matches keyAuth.

Also, will the last comparison work if EdDSA is used? No hardware, afaik supports it, but you're handling this case above.

maraino (changes requested):

acme/challenge.go
hslatman: I've added the function in https://github.com/smallstep/crypto/pull/209. It's currently on top of the `tpm` PR, but I can rebase it if you want. Is this going into the right direction? I think it makes sense to copy of the DNS and the other known SANs too, to get the complete set of SANs. Taken strictly it's not parsing those, so maybe `Extract...` covers the functionality better?

acme/challenge.go
hslatman: Removed the `EdDSA` case; you're right that it probably doesn't make sense. It will now return with an unsupported algorithm. 

When an application key is attested by an AK while including `QualifyingData`, that data is included with the certification data. The certified creation data is available in the `CreateAttestation` blob that's part of the `CertificationParameters`. When the input `CertificationParameters` have been verified (signed by AK, proper parameters, expected properties, etc) and the `CreateAttestation` data blob is decoded, the `QualifyingData` is present in the `ExtraData` property. The properties of the key that are checked include the data was created by a TPM, the key was generated by a TPM, cannot be exported, and some more.

The TPM spec mentions this about the property:
```
extraData | TPM2B_DATA | external info supplied by caller (often in qualifyingData parameter)

NOTE A TPM2B_DATA structure provides room for a digest and a method
indicator to indicate the components of the digest. The definition of
this method indicator is outside the scope of this specification.
```

hslatman:
@maraino I've addressed your remarks. Parsing the SANs is now done via https://github.com/smallstep/crypto/pull/209. Can you have another look?

acme/challenge.go
```diff
@@ -463,6 +500,193 @@ func deviceAttest01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose
...
+
+	certInfo, ok := att.AttStatement["certInfo"].([]byte)
+	if !ok {
+		return nil, NewError(ErrorBadAttestationStatementType, "invalid certInfo in attestation statement")
+	}
+	if len(certInfo) == 0 {
+		return nil, NewError(ErrorBadAttestationStatementType, "certInfo is empty")
+	}
+
+	alg, ok := att.AttStatement["alg"].(int64)
```
maraino, line 619: In the [webauthn-2 spec](https://www.w3.org/TR/webauthn-2/#typedefdef-cosealgorithmidentifier), the type `typedef long COSEAlgorithmIdentifier;` is identified as [long](https://heycam.github.io/webidl/#idl-long) which is int32.

maraino (commented):
Looks good, but `att.AttStatement["alg"]` is defined as 32-bits in the webauthn-2 spec. See the comment below.

acme/challenge.go
hslatman: Tried changing the type, but there's some (en|de)coding happening, resulting in the wrong type after decoding. Without depending on and implementing https://github.com/veraison/go-cose, I think we can't do more than doing the type cast after this assignment, so that's what I did in 1c38e252a6d943b27c8c88374d98db889b198d35.

acme/challenge.go
```diff
@@ -463,6 +500,193 @@ func deviceAttest01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose
...
+		Roots:         roots,
+		Intermediates: intermediates,
+		CurrentTime:   time.Now().Truncate(time.Second),
+		KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageAny},
+	})
+	if err != nil {
+		return nil, WrapError(ErrorBadAttestationStatementType, err, "x5c is not valid")
+	}
+
+	// TODO(hs): implement revocation check; Verify() doesn't perform CRL check nor OCSP lookup.
```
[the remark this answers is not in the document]
hslatman: Interesting! So we could implement an OCSP client, perform the request, and feed the response bytes to `Verify` when this lands.

acme/challenge.go
hslatman: I've opened a PR to support `RS256`: https://github.com/veraison/go-cose/pull/140. IMO not blocking this, but we might be able to use it later.

maraino (approved):
lgtm

Caddyserver go.mod shows 1.19, so using `slices` should be ok.

PR #1065, opened 2022-09-24
[actiono] Update workflows

dopey:


.github/workflows/ci.yml
```diff
@@ -19,3 +21,8 @@ jobs:
     uses: smallstep/workflows/.github/workflows/goCI.yml@main
     with:
       os-dependencies: "libpcsclite-dev"
```
maraino, line 23: We don't do CGO-enabled builds for step-ca, so this dependency is unnecessary.

maraino (changes requested):

.github/workflows/ci.yml
dopey: I tried taking it out and the CI workflows failed in multiple places where the package was expected.

maraino: This is because CGO is enabled here, but installing this dependency is okay.

maraino (approved):
lgtm

PR #1074, opened 2022-09-28
audience name should not be case sensitive, thus when case sensitivity is …

salihy:
…different between server/client, step-ca creates invalid audience error.

💔Thank you!


tashian:
Unfortunately the [JWT "aud" claim is case sensitive](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.3) and we need to respect that. It would be totally fine for the issuer to transform the string when it creates the token, but when the token is being validated it must be treated as case-sensitive.

PR #1075, opened 2022-09-29
Add `enableAdmin` and `enableACME` to Helm values.yml generation

hslatman:
This PR is an attempt at making it possible to process the `enableAdmin` and `enableACME` options, so that the generated Helm `values.yml` for `step-ca` includes the properties corresponding to these options.

It's an attempt, because it turns out to be more complicated for the `enableAdmin` option. The option is set to true when `--remote-management` is provided to `ca init`. When remote management is enabled, provisioners are stored in the database. But when we're writing the Helm configuration, we can't add provisioners to the DB (which _does_ happen for the non-Helm case), because at that time we don't have a database to write to. This results in the default provisioner not being stored in the database. 

@maraino, @dopey  what do you think about this? I don't think an `Init Container` solves this, at least not without also adding some additional logic for handling this. I think we could use some kind of auto migration process that reads the provisioners from `ca.json`, stores them in the database and deletes them from `ca.json`. The problem with that is that specifically in this case the `ca.json` is intended to be read-only, complicating this. One other way I thought of is adding the initial provisioners not through the `ca.json`, but through another parameter. It would conceptually be fairly close to an `Init Container`, I guess. Noticed there's a `postInitHook` that could be of use, maybe?


maraino:
Right now we have the `step` provisioner, but an initContainer-like functionality looks like a better idea.
https://github.com/smallstep/certificates/blob/master/authority/authority.go#L600-L616

hslatman:
Linting issue seems to be because of https://github.com/smallstep/workflows/commit/825ed94342b0fb12b4d3c27b2d5b4e3a19e3050e. Unfortunately doesn't show in which files the errors exist: https://github.com/smallstep/certificates/actions/runs/3226727287/jobs/5280607978.

dopey:
@Herman I'll take a look at those

hslatman:
@dopey your change worked; no huge number of `file exists` errors anymore 🙂 

Still would need your changes from #1104, but that will probably soon be in.

hslatman:
In 57001168a5e6757f0472eea24a20a58460fea04d I changed the Helm template output to include a default `SSHPOP` provisioner. This wasn't happening before, probably due to how some config and initialization is intertwined. I assume having it create this provisioner is the right behavior for the Helm case too.

authority/authority.go
```diff
@@ -91,8 +91,11 @@ type Authority struct {
 
 	adminMutex sync.RWMutex
 
-	// Do Not initialize the authority
+	// If true, do not initialize the authority
 	skipInit bool
+
+	// If true, does not output initialization logs
```
dopey, line 97:
```suggestion
	// If true, do not output initialization logs
```
To match the comments above.

authority/authority.go
```diff
@@ -600,20 +603,86 @@ func (a *Authority) init() error {
...
+					oldProvisioners := c.AuthorityConfig.Provisioners
+					c.AuthorityConfig.Provisioners = []provisioner.Interface{}
+					if err := c.Commit(); err != nil {
+						// Restore the provisioners in in-memory representation for consistency
+						// when writing the updated configuration fails. This is considered a soft
+						// error, so execution can continue.
+						c.AuthorityConfig.Provisioners = oldProvisioners
+						a.initLogf("Failed removing provisioners from configuration: %v", err)
+					}
+				}
```
maraino, line 653: I don't think we should rewrite this file. The format will change, it might have some annotations, and it might be the source of truth for migrating the same initial configuration to a different server. We won't use it again, that for sure, but as long as we make this clear, docs, using logs, ...

maraino (changes requested):
In my opinion, we shouldn't rewrite the configuration.

authority/authority.go
hslatman: I'm not entirely with you on this one. Sure, the file will change, the format may be different, etc. But this is not the only place in which we programmatically update the contents. It'll also not be consistent. But I'll remove it.

pki/pki.go
```diff
@@ -220,6 +221,15 @@ func WithProvisioner(s string) Option {
...
 
+// WithFirstSuperAdminSubject defines the subject of the first
+// super admin for use with the Admin API. The admin will belong
+// to the first JWK provisioner.
+func WithFirstSuperAdminSubject(s string) Option {
+	return func(p *PKI) {
+		p.options.firstSuperAdminSubject = s
+	}
+}
+
```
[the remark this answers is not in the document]
hslatman: I wanted it to be extra clear what this was about. I've debugged things in which variables weren't that clear. But I can update them.

authority/authority.go
maraino: As far as I know, the other place is when you do `step ca provisioner add --ca-config ...`, and as this is executed by a user, so he can expect that file to change. This code changes the file without any user interaction.

hslatman:
@maraino I've adopted your suggestions and added logging of the token configuration at startup.

maraino (approved):
lgtm

PR #1081, opened 2022-10-01
[action] updates and first pass at goreleaser deb

### LINKED ISSUE #1080: Add cosign signature for deb package
## Hello!

## Issue details

Cosign signature for deb package is missing (no step-ca_0.22.1_amd64.deb.sig).

## Why is this needed?

As a security product delivered outside of regular package management (install deb instead of getting it through a ppa or package repository), it would benefit security, if we could validate the signature, as we can already do on step-cli or many other packages delivered.

dopey:
Fixes #1080

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -26,7 +26,7 @@ builds:
     flags:
       - -trimpath
     main: ./cmd/step-ca/main.go
-    binary: bin/step-ca
+    binary: step-ca
     ldflags:
       - -w -X main.Version={{.Version}} -X main.BuildTime={{.Date}}
   -
@@ -47,7 +47,7 @@ builds:
     flags:
       - -trimpath
     main: ./cmd/step-cloudkms-init/main.go
-    binary: bin/step-cloudkms-init
+    binary: step-cloudkms-init
     ldflags:
       - -w -X main.Version={{.Version}} -X main.BuildTime={{.Date}}
   -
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -47,7 +47,7 @@ builds:
     flags:
       - -trimpath
     main: ./cmd/step-cloudkms-init/main.go
-    binary: bin/step-cloudkms-init
+    binary: step-cloudkms-init
```
maraino, line 50: These other binaries are now deprecated, we will not remove the code until the next release, but we can probably remove them from the packages.

.goreleaser.yml
```diff
@@ -68,7 +68,7 @@ builds:
     flags:
       - -trimpath
     main: ./cmd/step-awskms-init/main.go
-    binary: bin/step-awskms-init
+    binary: step-awskms-init
```
maraino, line 71: Deprecated

.goreleaser.yml
```diff
@@ -85,6 +85,40 @@ archives:
...
+  #
+  # Useful tools for debugging .debs:
+  # List file contents: dpkg -c dist/step_...deb
+  # Package metadata: dpkg --info dist/step_....deb
+  #
+  -
+    builds:
+      - step-ca
+      - step-awskms-init
+      - step-cloudkms-init
```
maraino, line 104: deprecated

maraino (changes requested):
I would remove deprecated binaries from the packages.

maraino (approved):

PR #1087, opened 2022-10-03
Set Challenge ValidateAt

maraino:
### Description

This PR sets ValidatedAt when a challenge status is set to invalid.


dopey (approved):
Fine. But I hate it.

PR #1089, opened 2022-10-03
Revert ValidatedAt

maraino:
### Description

This PR reverts #1087 and adds a missing case in the acme errors switch.

### DIFF
--- a/acme/errors.go
+++ b/acme/errors.go
@@ -75,6 +75,8 @@ func (ap ProblemType) String() string {
 		return "accountDoesNotExist"
 	case ErrorAlreadyRevokedType:
 		return "alreadyRevoked"
+	case ErrorBadAttestationStatementType:
+		return "badAttestationStatement"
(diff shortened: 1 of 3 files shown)


acme/errors.go
```diff
@@ -75,6 +75,8 @@ func (ap ProblemType) String() string {
 		return "accountDoesNotExist"
 	case ErrorAlreadyRevokedType:
 		return "alreadyRevoked"
+	case ErrorBadAttestationStatementType:
+		return "badAttestationStatement"
```
hslatman, line 79: This list of errors is based on the list at https://datatracker.ietf.org/doc/html/rfc8555#section-6.7. If we add this one, it probably also needs to end up in the RFC.

maraino: https://github.com/brandonweeks/draft-bweeks-acme-device-attest/issues/5

dopey (approved):

PR #1096, opened 2022-10-06
Attestation data

maraino:
### Description

This PR adds support for sending the attestation data on a linkedca and storing it in the database if it gets implemented.


dopey (approved):
lgtm

PR #1102, opened 2022-10-11
Add test simulating YubiKey v5.2.4

maraino:
### Description

There are YubiKeys v5.2.4 where the attestation intermediate (f9) does not have a basic constraint extension, so that certificate is not marked as a CA. 

The test and CA in this commit imitate that use case. Currently, the test case returns an error as we don't support it. But if we change the verification to support this use case, the test should change accordingly.

See https://github.com/Yubico/yubikey-manager/issues/522

### DIFF
--- a/acme/challenge_test.go
+++ b/acme/challenge_test.go
@@ -2780,9 +2780,9 @@ func Test_doStepAttestationFormat_noCAIntermediate(t *testing.T) {
 
 	// This CA simulates a YubiKey v5.2.4, where the attestation intermediate in
 	// the CA does not have the basic constraint extension. With the current
-	// validation of the certificate the test case bellow returns an error, if
-	// we change the validation to support this use case, the test case bellow
-	// should change..
+	// validation of the certificate the test case below returns an error. If
+	// we change the validation to support this use case, the test case below
+	// should change.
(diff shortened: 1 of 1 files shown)


acme/challenge_test.go
```diff
@@ -2774,3 +2774,97 @@ func Test_doStepAttestationFormat(t *testing.T) {
...
 }
+
+func Test_doStepAttestationFormat_noCAIntermediate(t *testing.T) {
+	ctx := context.Background()
+
+	// This CA simulates a YubiKey v5.2.4, where the attestation intermediate in
+	// the CA does not have the basic constraint extension. With the current
+	// validation of the certificate the test case bellow returns an error, if
+	// we change the validation to support this use case, the test case bellow
+	// should change..
```
hslatman, line 2785:
```suggestion
	// This CA simulates a YubiKey v5.2.4, where the attestation intermediate in
	// the CA does not have the basic constraint extension. With the current
	// validation of the certificate the test case below returns an error. If
	// we change the validation to support this use case, the test case below
	// should change..
```

hslatman (dismissed):

hslatman (approved):

PR #1104, opened 2022-10-12
Fix err assert linter warnings - upgrade outdated package

dopey:


authority/provisioner/sshpop_test.go
```diff
@@ -370,8 +372,10 @@ func TestSSHPOP_AuthorizeSSHRenew(t *testing.T) {
 			tc := tt(t)
 			if cert, err := tc.p.AuthorizeSSHRenew(context.Background(), tc.token); err != nil {
 				if assert.NotNil(t, tc.err) {
-					sc, ok := err.(render.StatusCodedError)
-					assert.Fatal(t, ok, "error does not implement StatusCodedError interface")
+					var sc render.StatusCodedError
+					if assert.True(t, errors.As(err, &sc)) {
+						assert.Equals(t, sc.StatusCode(), tc.code)
+					}
 					assert.Equals(t, sc.StatusCode(), tc.code)
```
hslatman, line 379: There are a couple of occasions of `assert.Equals(t, sc.StatusCode(), tc.code)` that seem to be duplicated in the `if` statement just before. I assume the ones outside the `if` can be removed.

errs/error.go
```diff
@@ -268,15 +270,11 @@ func NewError(status int, err error, format string, args ...interface{}) error {
...
+		var ste render.StatusCodedError
+		if errors.As(err, &ste) {
+			e = &Error{Status: ste.StatusCode(), Err: err}
 		} else {
-			cause := errors.Cause(err)
-			if sc, ok := cause.(render.StatusCodedError); ok {
-				e = &Error{Status: sc.StatusCode(), Err: err}
-			} else {
-				e = &Error{Status: status, Err: err}
-			}
```
hslatman, line 279: Kinda curious if there are cases that are now not handled like a `StatusCodedError` after the change. We may not have been testing those, or they just don't exist. Getting rid of the `.Cause` has my vote, though.

ca/client_test.go
```diff
@@ -1118,9 +1125,10 @@ func TestClient_SSHBastion(t *testing.T) {
 					t.Errorf("Client.SSHBastion() = %v, want nil", got)
 				}
 				if tt.responseCode != 200 {
-					sc, ok := err.(render.StatusCodedError)
-					assert.Fatal(t, ok, "error does not implement StatusCodedError interface")
-					assert.Equals(t, sc.StatusCode(), tt.responseCode)
+					var sc render.StatusCodedError
+					if assert.True(t, errors.As(err, &sc)) {
```
hslatman, line 1129: In some places the `assert.True()` was provided with the `"error does not implement StatusCodedError interface"` msg. For consistency I think it's good to have them in all.

hslatman (dismissed):
Small remark about the test error message, otherwise OK

hslatman (approved):

PR #1136, opened 2022-10-24
Add additional properties to `meta` object in ACME directory response

hslatman:
Hardware appliances from [Kemp](https://kemptechnologies.com/) seem to validate the contents of the `meta` object, even if none of the properties in the `meta` object is set. According to the [RFC](https://datatracker.ietf.org/doc/html/rfc8555#section-7.1.1), the `meta` object and all its properties are optional, so technically this should be fixed by the manufacturer.

This PR adds additional ACME provisioner configuration options that are added to the ACME directory response when set. With just the `termsOfService` configured, the Kemp hardware accepted the ACME directory response. This was tested with https://github.com/smallstep/certificates/releases/tag/v0.22.2-rc14. 

Currently only config based provisioners are supported. No validation of the contents of the input strings is performed. Will add support for linkedca etc. soon. It's probably good to not have this PR open for too long, so that this can be used with an official release. Additional changes can be made in follow up PRs if necessary.


hslatman:
I'll probably create an extended version of this that adds configuration options for setting the other ACME `meta` properties for completeness. Making the `meta` object optional is already in line with the spec and ACME clients don't have a problem with that. In most cases they also don't seem to care about the `meta` object being empty, except for the Kemp ACME client, which tries to find the `termsOfService` property, which we currently don't set. This is an optional property, so our implementation is in fact correct already, but we can make it work with Kemp hardware if the other properties also exist.

hslatman:
Waiting for https://github.com/smallstep/linkedca/pull/39 to be merged.

dopey (approved):

PR #1139, opened 2022-10-24
Upgrade pkcs7 to the latest patches branch

maraino:
### Description

This PR upgrades the pkcs7 package to the latest version in smallstep/pkcs7@patches. This includes now support for generic Decrypter methods so that KMS can be used instead of a key in disk with SCEP.

Related PR:
- [x] https://github.com/smallstep/pkcs7/pull/3


fuhry:
> instead of a key in disk with SCIM

I think this is a typo and should say SCEP. Otherwise LGTM, thanks!

hslatman (approved):

PR #1142, opened 2022-10-26
[action] keyless cosign for all release artifacts

dopey:


.goreleaser.yml
```diff
@@ -154,8 +155,8 @@ release:
 
     ```
     cosign verify-blob \
-      -key https://raw.githubusercontent.com/smallstep/certificates/master/cosign.pub \
-      -signature ~/Downloads/step-ca_darwin_{{ .Version }}_amd64.tar.gz.sig
+      --certificate ~/Downloads/step-ca_darwin_{{ .Version }}_amd64.tar.gz.sig.pem \
+      --signature ~/Downloads/step-ca_darwin_{{ .Version }}_amd64.tar.gz.sig \
```
hslatman, line 159: Have you tried doing the `verify-blob` without the `--certificate` too? I think that should also work and should say something like 

```console
$ COSIGN_EXPERIMENTAL=1 cosign verify-blob --signature some-bin.sig some-bin
tlog entry verified with uuid: "2b7ca2bfb7ee09114a15d60761c2ade8c97f07cc20c02e635a92ba137a08a6de" index: 1234567
Verified OK
```

This depends on the Rekor server to be available (as does the signing part), so there's a potential failure there. We could add both methods for completeness?

hslatman (commented):
Woohoooo 🎉

make/docker.mk
```diff
@@ -54,7 +54,7 @@ define DOCKER_BUILDX
 	# $(1) -- Image Tag
 	# $(2) -- Push (empty is no push | --push will push to dockerhub)
 	docker buildx build . --progress plain -t $(DOCKER_IMAGE_NAME):$(1) -f docker/Dockerfile.step-ca --platform="$(DOCKER_PLATFORMS)" $(2)
-	echo -n "$(COSIGN_PWD)" | cosign sign -key /tmp/cosign.key -r $(DOCKER_IMAGE_NAME):$(1)
+	cosign sign -r $(DOCKER_IMAGE_NAME):$(1)
```
maraino, line 57: This looks like that will also sign the image with the `:latest`. Cosign help mentions that we should not do this:

```
Make sure to sign the image by its digest (@sha256:...) rather than by tag
(:latest) so that you actually sign what you think you're signing! This prevents
race conditions or (worse) malicious tampering.
```

maraino: It looks like it's recommended for all tags, but it might be specially bad for `:latest`

tashian:
love this

.goreleaser.yml
dopey: I just tried this locally and it didn't work.

```
 max@Maxs-MacBook-Pro  ~/src/github.com/smallstep/gitops   main  COSIGN_EXPERIMENTAL=1 cosign verify-blob --signature ~/Downloads/checksums.txt.sig ~/Downloads/checksums.txt
WARNING: No valid entries were found in rekor to verify this blob.

Transparency log support for blobs is experimental, and occasionally an entry isn't found even if one exists.

We recommend requesting the certificate/signature from the original signer of this blob and manually verifying with cosign verify-blob --cert [cert] --signature [signature].
Error: verifying blob [/Users/max/Downloads/checksums.txt]: could not find a valid tlog entry for provided blob, found 1 invalid entries
main.go:62: error during command execution: verifying blob [/Users/max/Downloads/checksums.txt]: could not find a valid tlog entry for provided blob, found 1 invalid entries

⬆ [ exit: 1 ] [ time: 10/26:16:18:18 ]
 max@Maxs-MacBook-Pro  ~/src/github.com/smallstep/gitops   main  COSIGN_EXPERIMENTAL=1 cosign verify-blob --signature ~/Downloads/checksums.txt.sig --certificate ~/Downloads/checksums.txt.pem  ~/Downloads/checksums.txt
tlog entry verified with uuid: bab7cf6963f44c45f617ec8dec222e8171dc741d95e46a66663ac9e0c6e9b7c5 index: 5881305
Verified OK
```

dopey: FWIW this does work with verifying signatures on containers. `COSIGN_EXPERIMENTAL=1 cosign verify smallstep/step-sds:0.2.2-rc13` just works.

hslatman: I guess you hit one of the exceptional cases they mention in the error. Probably good to keep the `--certificate` as the main way to verify blobs now to not confuse people 🙂

maraino (approved):

PR #1156, opened 2022-11-04
Add support for renew when using stepcas

### LINKED ISSUE #1021: Add renewal support using RAs
### Description

RAs (stepcas, ~~vaultcas~~, ~~cloudcas~~) can now support renewals if the cli sends the authentication token instead. We should implement the `RenewCertificate` function on those three implementations.

maraino:
It supports renewing X.509 certificates when an RA is configured with stepcas. This will only work when the renewal uses a token, and it won't work with mTLS.

The audience cannot be properly verified when an RA is used, to avoid this we will get from the database if an RA was used to issue the initial certificate and we will accept the renew token.

Fixes #1021 for stepcas.

The support for cloudcas was already implemented, and the renew token cannot be used for vaultcas because the CSR is [required](https://developer.hashicorp.com/vault/api-docs/secret/pki#sign-certificate) and it's not available in a renew flow.

### DIFF
--- a/cas/apiv1/services.go
+++ b/cas/apiv1/services.go
@@ -83,3 +83,23 @@ func (e NotImplementedError) Error() string {
 func (e NotImplementedError) StatusCode() int {
 	return http.StatusNotImplemented
 }
+
+// ValidationError is the type of error returned if request is not properly
+// validated.
+type ValidationError struct {
+	Message string
+}
+
+// NotImplementedError implements the error interface.
+func (e ValidationError) Error() string {
+	if e.Message != "" {
+		return e.Message
+	}
+	return "bad request"
+}
+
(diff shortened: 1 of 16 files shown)


cas/apiv1/services.go
```diff
@@ -83,3 +83,23 @@ func (e NotImplementedError) Error() string {
...
 	return http.StatusNotImplemented
 }
+
+// ValidationError is the type of error returned if request is not properly
+// validated.
+type ValidationError struct {
+	Message string
+}
+
+// NotImplementedError implements the error interface.
```
dopey, line 93: bad comment

dopey (approved):

cas/apiv1/services.go
maraino: Fixed by https://github.com/smallstep/certificates/pull/1164

PR #1166, opened 2022-11-08
Remove deprecated docs directory

dopey:


tashian (approved):
This is great and I love that you moved contributing doc to the top-level. I will update the refs to that doc, which is the only one I think we still refer to anywhere else.

Issue #1171, opened 2022-11-09
[Bug?]: ACME certificate response has extraneous `Content-Type` parameter


hslatman:
### Steps to Reproduce

1. Create a `step-ca` instance with an `ACME` provisioner.
2. Setup a [ejabberd](https://github.com/processone/ejabberd) instance, configured with the `step-ca` ACME provisioner as the ACME server.
3. Watch it fail with an error similar to the one in https://github.com/processone/ejabberd/issues/3938

### Your Environment

Not relevant

### Expected Behavior

The certificate to be issued

### Actual Behavior

A failure is reported:

`2022-11-07 08:30:28.609162+00:00 [error] Failed to request certificate for ejabberd.example.com: HTTP error: unexpected content type: application/pem-certificate-chain; charset=utf-8`

### Additional Context

The Erlang ACME client in use by https://github.com/processone/ejabberd checks the resulting `Content-Type` to be equal to `application/pem-certificate-chain`. `step-ca` returns `application/pem-certificate-chain; charset=utf-8`. In general, the `charset` is optional, so it doesn't seem wrong to return it. The RFC isn't entirely conclusive about the additional parameters, though, which leaves room for interpretation and implementations to have some variations. It does lean towards not including the `charset`, imo.

We know many ACME clients work with `step-ca`, so they accept the `charset=utf-8` parameter. One concrete example is certbot, which added support for this: https://github.com/certbot/certbot/blob/f73e062c7a3b4856bffe2a7c54ece64c2f1af265/certbot/CHANGELOG.md#changed-29 in 2020, probably due to another (public) ACME server to include it in its responses. 

We could wait for a patch in the Erlang library, or we can simply remove the parameter from being returned. Removing it is most probably not going to cause issues with existing ACME clients, as they have probably been tested with Let's Encrypt, which I assume doesn't return the `charset`. Would need to perform a quick sanity check, though.

### Contributing

Vote on this issue by adding a 👍 reaction.
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

azazeal:
Is there a specific `User Agent` header value that `ejabberd` requests carry, @hslatman?

hslatman:
@azazeal I haven't tested it with `ejabberd` (yet), nor with the ACME client library (https://github.com/processone/p1_acme), so I'm not entirely sure. By default `inets`, the default HTTP client that the ACME client uses, doesn't seem to set it (ignore the 404):

```console
3> inets:start().
[...]

PR #1172, opened 2022-11-09
Remove `charset=utf-8` from ACME certificate responses

hslatman:
Addresses https://github.com/smallstep/certificates/issues/1171. 

In internal discussions the fact that some information is lost from the response when `charset` is removed came up. Ideally, we think the Erlang client in case should be updated to be more lenient in what it accepts, in line with the _robustness principle_. The change is small, so maybe it gets done. On the other hand, the change here is small too, so we may merge it. 

Awaiting approval and some compatibility testing.

### DIFF
--- a/acme/api/handler.go
+++ b/acme/api/handler.go
@@ -394,6 +394,6 @@ func GetCertificate(w http.ResponseWriter, r *http.Request) {
 	}
 
 	api.LogCertificate(w, cert.Leaf)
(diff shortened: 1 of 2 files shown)


hslatman:
Tested the change with `certbot`, [`acmez`](https://github.com/mholt/acmez) (the library powering `CertMagic` and thus Caddy) and the `step` CLI. All retrieved the certificate OK.

dopey (approved):

PR #1192, opened 2022-11-26
Add capabilities endpoint and client integration

dopey:


api/api.go
```diff
@@ -261,8 +266,10 @@ func New(auth Authority) RouterHandler {
 	return &caHandler{}
 }
 
+// Route defines routing for the API.
 func Route(r Router) {
 	r.MethodFunc("GET", "/version", Version)
+	r.MethodFunc("GET", "/capabilities", Capabilities)
```
maraino, line 272: Is this the path of an endpoint commonly used on other services?

maraino (commented):
Where is the authority implementation?

ca/client_test.go
```diff
@@ -206,6 +206,64 @@ func TestClient_Version(t *testing.T) {
...
+					t.Errorf("Client.Capabilities() = %v, want nil", got)
+				}
+				if tt.responseCode == http.StatusNotFound {
+					assert.True(t, errors.Is(err, ErrNotFound))
+				} else {
+					assert.HasPrefix(t, tt.expectedErr.Error(), err.Error())
+				}
+			default:
+				if !reflect.DeepEqual(got, tt.response) {
+					t.Errorf("Client.Version() = %v, want %v", got, tt.response)
```
hslatman, line 260:
```suggestion
					t.Errorf("Client.Capabilities() = %v, want %v", got, tt.response)
```

ca/client.go
```diff
@@ -605,6 +608,34 @@ retry:
 	return &version, nil
 }
 
+// Capabilities performs the capabilities request to the CA and returns the
+// api.Capabilities struct.
```
hslatman, line 612:
```suggestion
// api.CapabilitiesResponse struct.
```

api/api.go
```diff
@@ -211,6 +212,10 @@ type VersionResponse struct {
 	RequireClientAuthentication bool   `json:"requireClientAuthentication,omitempty"`
 }
 
+// CapabilitiesResponse is the response object that returns the version of the
+// server.
+type CapabilitiesResponse authority.Capabilities
```
hslatman, line 217: I think it's safer to make this a new struct instead of an alias. It keeps a cleaner separation between the API and the authority packages and won't "leak" new properties added to `authority.Capabilities` unintentionally.

ca/client.go
```diff
@@ -605,6 +608,34 @@ retry:
...
+func (c *Client) Capabilities() (*api.CapabilitiesResponse, error) {
```
hslatman, line 613: Probably OK to return it as a value instead of a pointer.

ca/client.go
```diff
@@ -605,6 +608,34 @@ retry:
...
+	case resp.StatusCode >= 400:
+		if !retried && c.retryOnError(resp) {
+			retried = true
+			goto retry
+		}
+		return nil, readError(resp.Body)
+	default:
+		var capabilities api.CapabilitiesResponse
+		if err := readJSON(resp.Body, &capabilities); err != nil {
+			return nil, errs.Wrapf(http.StatusInternalServerError, err, "client.Capabilities; error reading %s", u)
```
hslatman, line 633: Personal preference; feel free to ignore.

```suggestion
			return nil, errs.Wrapf(http.StatusInternalServerError, err, "client.Capabilities; error reading %q", u)
```

ca/client.go
dopey: The way you recommend is the way we've done it for other types - like Version I think. So it makes sense to be consistent. But I hate that I'll have to remember to update >1 types. Makes it harder on developers in my opinion. If the general consensus is to go with a new type, then I'm happy to switch.

api/api.go
dopey: The way you recommend is the way we've done it for other types - like Version I think. So it makes sense to be consistent. But I hate that I'll have to remember to update >1 types. Makes it harder on developers in my opinion. If the general consensus is to go with a new type, then I'm happy to switch.

ca/client.go
dopey: Both Health and Version return a pointer so I think it's better to be consistent.

api/api.go
maraino: I'm not sure this is the right word. If I search for capabilities, the results appear to be more related to policies than something like this. Linux, Vault, ...

PR #1207, opened 2022-12-16
Add IsEnabled method in AdminClient for checking admin API availability

dopey:


ca/adminClient.go
```diff
@@ -137,6 +140,26 @@ func (c *AdminClient) retryOnError(r *http.Response) bool {
...
+	defer resp.Body.Close()
+	// If the response is a 404 then the Admin Remote Management API is not
+	// enabled. Therefore we default to using the local ca.json.
+	switch resp.StatusCode {
+	case http.StatusNotFound, http.StatusNotImplemented:
+		return ErrAdminAPINotImplemented
+	case http.StatusUnauthorized:
+		return ErrAdminAPINotAuthorized
+	default:
+		return errors.Errorf("unexpected status code when performing is-enabled check for Admin API: %d", resp.StatusCode)
```
maraino, line 159: All options return an error, I suppose the status unauthorize case should return a nil. 

Instead of using a GET you might try to use a HEAD or even better an [OPTIONS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS) request to get the methods allowed in a specific endpoint.

maraino (changes requested):

maraino (approved):
lgtm

PR #1223, opened 2023-01-10
Docker: Generate and use independent provisioner and private key passwords

tashian:
When using our Docker image, it usually makes sense to enable Remote Provisioner Management.

In the current version, we only generate one password file for the CA when we initialize a PKI via `docker run`, and we pass it to `step ca init` for use with both the CA keys and the admin provisioner.

This change adds a bit more security by also generating a provisioner password file by default when the CA is being initialized. The CA admin provisioner username and password are printed after `step ca init` runs, and the provisioner password file is deleted.

I'm also updating our Docker README setup instructions to enable Remote Management by default.

cc @hslatman

### DIFF
--- a/docker/Dockerfile.step-ca
+++ b/docker/Dockerfile.step-ca
@@ -7,7 +7,6 @@ RUN apk add --no-cache curl git make
 RUN make V=1 download
 RUN make V=1 bin/step-ca bin/step-awskms-init bin/step-cloudkms-init
 
-
 FROM smallstep/step-cli:latest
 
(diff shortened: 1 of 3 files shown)


tashian:
I've also updated the Docker README as part of this. Since that file is not in source control, I'm attaching the updated one here in prep for the next `step-ca` release.

[docker-readme-updated.md](https://github.com/smallstep/certificates/files/10385586/docker-readme-updated.md)

maraino (approved):
lgtm

hslatman:
Nice 😄

Issue #1227, opened 2023-01-11
[Bug]: Starting docker container results in `flag '--admin-subject' is incompatible with '--remote-management'`


FibreFoX:
### Steps to Reproduce

Creating a fresh docker container with version `0.23.1` results in an error being printed
```sh
docker run --rm -it -p 9000:9000 -e "DOCKER_STEPCA_INIT_NAME=Smallstep" -e "DOCKER_STEPCA_INIT_DNS_NAMES=itsabug" smallstep/step-ca:0.23.1 step ca init
```

This works "normal" when using previous version:
```sh
docker run --rm -it -p 9000:9000 -e "DOCKER_STEPCA_INIT_NAME=Smallstep" -e "DOCKER_STEPCA_INIT_DNS_NAMES=itsabug" smallstep/step-ca:0.23.0 step ca init
```

### Your Environment

* OS - Windows 10, Docker-Desktop (no WSL)
* `step-ca` Version - 0.23.1

### Expected Behavior

Behavior like previous `0.23.0` is expected:
```
Generating root certificate... done!
Generating intermediate certificate... done!

✔ Root certificate: /home/step/certs/root_ca.crt
✔ Root private key: /home/step/secrets/root_ca_key
✔ Root fingerprint: 5ddb1d717a84670b3be78bf1e6f1123b95b81f91df95dd5395a03a689756a3aa
✔ Intermediate certificate: /home/step/certs/intermediate_ca.crt
✔ Intermediate private key: /home/step/secrets/intermediate_ca_key
✔ Database folder: /home/step/db
✔ Default configuration: /home/step/config/defaults.json
✔ Certificate Authority configuration: /home/step/config/ca.json

Your PKI is ready to go. To generate certificates for individual services see 'step help ca'.

```

### Actual Behavior

Error message is being printed:
```
flag '--admin-subject' is incompatible with '--remote-management'
```

### Contributing

Vote on this issue by adding a 👍 reaction.
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

tashian:
Thanks @FibreFoX.
I've just updated the Docker docs to reflect 0.23.1 to reflect new flags.
I've tested the quickstart and manual flows that are in the readme, but I will test out your init flow as well and follow up.

tashian:
Looks like there was an underlying issue with boolean handling on these init flags in the entrypoint script.
#1228 has the fix.

hslatman:
This was fixed and will be in the next release

PR #1228, opened 2023-01-11
Only pass --admin-subject if --remote-management is true; fix  boolean handling

tashian:


tashian:
Closes #1227

docker/entrypoint.sh
```diff
@@ -53,18 +52,20 @@ function step_ca_init () {
...
+    if [ "${DOCKER_STEPCA_INIT_SSH}" == "true" ]; then
         setup_args=("${setup_args[@]}" --ssh)
     fi
-    if [ -n "${DOCKER_STEPCA_INIT_ACME}" ]; then
+    if [ "${DOCKER_STEPCA_INIT_ACME}" == "true" ]; then
         setup_args=("${setup_args[@]}" --acme)
     fi
-    if [ -n "${DOCKER_STEPCA_INIT_REMOTE_MANAGEMENT}" ]; then
-        setup_args=("${setup_args[@]}" --remote-management)
+    if [ "${DOCKER_STEPCA_INIT_REMOTE_MANAGEMENT}" == "true" ]; then
```
hslatman, line 61: I'm fine with having to set them to `true`. It might be an option to have it take multiple truthy/falsy values that then get used, but that could be done later.

hslatman (approved):

PR #1235, opened 2023-01-20
Improve validation and error messages for Orders with Permanent Identifier

hslatman:
A CSR with a Subject Common Name that is not equal to the Permanent Identifier will be blocked from signing. Without the equality check, a CSR can be submitted that contains a different name than the Permanent Identifier, which is then not challenged in any way.

In addition to that, this PR also adds the ACME `Subproblem` type to ACME `Error` objects. These can be used to provide a more granular and detailed error message to ACME clients. The additional details have so far only been added to the `step` format validation, but will be included in other ACME Errors too when they work as expected.


hslatman:
@dopey: do you know if the `Subproblems` and `Identifier` have been used in the past? I looked through the current state of the code and didn't find any usage of these properties, so they likely never have been stored in `nosql` and they're pretty safe to delete/change.

acme/db/nosql/challenge.go
```diff
@@ -6,8 +6,10 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"github.com/smallstep/certificates/acme"
+
 	"github.com/smallstep/nosql"
```
dopey, line 10: why are we splitting these up one per line like this?

dopey (approved):

acme/order.go
```diff
@@ -165,6 +165,15 @@ func (o *Order) Finalize(ctx context.Context, db DB, csr *x509.CertificateReques
...
 			permanentIdentifier = o.Identifiers[i].Value
+			// the first (and only) Permanent Identifier that gets added to the certificate
+			// should be equal to the Subject Common Name if it's set. If not equal, the CSR
+			// is rejected, because the Common Name hasn't been challenged in that case. This
+			// could result in unauthorized access if a relying system relies on the Common
+			// Name in its authorization logic.
+			if csr.Subject.CommonName != "" && csr.Subject.CommonName != permanentIdentifier {
+				return NewError(ErrorBadCSRType, "CSR Subject Common Name does not match identifiers exactly: "+
+					"CSR Subject Common Name = %s, Order Permanent Identifier = %s", csr.Subject.CommonName, permanentIdentifier)
+			}
```
maraino, line 176: **TL;DR**: should we check the SAN as we do with the rest?
---
According to [RFC-8555#section-7.4](https://www.rfc-editor.org/rfc/rfc8555#section-7.4)
```
The CSR encodes the client's requests with regard to the content of
the certificate to be issued.  The CSR MUST indicate the exact same
set of requested identifiers as the initial newOrder request.
Identifiers of type "dns" MUST appear either in the commonName
portion of the requested subject name or in an extensionRequest
attribute [[RFC2985](https://www.rfc-editor.org/rfc/rfc2985)] requesting a subjectAltName extension, or both.
(These identifiers may appear in any sort order.)  Specifications
that define new identifier types must specify where in the
certificate signing request these identifiers can appear.
```

First looking at the code it looks like we enforce DNS to be in a SAN, it seems that we do not support when the DNS is only in the commonName. We can decide later if we should fix this or not.

The last part says `Specifications that define new identifier types must specify where in the certificate signing request these identifiers can appear`. Looking at Brandon's [draft](https://datatracker.ietf.org/doc/html/draft-bweeks-acme-device-attest-01#section-3), the only thing that it says is:

```
Clients MAY include this identifier in the certificate signing request (CSR). 
Alternatively if the server wishes to only issue privacy-preserving certificates, 
it MAY reject CSRs containing a PermanentIdentifier in the subjectAltName extension.
```

It doesn't really force you to do something specific. Looking at just the common name might be a good compromise, especially because it's a little hard to look at the permanent identifier SAN. But for the rest of the identifiers, we're looking at the SANs. 

Another potential problem is that the max length of the commonName is `64` bytes. So if the permanent identifier is larger than this, the check might fail. I don't think we're going to encounter this issue though.

Looking at Boulder, it will validate always the CommonName if it is available, if it's not it will be set to the first DNS in the CSR with a length lower or equal to 64 bytes. And it will fail if there are none.

For TLS, because the CommonName was a valid FQDN the limit was generally 64 bytes, but the CN is deprecated, and I can't find any reference to a limit other than that 🤷‍♂️

acme/db/nosql/challenge.go
hslatman: The split is for imports from the stdlib, third parties, 1st party and local imports. So far I've been doing it when I change a file with imports. I'd like to automate it sometime using https://github.com/incu6us/goimports-reviser (using their `company` options)

acme/order.go
hslatman: > First looking at the code it looks like we enforce DNS to be in a SAN, it seems that we do not support when the DNS is only in the commonName. We can decide later if we should fix this or not.

Yes, that's a bit of a compromise. We currently change the provided CSR to include either the IP or DNS in the SANs when it's provided in the Common Name. I did think about changing that as part of this PR, but I decided to postpone that to a different PR specific to that improvement. I think we should perform (some parts of) the validation on a copy of the CSR (or on a derived data struct), so that the original contents and ordering can be maintained. 

> The last part says Specifications that define new identifier types must specify where in the certificate signing request these identifiers can appear. Looking at Brandon's [draft](https://datatracker.ietf.org/doc/html/draft-bweeks-acme-device-attest-01#section-3), the only thing that it says is:

> Clients MAY include this identifier in the certificate signing request (CSR). 
Alternatively if the server wishes to only issue privacy-preserving certificates, 
it MAY reject CSRs containing a PermanentIdentifier in the subjectAltName extension.

> It doesn't really force you to do something specific. Looking at just the common name might be a good compromise, especially because it's a little hard to look at the permanent identifier SAN. But for the rest of the identifiers, we're looking at the SANs.

The RFC mentions this about the Permanent Identifier: 

```
The identity along with the assigning organization can be included in the 
Subject Alternate Name Extension using the PermanentIdentifier form described in RFC4043
```

This doesn't mention the Subject. Our CLI currently puts the value in the CSR CN too, though, which is the default behavior when providing a subject for the certificate, but may be something to change for Permanent Identifiers. Ideally I would've performed the validation similarly to the IPs and DNSs, but it's not always possible to distinguish a Permanent Identifier in the CN from a DNS value. I think the logic can possibly be changed to check that all values in the ACME Order correspond to a value from one of the SANs and/or the Common Name, relying less on the distinction between DNS and Permanent Identifier. If we then also keep enforcing that a Permanent Identifier cannot co-exist with DNS SANs, I think we've got the logic we need.

> Another potential problem is that the max length of the commonName is 64 bytes. So if the permanent identifier is larger than this, the check might fail. I don't think we're going to encounter this issue though.

Currently we only check equivalence when the value in the CN is not empty. A user can submit a CSR without the CN as a workaround. We could provide a flag like `--exclude-common-name` to make its value empty in the generated CSR a bit more easily?

hslatman:
Will briefly test this with our hosted option before merging.

acme/order.go
maraino: > @maraino: OK with changing how validation works in a follow-up PR? I think it makes sense to revisit it sometime soon, for example with the tpm format and for HardwareModuleName support.

sounds good

maraino (approved):

PR #1252, opened 2023-01-31
Helpful message on CA startup when config can't be opened

### LINKED ISSUE #1248: When running `step-ca` when no config file is present, the output could be more helpful.
See https://github.com/microsoft/winget-pkgs/pull/95092#issuecomment-1409575711

tashian:
Closes #1248

### DIFF
--- a/commands/app.go
+++ b/commands/app.go
@@ -141,6 +141,9 @@ func appAction(ctx *cli.Context) error {
 
 	cfg, err := config.LoadConfiguration(configFile)
 	if err != nil && token == "" {
+		fmt.Println("step-ca can't find or open the configuration file for your CA.")
+		fmt.Println("You may need to create a CA first by running `step ca init`.")
+		fmt.Println("Documentation: https://u.step.sm/docs/ca\n")
 		fatal(err)
 	}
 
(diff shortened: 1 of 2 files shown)


maraino (changes requested):
This is probably not the right place to put this. If you look at https://github.com/smallstep/certificates/blob/40538d8224a2b2ef897d4feb8997ecd93f932e6e/commands/app.go#L149-L152

This is expected to fail if the server is using the automatic RA configuration, without a ca.json.

commands/app.go
```diff
@@ -141,6 +141,10 @@ func appAction(ctx *cli.Context) error {
 
 	cfg, err := config.LoadConfiguration(configFile)
 	if err != nil && token == "" {
+		fmt.Println("step-ca can't find or open the configuration file for your CA.")
+		fmt.Println("You may need to create a CA first by running `step ca init`.")
+		fmt.Println("Documentation: https://u.step.sm/docs/ca")
+		fmt.Println("")
```
maraino, line 147: It might be better to print only this if the file does not exists or cannot be opened for any reason:
```suggestion
		var pathErr *os.PathError
		if errors.As(err, &pathErr) {
			fmt.Println("step-ca can't find or open the configuration file for your CA.")
			fmt.Println("You may need to create a CA first by running `step ca init`.")
			fmt.Println("Documentation: https://u.step.sm/docs/ca")
			os.Exit(1)
		}
```

dopey (approved):

maraino (approved):
lgtm

PR #1255, opened 2023-02-03
There is an error during RA installation which shows.

zyzyx03:
"install: cannot stat 'step-ca_0.23.2/bin/step-ca': No such file or directory" Upon checking the is no bin directory after step-ca_linux_0.23.2_amd64.tar.gz is extracted so by simply changing from step-ca_${CA_VERSION:1}/bin/step-ca to step-ca_${CA_VERSION:1}/step-ca the installation succeed.

### DIFF
--- a/scripts/install-step-ra.sh
+++ b/scripts/install-step-ra.sh
@@ -188,7 +188,7 @@ CA_VERSION=$(curl -s https://api.github.com/repos/smallstep/certificates/release
 
 curl -sLO https://github.com/smallstep/certificates/releases/download/$CA_VERSION/step-ca_linux_${CA_VERSION:1}_$arch.tar.gz
 tar -xf step-ca_linux_${CA_VERSION:1}_$arch.tar.gz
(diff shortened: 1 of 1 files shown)


tashian (approved):
lgtm, thank you for the PR!

tashian:
@dopey It looks like this changed between step-ca 0.22.0 and 0.23.0, perhaps with the removal of step-*-init binaries?

I will look through some other places to see if this broke anything else:
- [ ] docs
- [ ] certificates repo
- [ ] cli repo

Any other spots that may need changes?

dopey:
@tashian wish I could say for sure. Seems like this change could be responsible - https://github.com/smallstep/certificates/commit/97cc09e26cc776e14dd777b422ab36ca68d01727. Maybe, as you said, by removing the deprecated binaries goreleaser no longer needed to create a bin dir since there was only one binary 🤷 .

dopey (approved):

zyzyx03:
TQVM everyone, I really appreciate it.

PR #1261, opened 2023-02-08
Clarify bootstrap error when root certificate is not found

tashian:
New output:

```
bin/step ca bootstrap --ca-url https://ca:4443 --fingerprint abc123
The request could not be completed: a root certificate with that fingerprint was not found.
```

### DIFF
--- a/ca/client.go
+++ b/ca/client.go
@@ -671,6 +671,10 @@ retry:
 	if err != nil {
 		return nil, clientError(err)
 	}
+	if resp.StatusCode == 404 {
+		defer resp.Body.Close()
+		return nil, errs.BadRequest("a root certificate with that fingerprint was not found")
+	}
 	if resp.StatusCode >= 400 {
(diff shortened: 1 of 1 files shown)


ca/client.go
```diff
@@ -671,6 +671,10 @@ retry:
 	if err != nil {
 		return nil, clientError(err)
 	}
+	if resp.StatusCode == 404 {
+		defer resp.Body.Close()
```
tashian, line 675: I'm not sure if this was actually needed here.

maraino (changes requested):
I don't think this is the right place to do it.

We should manage this in the CLI. And in the API side of certificates if we want a better message, currently it returns:

```console
$ curl -s https://localhost:9000/root/foo | jq
{
  "status": 404,
  "message": "The requested resource could not be found. Please see the certificate authority logs for more info."
}
```

PR #1265, opened 2023-02-10
Verify CSR key fingerprint with attestation certificate key

maraino:
### Description

This commit ensures that the attestation certificate key matches the key used on the CSR on an ACME device attestation flow.

### DIFF
--- a/acme/authorization.go
+++ b/acme/authorization.go
@@ -8,15 +8,16 @@ import (
 
 // Authorization representst an ACME Authorization.
 type Authorization struct {
-	ID         string       `json:"-"`
-	AccountID  string       `json:"-"`
-	Token      string       `json:"-"`
-	Identifier Identifier   `json:"identifier"`
-	Status     Status       `json:"status"`
-	Challenges []*Challenge `json:"challenges"`
-	Wildcard   bool         `json:"wildcard"`
-	ExpiresAt  time.Time    `json:"expires"`
-	Error      *Error       `json:"error,omitempty"`
+	ID          string       `json:"-"`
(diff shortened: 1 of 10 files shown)


acme/authorization.go
```diff
@@ -8,15 +8,16 @@ import (
...
-	Error      *Error       `json:"error,omitempty"`
+	ID          string       `json:"-"`
+	AccountID   string       `json:"-"`
+	Token       string       `json:"-"`
+	Identifier  Identifier   `json:"identifier"`
+	Status      Status       `json:"status"`
+	Challenges  []*Challenge `json:"challenges"`
+	Wildcard    bool         `json:"wildcard"`
+	ExpiresAt   time.Time    `json:"expires"`
+	Fingerprint string       `json:"fingerprint,omitempty"`
```
hslatman, line 19: Remove the `json` tag. The `Authorization` is used in a JSON response and this property must not be sent to a client.

hslatman (commented):
It looks good. 

I still feel very strongly about (temporarily) storing either the extracted data (e.g. the apple and step data structs) or even the entire `attObj`, so that it can be used a whole. This would allow comparing data with the CSR, checking additional properties, setting additional extensions (if there's a need for that), etc. Your changes do the job for this particular case, but I think there'll be more in the (near) future that can be implemented having the attestation cert or the data we extracted earlier.

acme/authorization.go
maraino: Fixed with 5ff0dde8

hslatman (approved):

PR #1291, opened 2023-02-23
api/render: use proper indentation when rendering JSON

azazeal:
api/render: This PR amends `JSONStatus` so that it uses indentation when rendering JSON.


azazeal:
Nevermind :rofl:

PR #1300, opened 2023-03-03
Fix RSA-OAEP decryption for SCEP

hslatman:
With https://github.com/smallstep/pkcs7/pull/4 the PKCS7 package now supports RSA-OAEP decryption. This type of encryption is used with SCEP.


maraino (approved):
lgtm

PR #1341, opened 2023-04-08
Remove check that requires acme account kid to exact match current NewAccount URI

### LINKED ISSUE #623: Changing ACME server host or port causes existing accounts to craft malformed requests
### Subject of the issue

**Please note, I am not sure if this is a bug.** It is just unexpected behavior I found after wasting a little of Max Furman's time (sorry about that) troubleshooting #193.

Basically, the `kid` used with JWS contain the scheme, host, port, and path (i.e. `https://localhost:9000/acme/myca/account/...`) from the ACME server's configuration at the time.

However, if the ACME server changes any part of the address, then the kid prefix check fails here: https://github.com/smallstep/certificates/blob/6476eb45a7e97ac43a37da7ab4893555625de249/acme/api/middleware.go#L324-L331

### Steps to reproduce
1. Create an ACME account with the server at one URL
2. Move the ACME server to a new URL (any of scheme, host, port, or path).
3. Try another ACME transaction: `urn:ietf:params:acme:error:malformed - The request message was malformed`

### Expected behaviour
I am not sure. Maybe that error is expected.

### Actual behaviour
`urn:ietf:params:acme:error:malformed - The request message was malformed`

### Additional context
I do think this is unexpected, especially with internal/development ACME servers, since in my case I originally had the account created at `localhost:9000` then set up a reverse proxy for the ACME server at `localhost:9001` and started pointing the client to use that one, but reuse the same account. Since the client stores the account `kid` from when it created the account, it retained the `localhost:9000` value.

Essentially, this means the ACME server can never change its address after it is being used, unless new accounts are created _or_ clients change their stored account locations to reflect the new server address. ~~Maybe that is the best way to handle it.~~ Edit: RFC 8555 suggests otherwise, see discussion below.

I am not familar enough with JWK/JWS to know the value of the prefix check, but I assume it wasn't invented for nothing. So this might very well be on the client to fix.

(Incidentally, setting STEPDEBUG=1 was not enough to emit detailed error messages. I had to add my own debug logging to find this.)

Again, I am not sure if this is a server bug. Wanted to open a discussion anyway.

dopey:
Fixes #623.

This is the relevant section from the current ACME Spec (https://datatracker.ietf.org/doc/html/rfc8555#section-6.2): 

```
...
For all other requests, the request is signed using an existing
account, and there MUST be a "kid" field.  This field MUST contain
the account URL received by POSTing to the newAccount resource.
```

I _think_ I interpreted this too strictly, due to the use of the term `MUST`. But, I think rather than the server verifying that this data matches exactly, this line is more a guidance to clients - "the `kid` _MUST_ be of this form (so that the server understands how to parse the accountID)". If that's the case, then we can remove the check all together. Maybe it's just cuz it's late on a Friday, but I don't see how this check is adding any security value anyway. The server grabs the public key using the accountID from the `kid` and verifies the signature on the request - as long as that checks out, I don't see any value in the `kid` being an exact match with the NewAccount URI. @maraino @hslatman @mholt let me know if I'm missing something.

### DIFF
--- a/acme/api/middleware.go
+++ b/acme/api/middleware.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"path"
 	"strings"
 
 	"go.step.sm/crypto/jose"
@@ -293,7 +294,6 @@ func lookupJWK(next nextHTTP) nextHTTP {
 	return func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 		db := acme.MustDatabaseFromContext(ctx)
-		linker := acme.MustLinkerFromContext(ctx)
 
 		jws, err := jwsFromContext(ctx)
 		if err != nil {
@@ -301,16 +301,7 @@ func lookupJWK(next nextHTTP) nextHTTP {
 			return
 		}
 
-		kidPrefix := linker.GetLink(ctx, acme.AccountLinkType, "")
-		kid := jws.Signatures[0].Protected.KeyID
-		if !strings.HasPrefix(kid, kidPrefix) {
-			render.Error(w, acme.NewError(acme.ErrorMalformedType,
-				"kid does not have required prefix; expected %s, but got %s",
-				kidPrefix, kid))
-			return
-		}
-
-		accID := strings.TrimPrefix(kid, kidPrefix)
+		accID := path.Base(jws.Signatures[0].Protected.KeyID)
 		acc, err := db.GetAccount(ctx, accID)
 		switch {
 		case nosql.IsErrNotFound(err):


mholt:
Oh neat!! I'll look at this when I get back from vacation.

maraino:
For the reasons mentioned above and in the issue, I don't think we need to perform any checks on the host path on the URL.

One thing that we need to consider is that `step-ca` has a distinct quality that no other ACME client or server has or expects, it supports multiple ACME provisioners. Both `certbot` and `lego` creates a new account per `domain[:port]` combination, so you will have an account for `localhost` if you use 443 and another one for `localhost:9000`. They don't care about the path of the directory or account other than the one they store in the account.

The current check also prevents one account from being used by multiple provisioners. Is this something that we want?

I think that question is the main one we need to respond to decide which check we want to do.

hslatman:
Responding to @maraino's question: I think we should prevent one ACME account from being used with different provisioners. That means we should at least keep checking that the account being used "belongs to" the provisioner that's being accessed. With the current state of things, that would mean to keep checking at least the path part of the URL.

The reason for this is that there might be different settings active on the different ACME provisioners. ACME accounts enrolled with provisioner A may not be allowed to use settings/templates for provisioner B. A concrete example would be when you have EAB enabled on A only, and not on B: an ACME client could create an account with provisioner B, and then continue using provisioner A (which could allow issuance of identifiers reserved to more critical systems, for example). Similar things can happen with different webhook endpoints configured. I foresee that different configurations being available and enforced might become more important with our upcoming feature work around policies too.

Instead of relying on the path part to check the relation between the account and the provisioner, it might be better to make this depend on the DB, by adding it as a property to the ACME account. That change might break existing deployments, and a migration or fallback logic would be required.

I tend to agree with @mholt that a strict equality check based on data in the DB makes sense. It seems to be more in line with how the `url` property has to be validated when proxies are involved (it's a different thing, but it's kind of a similar situation). However, I don't see a big problem with omitting the authority part of the URL: `step-ca` is capable of determining (or simply ignoring) that, and checking that the account for a KID exist (for a specific provisioner) sounds sufficient to me.

dopey:
We had the opportunity to discuss this issue / PR a bit more today.

@mholt generally speaking, I concur with your assessment - it's technically safer and more "correct" to test against the NewAccount URI that was used when the account was created. To do this we would need to add a field to the ACME account and have logic for backwards compatibility. I'm not sure that the cost here is worth the marginal benefit.

Instead, I prefer the solution that @hslatman is proposing (to address the issue both @hslatman and @maraino talk about in their comments) - compare the `kid` sent by the client against the NewAccount path, excluding the host:port. This has the benefit of asserting that the account ID was created by/for the given provisioner and I believe that it still addresses the spirit of the RFC. 

Let us know your thoughts @mholt.

dopey:
Backpedaling.

I'm now thinking that comparing the kid against the derived path does not necessarily result in the security profile we desire (the account must belong to the provisioner). I think @mholt's recommendation is the only one (of the one's we've discussed) that completely addresses this concern.

mholt:
Hi, sorry for the delay (it was spring break here). @dopey Yeah, I can understand the desire for not making that change. I can't really think of another way that gives you the desired guarantees/behavior for both client and server, either.

Really appreciating all the thought that is going into this :blush:

Let me know if I can help move things forward in any way.

PR #1366, opened 2023-04-28
Dynamic SCEP challenge validation using webhooks

hslatman:


scep/api/webhook/webhook.go
```diff
@@ -0,0 +1,65 @@
...
+// Controller controls webhook execution
+type Controller struct {
+	client   *http.Client
+	webhooks []*provisioner.Webhook
+}
+
+// New returns a new SCEP webhook Controller
+func New(webhooks []*provisioner.Webhook) (*Controller, error) {
+	return &Controller{
+		client:   http.DefaultClient,
```
areed, line 23: This should be configurable since hosted uses a custom transport that only allows connections to public IPs

hslatman: Fixed this in e8c1e8719d35aeedebd8ca19a407afc797a8f663.

scep/api/webhook/webhook.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		req := &webhook.RequestBody{
+			SCEPChallenge:     challenge,
+			SCEPTransactionID: transactionID,
+		}
+		resp, err := wh.DoWithContext(ctx, c.client, req, nil) // TODO(hs): support templated URL? Requires some refactoring
+		if err != nil {
+			return fmt.Errorf("failed executing webhook request: %w", err)
+		}
+		if resp.Allow {
```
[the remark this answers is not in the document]
hslatman: Will create a docs PR for this.

authority/authority.go
```diff
@@ -684,6 +640,50 @@ func (a *Authority) init() error {
 		return err
 	}
 
+	// Check if a KMS with decryption capability is required and available
```
areed, line 643: Why was this moved down here?

hslatman: It depends on the most recent state of the provisioners, specifically SCEP provisioners (the only ones using RSA decrypters). When they're (re)loaded from the database the provisioners in the authority configuration get updated in-memory. If checked before the reload, the list of provisioners would be empty, or would contain (potentially) outdated info.

This essentially was an existing bug that I found when testing this.

authority/provisioner/scep.go
```diff
@@ -82,6 +89,67 @@ func (s *SCEP) DefaultTLSCertDuration() time.Duration {
...
 }
 
+type challengeValidationController struct {
+	client   *http.Client
+	webhooks []*Webhook
+}
+
+// newChallengeValidationController creates a new challengeValidationController
+// that performs challenge validation through webhooks.
+func newChallengeValidationController(client *http.Client, webhooks []*Webhook) (*challengeValidationController, error) {
```
areed, line 99: error isn't used

areed (approved):

authority/provisioner/scep.go
hslatman: Removed in c73f157ea487376c1829e315a2f4c473740d393b

areed (approved):

Issue #1284, opened 2023-02-21
Serve CRL on HTTP


unreality:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

CRL generally wants to be served over HTTP, but currently is served over HTTPS

Would it be sufficient to alter the check at https://github.com/smallstep/certificates/blob/ca9f8dc576f0cb8cb5e4aed703783533d552e62e/ca/ca.go#L307 and start the HTTP server if CRL and/or SCEP is enabled and an insecure address is configured?

hslatman:
Yes, I think it makes sense to change the logic to also "activate" the insecure server to start if CRL is enabled. Additionally, the CRL endpoint must be mounted to the insecure server. Users of the CRL functionality should also ensure that the HTTP CRL URL is set in the issued certificates instead of the HTTPS one.

LarsBingBong:
This would be so nice ... currently I and other as well ( see [Certificate Revocation List ](#206) ) are doing hacks to fake HTTP to get CRL CDP and IDP endpoints to work over HTTP.

Thanks

hslatman:
I have opened a PR for this: https://github.com/smallstep/certificates/pull/1372

PR #1372, opened 2023-05-02
Add `/crl` and `/1.0/crl` to the insecure HTTP handler

hslatman:
In addition to being served over HTTPS, the `/crl` and `/1.0/crl` endpoints are now also served on the HTTP server if the `insecureAddress` has been configured. This should help users that currently have to use workarounds to get this to work in certain scenarios.

Related issues:
- https://github.com/smallstep/certificates/issues/1284 
- https://github.com/smallstep/certificates/issues/206#issuecomment-1530219201


maraino (approved):
lgtm

PR #1378, opened 2023-05-06
Store and verify account location on acme requests

dopey:


acme/api/middleware.go
```diff
@@ -324,6 +317,13 @@ func lookupJWK(next nextHTTP) nextHTTP {
 				render.Error(w, acme.NewError(acme.ErrorUnauthorizedType, "account is not active"))
 				return
 			}
+
+			if kid != acc.Location {
```
hslatman, line 321: It's a bit unclear what the values for `kid` and `Location` represent. In the first part, the `Location` seems to only contain the `new-account` URL; with the `kid` being the full key ID. But here it's compared directly, instead of the `Location` being suffixed with data. That suffix seems to be done in the storage layer. I think that shouldn't be done there for clarity.

acme/api/account.go
```diff
@@ -125,9 +125,10 @@ func NewAccount(w http.ResponseWriter, r *http.Request) {
 		}
 
 		acc = &acme.Account{
-			Key:     jwk,
-			Contact: nar.Contact,
-			Status:  acme.StatusValid,
+			Key:            jwk,
+			Contact:        nar.Contact,
+			Status:         acme.StatusValid,
+			LocationPrefix: linker.GetLink(r.Context(), acme.AccountLinkType, ""),
```
hslatman, line 131:
```suggestion
			LocationPrefix: linker.GetLink(ctx, acme.AccountLinkType, ""),
```

hslatman, line 131: This line of code is used twice and it may not be immediately obvious what it's doing. Can you create an unexported function that generates the prefix, use that in both places, and add a comment what the format is?

hslatman (commented):
Looks more logical now 😃 

I had one more thought about this: by storing the full URL (except for the ID), it _could_ happen that a provisioner with a specific name is deleted/replaced with another using the same name, resulting in the ACME account still being found (the URL would be the same), but now "belonging" to a different ACME provisioner. I think we should protect against that too, similar to how EAB credentials are ACME provisioner specific. So, if the URL checks out, but the provisioner isn't the correct one, we would create a new account (or when it's being looked up, report it's not found). WDYT?

acme/account.go
```diff
@@ -20,6 +20,15 @@ type Account struct {
...
+	LocationPrefix         string           `json:"-"`
+}
+
+// GetLocation returns the URL location of the given account.
+func (a *Account) GetLocation() string {
+	if a.LocationPrefix == "" {
+		return ""
+	}
+	return a.LocationPrefix + a.ID
 }
```
maraino, line 32: Can we add a test that ensures that the location prefix generated below, using `getAccountLocationPath(ctx, linker, "")`, always ends with `/` so this URL is properly formed?

acme/api/middleware.go
```diff
@@ -324,6 +317,23 @@ func lookupJWK(next nextHTTP) nextHTTP {
...
+
+			if storedLocation := acc.GetLocation(); storedLocation == "" {
+				// Old ACME accounts may not have a stored location.
+				if !strings.HasPrefix(kid, kidPrefix) {
+					render.Error(w, acme.NewError(acme.ErrorMalformedType,
+						"kid does not have required prefix; expected %s, but got %s",
+						kidPrefix, kid))
+					return
+				}
+			} else if kid != storedLocation {
```
maraino, line 329: Does it makes sense to use `!strings.EqualFold(kid, storeLocation)`?
If the kid is has base64 perhaps not.

maraino: I think we should store the full kid instead of constructing it. We're just saving a few bytes and we're adding more complexity.

maraino (changes requested):

PR #1386, opened 2023-05-11
Store and verify Acme account location 

### LINKED ISSUE #623: Changing ACME server host or port causes existing accounts to craft malformed requests

dopey:
Fixes #623

### DIFF
--- a/acme/account.go
+++ b/acme/account.go
@@ -20,7 +20,12 @@ type Account struct {
 	Status                 Status           `json:"status"`
 	OrdersURL              string           `json:"orders"`
 	ExternalAccountBinding interface{}      `json:"externalAccountBinding,omitempty"`
-	Location               string           `json:"-"`
+	LocationPrefix         string           `json:"-"`
+}
+
+// GetLocation returns the URL location of the given account.
+func (a *Account) GetLocation() string {
+	return a.LocationPrefix + a.ID
 }
 
 // ToLog enables response logging.
--- a/acme/account.go
+++ b/acme/account.go
@@ -32,7 +32,7 @@ func (a *Account) GetLocation() string {
 	return a.LocationPrefix + a.ID
 }
 
-// GetProvisionerName returns the provisioner name extracted from the stored
+// GetProvisionerName returns the provisioner name used to generate the account.
 // location.
 func (a *Account) GetProvisionerName() string {
 	return a.ProvisionerName
@@ -88,6 +88,7 @@ func (p *Policy) GetAllowedNameOptions() *policy.X509NameOptions {
 		IPRanges:   p.X509.Allowed.IPRanges,
 	}
 }
+
 func (p *Policy) GetDeniedNameOptions() *policy.X509NameOptions {
 	if p == nil {
 		return nil
--- a/acme/account.go
+++ b/acme/account.go
@@ -32,12 +32,6 @@ func (a *Account) GetLocation() string {
 	return a.LocationPrefix + a.ID
 }
 
-// GetProvisionerName returns the provisioner name used to generate the account.
-// location.
-func (a *Account) GetProvisionerName() string {
-	return a.ProvisionerName
-}
-
 // ToLog enables response logging.
 func (a *Account) ToLog() (interface{}, error) {
 	b, err := json.Marshal(a)
--- a/acme/account.go
+++ b/acme/account.go
@@ -4,6 +4,7 @@ import (
 	"crypto"
 	"encoding/base64"
 	"encoding/json"
+	"regexp"
 	"time"
 
 	"go.step.sm/crypto/jose"
@@ -29,6 +30,26 @@ func (a *Account) GetLocation() string {
 		return ""
 	}
 	return a.LocationPrefix + a.ID
+}
+
+// GetProvisionerName returns the provisioner name extracted from the stored
+// location.
+func (a *Account) GetProvisionerName() (string, error) {
+	if a.LocationPrefix == "" {
+		return "", NewError(ErrorServerInternalType, "cannot parse provisioner "+
+			"from account missing locationPrefix; accID = %q", a.ID)
+
+	}
+
+	re := regexp.MustCompile(`.*\/acme\/(.+?)\/.*`)
+	match := re.FindStringSubmatch(a.LocationPrefix)
+	if match == nil || len(match) < 2 {
+		// ACME provisioner not found in stored kid.
+		return "", NewError(ErrorServerInternalType,
+			"provisioner could not be parsed from stored locationPrefix; locationPrefix = %q", a.LocationPrefix)
+	}
+
+	return match[1], nil
 }
 
 // ToLog enables response logging.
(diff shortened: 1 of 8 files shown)


acme/account_test.go
```diff
@@ -66,6 +66,23 @@ func TestKeyToID(t *testing.T) {
 	}
 }
 
+func TestAccount_GetLocation(t *testing.T) {
+	locationPrefix := "https://test.ca.smallstep.com/acme/foo/account/"
```
maraino, line 70: To properly test that the URL is the right one you should get an account for NewAccount, or at least use something like:
```go
locationPrefix := getAccountLocationPath(ctx, linker, "")
```

acme/account_test.go
```diff
@@ -66,6 +66,23 @@ func TestKeyToID(t *testing.T) {
...
+	type test struct {
+		acc *Account
+		exp string
+	}
+	tests := map[string]test{
+		"empty":     {acc: &Account{LocationPrefix: ""}, exp: ""},
+		"not-empty": {acc: &Account{ID: "bar", LocationPrefix: locationPrefix}, exp: locationPrefix + "bar"},
```
maraino, line 77: With the above comment the expected should be `https://test.ca.smallstep.com/acme/foo/account/bar`

acme/api/middleware_test.go
```diff
@@ -754,7 +730,49 @@ func TestHandler_lookupJWK(t *testing.T) {
...
+				linker: acme.NewLinker("test.ca.smallstep.com", "acme"),
+				db: &acme.MockDB{
+					MockGetAccount: func(ctx context.Context, id string) (*acme.Account, error) {
+						assert.Equals(t, id, accID)
+						return acc, nil
+					},
+				},
+				ctx:        ctx,
+				statusCode: 401,
+				err:        acme.NewError(acme.ErrorUnauthorizedType, "kid does not matc hstored account location; expected foobar, but %q", prefix+accID),
```
maraino, line 747:
```suggestion
				err:        acme.NewError(acme.ErrorUnauthorizedType, "kid does not match stored account location; expected foobar, but %q", prefix+accID),
```

acme/db/nosql/account_test.go
maraino, line 1: Can we add LocationPrefix to one of the ok tests in UpdateAccount

maraino (changes requested):
Looks good, but I'd like a few changes in the tests.

maraino (approved):
lgtm

acme/account.go
```diff
@@ -20,6 +20,22 @@ type Account struct {
...
+func (a *Account) GetLocation() string {
+	if a.LocationPrefix == "" {
+		return ""
+	}
+	return a.LocationPrefix + a.ID
+}
+
+// GetProvisionerName returns the provisioner name extracted from the stored
+// location.
+func (a *Account) GetProvisionerName() string {
```
hslatman, line 37: The comment seems to reflect the old logic

acme/api/middleware.go
```diff
@@ -16,7 +17,6 @@ import (
 	"github.com/smallstep/certificates/api/render"
 	"github.com/smallstep/certificates/authority/provisioner"
 	"github.com/smallstep/certificates/logging"
-	"github.com/smallstep/nosql"
```
hslatman, line 19: Nice 😄

acme/api/middleware.go
```diff
@@ -324,6 +320,45 @@ func lookupJWK(next nextHTTP) nextHTTP {
...
+				// Verify that the provisioner with which the account was created
+				// matches the provisioner in the request URL.
+				reqProv := acme.MustProvisionerFromContext(ctx)
+				reqProvName := reqProv.GetName()
+				accProvName := acc.GetProvisionerName()
+				if reqProvName != accProvName {
+					// Provisioner in the URL must match the provisioner with
+					// which the account was created.
+					render.Error(w, acme.NewError(acme.ErrorUnauthorizedType,
+						"account provisioner does not match requested provisioner; account provisioner = %s, reqested provisioner = %s",
```
hslatman, line 343:
```suggestion
						"account provisioner does not match requested provisioner; account provisioner = %s, requested provisioner = %s",
```

acme/db/nosql/account_test.go
```diff
@@ -686,12 +708,7 @@ func TestDB_UpdateAccount(t *testing.T) {
 					assert.HasPrefix(t, err.Error(), tc.err.Error())
 				}
 			} else {
-				if assert.Nil(t, tc.err) {
-					assert.Equals(t, tc.acc.ID, dbacc.ID)
-					assert.Equals(t, tc.acc.Status, dbacc.Status)
-					assert.Equals(t, tc.acc.Contact, dbacc.Contact)
-					assert.Equals(t, tc.acc.Key.KeyID, dbacc.Key.KeyID)
-				}
```
hslatman, line 694: Is there a reason for removing these assertions? I guess they were kind of superfluous with the assertions being done in the mocked functions?

dopey: Exactly this. They were already being asserted from within the mock.

maraino (approved):
lgtm

There's no need for `GetProvisionerName()` because the ProvisionerName is public, and it doesn't have any logic.

maraino (approved):

PR #1400, opened 2023-05-22
Bump to nebula 1.7.1 updated cert.Sign api

dopey:


dopey:
Closing for now because Nebula v1.7.1 doesn't compile with go1.19.

PR #1404, opened 2023-05-25
add AWS public certificates for me-central-1 and ap-southeast-3

rnugmanov:
As per https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-signature.html

#### Name of feature:
adding missing regions
AWS provisioner doesn't work in these two regions as the certificates are different

💔Thank you!


maraino:
Curiously, these certificates are the most insecure of all of them, 1024-bit RSA with SHA-1.

maraino (approved):
Thanks @rnugmanov, lgtm.

FYI: you can also define the `iidRoots` in the provisioner configuration to support new ones.

rvichery:
> Thanks @rnugmanov, lgtm.
> 
> FYI: you can also define the `iidRoots` in the provisioner configuration to support new ones.

@maraino when defining the iidRoots in the provisioner would not that only work on the server side? I thought there was also a signature check on the client side (step-cli)?

maraino:
> when defining the iidRoots in the provisioner would not that only work on the server side? I thought there was also a signature check on the client side (step-cli)?

As it yes, the cli only uses the hardcoded certificates in the CA.

rvichery:
> > when defining the iidRoots in the provisioner would not that only work on the server side? I thought there was also a signature check on the client side (step-cli)?
> 
> As it yes, the cli only uses the hardcoded certificates in the CA.

So, what is the use case for using `iidRoots` on the server side? 
I am preparing a PR to add all remaining region, so I was just curious if there is a better way to handle this without changing the hardcoded certs.

maraino:
@rvichery I've created this https://github.com/smallstep/cli/issues/1005

maraino:
> I am preparing a PR to add all remaining region, so I was just curious if there is a better way to handle this without changing the hardcoded certs.

I created the issue above, but please add a PR with the missing regions.

rvichery:
> > I am preparing a PR to add all remaining region, so I was just curious if there is a better way to handle this without changing the hardcoded certs.
> 
> I created the issue above, but please add a PR with the missing regions.

Thank you for creating the issue, here is the PR with all existing AWS regions https://github.com/smallstep/certificates/pull/1510

PR #1414, opened 2023-06-01
Add SCEP provisioner decrypter

hslatman:
This PR introduces support for provisioner specific decrypters for SCEP. This allows a SCEP provisioner to be configured with an RSA key used for decrypting SCEP requests independent of the authority intermediate signing key. In turn, this means that one can then add SCEP provisioners to CAs that have been configured with e.g. an ECDSA intermediate signing key.

Changes from https://github.com/smallstep/linkedca/pull/55 need to be merged first.

To configure a SCEP provisioner using the CLI, changes from https://github.com/smallstep/cli/pull/950 are needed.

In `ca.json`, the configuration would look similar to the below examples:

```json
        {
                "type": "SCEP",
                "name": "scepca-with-cloudkms",
                "forceCN": true,
                "includeRoots": false,
                "encryptionAlgorithmIdentifier": 2,
                "decrypterCertificate": "LS0tLS1CRUdJT....FURS0tLS0tCg==",
                "decrypterKey": "cloudkms:projects/myProjectID/locations/global/keyRings/myKeyRingt/cryptoKeys/myRSAKey/cryptoKeyVersions/1"
        },
        {
                "type": "SCEP",
                "name": "scepca-with-softkms",
                "forceCN": true,
                "includeRoots": false,
                "encryptionAlgorithmIdentifier": 2,
                "decrypterCertificate": "LS0tLS1CRUdJTiB....VEUtLS0tLQo=",
                "decrypterKey": "softkms:path/to/private.key",
                "decrypterKeyPassword": "c3VwZXItc2VjcmV0LXBhc3N3b3Jk"
        },
        {
                "type": "SCEP",
                "name": "scepca-with-pem",
                "forceCN": true,
                "includeRoots": false,
                "encryptionAlgorithmIdentifier": 2,
                "decrypterCertificate": "LS0tLS1CRUdJTiB....VEUtLS0tLQo=",
                "decrypterKeyPEM": "LS0tLS1CRUdJTiBSU....kFURSBLRVktLS0tLQo=",
                "decrypterKeyPassword": "c3VwZXItc2VjcmV0LXBhc3N3b3Jk"
        },
```

The `decrypterCertificate` should contain the base64 encoding of a PEM formatted certificate.

This has been tested with `softkms` and `cloudkms`, so far. One limitation with using `softkms` is that the password currently needs to be provided in the configuration; it can't be read from a file or the environment at this time.

In addition to the functional changes, this PR also simplifies how the SCEP authority is initalized and validated, as well as how the SCEP provisioner is passed by context.


authority/provisioner/scep.go
```diff
@@ -162,21 +179,97 @@ func (s *SCEP) Init(config Config) (err error) {
...
 		return errors.New("only encryption algorithm identifiers from 0 to 4 are valid")
 	}
 
+	// Prepare the SCEP challenge validator
 	s.challengeValidationController = newChallengeValidationController(
 		config.WebhookClient,
 		s.GetOptions().GetWebhooks(),
 	)
 
+	if decryptionKey := s.DecrypterKey; decryptionKey != "" {
```
maraino, line 198: I started to look into this, and when a DB backs step-ca, we should store the key in the DB. We can also consider only supporting this case, even if the ca.json is used:

```json
{
  "...": "...",
  "decrypterKey": "LS0tLS1CRUdJTiBSU0EgUFJJVkFURSBLRVktLS0tLQo="
}
```

We must redact this and the password in the HTTP responses for provisioner and admin endpoints.

If we want to use files, we can consider an strategy used in the templates:

```go
type X509Options struct {
  Template string `json:"template,omitempty"`
  TemplateFile string `json:"templateFile,omitempty"`
  // ...
}
```

hslatman: Redaction can happen in the `api` package. See https://github.com/smallstep/certificates/pull/1414/files#diff-bf98d5fab5bcfabded9769069e7318683f549ec91cff84c5f87a9ae3a3ca3d8dR237. It's not super nice to have to replicate all the methods, but it doesn't rely on changing the values in the collection directly and also prevents refactoring the configuration of the provisioner vs. it being used.

Storing just the private key as bytes sounds OK to me, but we may need to support a password too? And there's also the case when it's provided as URI?

Having an option with and without a file sounds OK too.

maraino: Perhaps something like this will support most of the cases:
```proto
message SCEPDecrypter {
    bytes decrypterCertificate = 1; // base64 cert
    bytes decrypterKey = 2;         // base64 key
    bytes decrypterKeyPassword = 3; // base64 (optional) password
    string decrypterKeyUri = 4;     // kmsuri
}
```

A URI can also be a file using `softkms:path/to/file.crt`. We can also add support to detect full/relative paths using `[/]path/to/file.crt`, or we can simplify it by adding a new `string decrypterKeyFile = 5;`

We might need a `KMS kms = 1;` like you have in the linkedca PR, but if we can avoid it, it would be great.

hslatman: The `linkedca.KMS` probably can be removed. It's what I tried first, but then the changes to parsing the KMS from an URI were introduced, and that's in use now to determine the type of KMS to use when configuring a key. There's a couple of special cases for `softkms` that I think can be improved, but I don't think they're blockers.

Having just the base64 cert is probably OK. That just works, and keeps thing simple in terms of storage, both in standalone (with and without remote administration enabled) and in hosted.

For the key, we would thus detect if the URI is set vs. the key is set, enforce and use only one of them? The key password could be empty, if the key URI has the appropriate fields set?

api/api.go
```diff
@@ -231,31 +232,52 @@ type ProvisionersResponse struct {
...
+		ForceCN:                       p.ForceCN,
+		ChallengePassword:             redacted,
+		Capabilities:                  p.Capabilities,
+		IncludeRoot:                   p.IncludeRoot,
+		ExcludeIntermediate:           p.ExcludeIntermediate,
+		MinimumPublicKeyLength:        p.MinimumPublicKeyLength,
+		DecrypterCertificate:          []byte(redacted),
+		DecrypterKeyPEM:               []byte(redacted),
+		DecrypterKeyURI:               redacted,
+		DecrypterKeyPassword:          []byte(redacted),
```
maraino, line 251: I've fixed this in https://github.com/smallstep/certificates/pull/1554

authority/authority.go
```diff
@@ -640,48 +662,83 @@ func (a *Authority) init() error {
...
-				Password:      a.password,
-			})
-			if err != nil {
-				return err
+		if a.validateSCEP {
+			// validate the SCEP authority
+			if err := scepAuthority.Validate(); err != nil {
+				a.initLogf("failed validating SCEP authority: %v", err)
 			}
 		}
```
maraino, line 728: This is not ideal, but for now, it is ok.

authority/provisioner/scep.go
```diff
@@ -29,9 +36,19 @@ type SCEP struct {
...
+	ExcludeIntermediate bool `json:"excludeIntermediate,omitempty"`
+
 	// MinimumPublicKeyLength is the minimum length for public keys in CSRs
 	MinimumPublicKeyLength int `json:"minimumPublicKeyLength,omitempty"`
 
+	// TODO(hs): also support a separate signer configuration?
+	DecrypterCertificate []byte `json:"decrypterCertificate,omitempty"`
+	DecrypterKeyPEM      []byte `json:"decrypterKeyPEM,omitempty"`
+	DecrypterKeyURI      string `json:"decrypterKey,omitempty"`
+	DecrypterKeyPassword []byte `json:"decrypterKeyPassword,omitempty"`
```
maraino, line 50: We might have users using a release candidate where this is a `string`. I would prefer to use `[]byte` for this. But custom `Password` type that can have either base64 or plain passwords can be considered.

scep/options.go
```diff
@@ -4,65 +4,78 @@ import (
...
 func (o *Options) Validate() error {
-	if o.CertificateChain == nil {
-		return errors.New("certificate chain not configured correctly")
+	switch {
+	case len(o.Intermediates) == 0:
+		return errors.New("no intermediate certificate available for SCEP authority")
+	case o.Signer == nil:
+		return errors.New("no signer available for SCEP authority")
+	case o.SignerCert == nil:
+		return errors.New("no signer certificate available for SCEP authority")
```
maraino, line 43: This is ok for now, but when these fields are always configured at the provisioner level, these fields are not required.

authority/provisioner/scep.go
```diff
@@ -162,21 +245,139 @@ func (s *SCEP) Init(config Config) (err error) {
...
+		if s.signer, err = s.keyManager.CreateSigner(&kmsapi.CreateSignerRequest{
+			SigningKeyPEM:    decryptionKeyPEM, // TODO(hs): support distinct signer key in the future?
+			Password:         s.DecrypterKeyPassword,
+			PasswordPrompter: kmsapi.NonInteractivePasswordPrompter,
+		}); err != nil {
+			return fmt.Errorf("failed creating signer: %w", err)
+		}
+	}
+
+	if decryptionKeyURI := s.DecrypterKeyURI; len(decryptionKeyURI) > 0 {
```
maraino, line 306: Ok for now, but we might want to consider using only one option.

maraino (approved):
Approving this for now, but there are some changes we might want to consider before creating a new release.

api/api.go
hslatman: The idea was to use a separate model for the public API response, so that `*** REDACTED ***` would show up instead of the base64 encoding of that. So we may want to revisit this, and make the CLI use the "public" model instead. But there's probably more to be discussed around that, since I only did this for this single provisioner/endpoint.

authority/provisioner/scep.go
hslatman: People may have been using it, but those changes were only in a (draft) PR until now, so I don't think we need to spend a lot of time on keeping this compatible. I know of at least one user who tested this with the `decrypterKey` with `cloudkms`, and that doesn't need the password.

scep/options.go
hslatman: A thought just occurred to me: we probably still need this, and need some way to explicitly _not_ rely on the provisioner specific signer in specific situations.

PR #1446, opened 2023-06-21
Address Makefile issues around the cgo build

tashian:
Ok y’all, I’m pulling something out of the rabbit hole here.

I’ve noticed an issue with the Makefile for step-ca, where we use the variable `GOFLAGS`.
Actually, this has been confusing to me for a long time and I finally discovered why.
In the Makefile, we run, for example `$(GOOS_OVERRIDE) $(GOFLAGS) go build -v -o $(PREFIX)bin/$(BINNAME) $(LDFLAGS) $(PKG)`
By default, `GOFLAGS` is set to the value `CGO_ENABLED=0`
But if I run `make build GOFLAGS="CGO_ENABLED=1"`, I get:
```
go: parsing $GOFLAGS: non-flag "CGO_ENABLED=1"
make: *** [Makefile:67: download] Error 1
```

In this case, `GOFLAGS ` is being used both by the Makefile (as go env variables) _and by go itself_ (as “A space-separated list of -flag=value settings to apply to go commands by default”)

So, then, how does the default case even work, since by default `GOFLAGS` is equal to `CGO_ENABLED=0`, a non-flag?
It works because by default `GOFLAGS` is set locally in the Makefile, and it won’t get exported to go as an env.

The only valid value for `GOFLAGS` that satisfies both go and the Makefile is “”
And this is the only value we use in automations and docs, because it creates a cgo build.
Or, it’s supposed to.
But sometimes it results in a non-cgo build.

Quick sideshow:
Normally, Go creates cgo binaries by default.
But (I think this is a recent change to Go), if `gcc`  is not in `$PATH`, it will silently not use cgo.
So, the second issue is that our instructions for cgo builds might silently create a non-cgo build, as a user just encountered.
So we need to tell people to run make with `CGO_ENABLED=1` set explicitly.
But there isn’t a way to do that. For some reason, make build `CGO_ENABLED=1` does not pass `CGO_ENABLED=1` to go.

Bottom line:
- The way we’re using it, `GOFLAGS` should be called `GO_ENVS`
- We should explicitly set `CGO_ENABLED=1` if we want a cgo binary, so that when `gcc` doesn't exist, `go` will fail to compile instead of falling back to non-cgo
- `GOFLAGS` should be reserved for `go`

A couple backward-compatible options for getting out of this:
1. Change `GOFLAGS` to `GO_ENVS` to represent go build flags; and if `GOFLAGS` is passed in as `""`, silently set `GO_ENVS` to `CGO_ENABLED=1`. Otherwise, default it to `CGO_ENABLED=0`. This is what I did in this PR.
2. Don't bother with `GO_ENVS`, just accept `CGO_ENABLED=1` into `make` and export it to `go`.  If `GOFLAGS` is passed in as `""`, silently set `CGO_ENABLED=1`.

Alternatively, we could make a breaking change and just update the docs and cgo-related automations.


Makefile
```diff
@@ -61,7 +61,15 @@ endif
...
 LDFLAGS := -ldflags='-w -X "main.Version=$(VERSION)" -X "main.BuildTime=$(DATE)"'
-GOFLAGS := CGO_ENABLED=0
+GOFLAGS := -v
+
+ifeq (,$(GOFLAGS))
+	ifeq (,$(findstring CGO_ENABLED=0,$(GO_ENVS)))
+		GO_ENVS := $(GO_ENVS) CGO_ENABLED=1
+	endif
+else
+	GO_ENVS := $(GO_ENVS) CGO_ENABLED=0
```
maraino, line 71: I cannot make this work as "expected". Those variables are only expanded depending on how you run the command:

| Command                        | GO_ENVS                 |
| ------------------------------ | ----------------------- |
| `make build GO_ENVS="A=1 B=2"` | `A=1 B=2`               |
| `GO_ENVS="A=1 B=2" make build` | `A=1 B=2 CGO_ENABLED=0` |

So following the same convention:
| Command                              | GO_ENVS                      |
| ------------------------------------ | ---------------------------- |
| `GO_ENVS="CGO_ENABLED=1" make build` |`CGO_ENABLED=1 CGO_ENABLED=0` |

Makefile
```diff
@@ -61,7 +61,15 @@ endif
 
 DATE    := $(shell date -u '+%Y-%m-%d %H:%M UTC')
 LDFLAGS := -ldflags='-w -X "main.Version=$(VERSION)" -X "main.BuildTime=$(DATE)"'
-GOFLAGS := CGO_ENABLED=0
+GOFLAGS := -v
+
+ifeq (,$(GOFLAGS))
+	ifeq (,$(findstring CGO_ENABLED=0,$(GO_ENVS)))
+		GO_ENVS := $(GO_ENVS) CGO_ENABLED=1
```
maraino, line 68: It's weird having `GOFLAGS` and `GO_ENVS` instead of `GO_FLAGS`/`GOENVS`

Makefile
```diff
@@ -61,7 +61,15 @@ endif
 
 DATE    := $(shell date -u '+%Y-%m-%d %H:%M UTC')
 LDFLAGS := -ldflags='-w -X "main.Version=$(VERSION)" -X "main.BuildTime=$(DATE)"'
-GOFLAGS := CGO_ENABLED=0
+GOFLAGS := -v
+
+ifeq (,$(GOFLAGS))
+	ifeq (,$(findstring CGO_ENABLED=0,$(GO_ENVS)))
```
maraino, line 67: I guess this is for backward compatibility, right?

maraino (changes requested):
The expansion of the variables, is a little weird `GO_ENVS ?= CGO_ENABLED=0` might be better.

Makefile
tashian: Yes. If we decide to go with this approach, then I'll add a comment to it to be clear.

Makefile
tashian: All of the `go` commmand [environment variables](https://pkg.go.dev/cmd/go#hdr-Environment_variables) are `GO*`, and I didn't want to clash with that namespace at all. So I added `GO_ENVS` instead of `GOENVS`.

And I'm using `GOFLAGS` to refer to the official Go environment variable of that name.

Makefile
tashian: Good catch, this is fixed now.

tashian:
@maraino ok I think this is ready.

Tests I've run:
| Command     |  Result |
| ------- | --------------- |
| `make build` | cgo disabled |
| `make build GOFLAGS=""` | cgo enabled | 
| `make build GOFLAGS="-x"` | cgo disabled; -x passed to `go` |
| `make build GO_ENVS="CGO_ENABLED=0"` | cgo disabled |
| `make build GO_ENVS="CGO_ENABLED=1"` | cgo enabled |
| `make build GO_ENVS="CGO_ENABLED=0" GOFLAGS=""` | cgo disabled |
| `make build GO_ENVS="CGO_ENABLED=1" GOFLAGS="-x"` | cgo enabled; -x passed to `go` |

Makefile
```diff
@@ -93,7 +109,7 @@ generate:
 test: testdefault testtpmsimulator combinecoverage
 
 testdefault:
-	$Q $(GOFLAGS) gotestsum -- -coverprofile=defaultcoverage.out -short -covermode=atomic ./...
+	$Q $(GO_ENVS) gotestsum -- -coverprofile=defaultcoverage.out -short -covermode=atomic ./...
 
 testtpmsimulator:
 	$Q CGO_ENALBED=1 gotestsum -- -coverprofile=tpmsimulatorcoverage.out -short -covermode=atomic -tags tpmsimulator ./acme
```
hslatman, line 115: Hijacking your PR for this quick fix:
```suggestion
	$Q CGO_ENABLED=1 gotestsum -- -coverprofile=tpmsimulatorcoverage.out -short -covermode=atomic -tags tpmsimulator ./acme 
```

README.md
```diff
@@ -119,18 +119,12 @@ See our installation docs [here](https://smallstep.com/docs/step-ca/installation
...
-
-1. On the web at https://smallstep.com/docs/step-ca.
-
-2. On the command line with `step help ca xxx` where `xxx` is the subcommand
-you are interested in. Ex: `step help ca provisioner list`.
-
-3. In your browser, by running `step help --http=:8080 ca` from the command line
+* [Official documentation](https://smallstep.com/docs/step-ca) is on smallstep.com
+* The `step` command reference is available via `step help`,
+[or on smallstep.com](https://smallstep.com/docs/step-cli/reference/),
```
hslatman, line 124:
```suggestion
[on smallstep.com](https://smallstep.com/docs/step-cli/reference/),
```

hslatman (approved):

maraino (approved):

PR #1471, opened 2023-07-10
Cache-Control HTTP header for server endpoints

### LINKED ISSUE #793: Add `Cache-Control: private, no-store` headers where appropriate
The recommendation is to set `Cache-Control: private, no-store` on any endpoint with sensitive information. Because while you can protect the traffic with TLS, you also need to keep sensitive information out of a client's (unencrypted) HTTP cache. I'm not sure how relevant this is to the API context though—I've never seen an HTTP client library that caches content. But I guess the point here is that a client _could_ cache any content unless we tell it not to.

km274:
Add `Cache-Control: private, no-store` HTTP header to server endpoints that respond with sensitive info.

Fixes #793

I decided which endpoints to add the header to by investigating each endpoint and the data it returns.

Please let me know if it would be a good idea to update or add tests to check for the new header where needed. My own judgment as of the time of opening this PR was to not update the tests because 1) existing tests aren't checking HTTP response headers from what I saw and 2) I felt like the addition of a header might be too trivial to be worth adding new test code.

### DIFF
--- a/authority/admin/api/provisioner.go
+++ b/authority/admin/api/provisioner.go
@@ -55,6 +55,8 @@ func GetProvisioner(w http.ResponseWriter, r *http.Request) {
 		render.Error(w, err)
 		return
 	}
+
+	w.Header().Set("Cache-Control", "private, no-store")
 	render.ProtoJSON(w, prov)
 }
 
@@ -72,6 +74,7 @@ func GetProvisioners(w http.ResponseWriter, r *http.Request) {
 		render.Error(w, errs.InternalServerErr(err))
 		return
 	}
+
 	render.JSON(w, &GetProvisionersResponse{
 		Provisioners: p,
 		NextCursor:   next,
@@ -102,6 +105,8 @@ func CreateProvisioner(w http.ResponseWriter, r *http.Request) {
 		render.Error(w, admin.WrapErrorISE(err, "error storing provisioner %s", prov.Name))
 		return
 	}
+
+	w.Header().Set("Cache-Control", "private, no-store")
 	render.ProtoJSONStatus(w, prov, http.StatusCreated)
 }
 
@@ -198,6 +203,8 @@ func UpdateProvisioner(w http.ResponseWriter, r *http.Request) {
 		render.Error(w, err)
 		return
 	}
+
+	w.Header().Set("Cache-Control", "private, no-store")
 	render.ProtoJSON(w, nu)
 }
 
--- a/authority/admin/api/webhook.go
+++ b/authority/admin/api/webhook.go
@@ -127,6 +127,7 @@ func (war *webhookAdminResponder) CreateProvisionerWebhook(w http.ResponseWriter
 		return
 	}
 
+	w.Header().Set("Cache-Control", "private, no-store")
 	render.ProtoJSONStatus(w, newWebhook, http.StatusCreated)
 }
 
@@ -231,5 +232,7 @@ func (war *webhookAdminResponder) UpdateProvisionerWebhook(w http.ResponseWriter
 		Auth:                 newWebhook.Auth,
 		DisableTlsClientAuth: newWebhook.DisableTlsClientAuth,
 	}
+
+	w.Header().Set("Cache-Control", "private, no-store")
 	render.ProtoJSONStatus(w, whResponse, http.StatusCreated)
 }


kippmorris7:
I checked out the CI test failure (see excerpt from logs below). Since this particular failure is present in some other recent CI runs and isn't related to my changes, I'll just continue to wait for feedback.

```
CGO_ENALBED=1 gotestsum -- -coverprofile=tpmsimulatorcoverage.out -short -covermode=atomic -tags tpmsimulator ./acme 
# github.com/smallstep/certificates/acme [github.com/smallstep/certificates/acme.test]
Error: acme/challenge_tpmsimulator_test.go:52:8: assignment mismatch: 1 variable but simulator.New returns 2 values
WARN invalid TestEvent: FAIL	github.com/smallstep/certificates/acme [build failed]
bad output from test2json: FAIL	github.com/smallstep/certificates/acme [build failed]

=== Errors
Error: acme/challenge_tpmsimulator_test.go:52:8: assignment mismatch: 1 variable but simulator.New returns 2 values

DONE 0 tests, 1 error in 25.586s
make: *** [Makefile:99: testtpmsimulator] Error 2
Error: Process completed with exit code 2.
```

hslatman:
Hey @kippmorris7, you're right about the CI issue. I fixed it yesterday with https://github.com/smallstep/certificates/pull/1464. If you merge the latest master, yours will run OK too.

acme/api/account.go
```diff
@@ -162,6 +162,7 @@ func NewAccount(w http.ResponseWriter, r *http.Request) {
 	linker.LinkAccount(ctx, acc)
 
 	w.Header().Set("Location", getAccountLocationPath(ctx, linker, acc.ID))
+	w.Header().Set("Cache-Control", "private, no-store")
```
areed, line 165: I don't see anything sensitive in the ACME account object. What data was of concern on this response?

acme/api/account.go
```diff
@@ -212,6 +213,7 @@ func GetOrUpdateAccount(w http.ResponseWriter, r *http.Request) {
 	linker.LinkAccount(ctx, acc)
 
 	w.Header().Set("Location", linker.GetLink(ctx, acme.AccountLinkType, acc.ID))
+	w.Header().Set("Cache-Control", "private, no-store")
```
areed, line 216: Another ACME Account.

acme/api/handler.go
```diff
@@ -306,6 +306,7 @@ func GetAuthorization(w http.ResponseWriter, r *http.Request) {
 	linker.LinkAuthorization(ctx, az)
 
 	w.Header().Set("Location", linker.GetLink(ctx, acme.AuthzLinkType, az.ID))
+	w.Header().Set("Cache-Control", "private, no-store")
```
areed, line 309: This is an ACME authorization object which includes a list of challenges with tokens. I don't believe the token is  sensitive since leaking it does not compromise the protocol. Is there another field on here of concern?

acme/api/handler.go
```diff
@@ -359,6 +360,7 @@ func GetChallenge(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Add("Link", link(linker.GetLink(ctx, acme.AuthzLinkType, azID), "up"))
 	w.Header().Set("Location", linker.GetLink(ctx, acme.ChallengeLinkType, azID, ch.ID))
+	w.Header().Set("Cache-Control", "private, no-store")
```
areed, line 363: This is an ACME challenge object with a token.

api/api.go
```diff
@@ -379,6 +379,7 @@ func Provisioners(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	w.Header().Set("Cache-Control", "private, no-store")
```
areed, line 382: Provisioner responses do not contain sensitive information.

api/api.go
```diff
@@ -394,6 +395,7 @@ func ProvisionerKey(w http.ResponseWriter, r *http.Request) {
...
```
areed, line 398: This key is encrypted.

authority/admin/api/provisioner.go
```diff
@@ -55,6 +55,8 @@ func GetProvisioner(w http.ResponseWriter, r *http.Request) {
 		render.Error(w, err)
 		return
 	}
+
+	w.Header().Set("Cache-Control", "private, no-store")
```
areed, line 59: Provisioner response.

authority/admin/api/provisioner.go
```diff
@@ -72,6 +74,8 @@ func GetProvisioners(w http.ResponseWriter, r *http.Request) {
 		render.Error(w, errs.InternalServerErr(err))
 		return
 	}
+
+	w.Header().Set("Cache-Control", "private, no-store")
```
areed, line 78: Provisioner respone.

authority/admin/api/provisioner.go
```diff
@@ -102,6 +106,8 @@ func CreateProvisioner(w http.ResponseWriter, r *http.Request) {
 		render.Error(w, admin.WrapErrorISE(err, "error storing provisioner %s", prov.Name))
 		return
 	}
+
+	w.Header().Set("Cache-Control", "private, no-store")
```
areed, line 110: Provisioner response

authority/admin/api/provisioner.go
```diff
@@ -198,6 +204,8 @@ func UpdateProvisioner(w http.ResponseWriter, r *http.Request) {
 		render.Error(w, err)
 		return
 	}
+
+	w.Header().Set("Cache-Control", "private, no-store")
```
areed, line 208: Provisioner response

areed (commented):
Webhook create and update responses contain secrets and should not be cached.

Provisioner responses never contain sensitive data.

ACME responses contain some data that may be considered sensitive. I'm not certain how to treat these.

api/api.go
kippmorris7: I see that now. I'll remove the header here.

api/api.go
kippmorris7: I decided to add the header to the provisioner responses because I saw that a few of the provisioners (or at least the ones I checked) contain webhooks nested inside. Taking the AWS provisioner in `authority/provisioners/aws.go` as an example, it's `ctl` field of type `Controller` has a field `webhooks` that stores a list of webhooks.

And regarding the provisioner responses over in `authority/admin/api/provisioners.go`, it looks like the provisioners are represented with a different data type (`linkedca.Provisioner`) there, but it still contains webhooks.

So if I'm understanding correctly the provisioner responses in both packages do contain sensitive information.

acme/api/account.go
kippmorris7: When I looked to see what it contained, I saw the `jose.JSONWebKey` which, if you go to the type definition, has a comment saying it "represents a public or private key". But after looking into the ACME protocol a bit, it seems to me that the server would be returning the public key here. So I'll remove the header here if you agree that we should.

acme/api/handler.go
kippmorris7: I read into the protocol a bit and I see why caching the token wouldn't compromise the protocol. I'll remove the header here.

acme/api/handler.go
kippmorris7: Will remove the header here per my response to the other comment about the challenge tokens.

kippmorris7:
Thank you for the review, @areed ! After looking back over my work, I agree that we can remove the headers in most of the places you pointed out, but I think we still need it on the provisioner responses.

areed:
@kippmorris7 [This is the type](https://github.com/smallstep/certificates/blob/d39a28535df1e676dd9e548be1a589d44589ec72/authority/provisioner/webhook.go#L94) used for webhooks when they appear in provisioners. The secret and credentials are always omitted.

kippmorris7:
> @kippmorris7 [This is the type](https://github.com/smallstep/certificates/blob/d39a28535df1e676dd9e548be1a589d44589ec72/authority/provisioner/webhook.go#L94) used for webhooks when they appear in provisioners. The secret and credentials are always omitted.

I see. In that case I understand why we can remove the header from `Provisioners` in `api/api.go`. I double checked the `linkedca.Provisioner` though, and it looks like the the secrets will not be omitted there:

- Link to `linkedca.Provisioner`: https://github.com/smallstep/linkedca/blob/1389fd5b2d901b09abf84d678c89f63c3d29ae54/provisioners.pb.go#L324
- Link to nested `Webhook` type: https://github.com/smallstep/linkedca/blob/1389fd5b2d901b09abf84d678c89f63c3d29ae54/provisioners.pb.go#L1950

With this in mind, it seems to me that in `authority/admin/api/provisioners.go`, we can safely remove the header from `GetProvisioners`, but we should leave it in the other functions, since they respond with `linkedca.Provisioner`s. What do you think?

areed:
> With this in mind, it seems to me that in authority/admin/api/provisioners.go, we can safely remove the header from GetProvisioners, but we should leave it in the other functions, since they respond with linkedca.Provisioners. What do you think?

Yes, you're right.

api/api.go
kippmorris7: Removed the header here.

authority/admin/api/provisioner.go
kippmorris7: Removed the header here.

authority/admin/api/provisioner.go
kippmorris7: Leaving the header here.

authority/admin/api/provisioner.go
kippmorris7: Leaving the header here.

authority/admin/api/provisioner.go
kippmorris7: Leaving the header here.

kippmorris7:
Cool, I removed the header from two of the provisioner responses and left it on the others like we discussed.

We still need to address the ACME endpoints. Having done my reading, I'm ready to remove them based on my own understanding, but I wasn't clear on whether you were waiting for a third opinion first.

acme/api/account.go
kippmorris7: Removed the header here.

acme/api/handler.go
kippmorris7: Removed the header here.

scep/api/api.go
```diff
@@ -359,6 +359,7 @@ func writeResponse(w http.ResponseWriter, res Response) {
 	}
 
 	w.Header().Set("Content-Type", contentHeader(res))
+	w.Header().Set("Cache-Control", "private, no-store")
```
areed, line 362: @hslatman Would the SCEP response data ever contain sensitive information? From what I can tell it's either a certificate or a list of capabilities and would never contain the challenge secret.

tashian:
I'm glad this is finally being addressed; thank you @kippmorris7 !

scep/api/api.go
hslatman: Any secrets are encrypted before being sent to the CA, and not returned to clients, so no obvious sensitive information is leaked. There's some bits of SCEP transaction related information that can be extracted, such as transaction ID and nonces, but I don't think these leak information; at least not directly.  

While some responses to GET requests _could_ be cached for a specific amount of time (e.g. `GetCACert`, `GetCACaps`), I feel responses to POST requests must not be cached. The last (informational) SCEP RFC doesn't say much about caching, only that it has caused issues in the past with large requests and when requests weren't actually idempotent. 

We currently support and signal the more modern `POSTPKIOperation` (via POST), so clients generally don't send the large GET requests. The responses for POST requests contain client specific results and are unlikely to be fully idempotent, so IMO shouldn't be cached.

In terms of interoperability: I would have to take a look at other SCEP servers if they ever set cache headers. I think clients should simply ignore them when they don't understand them or can't do anything with them, but SCEP being legacy, there's a lot of old systems that may be doing weird things.

areed: @kippmorris7 Let's remove this one and then the PR will be good to merge.

kippmorris7: I've removed it. To double check my understanding, why did you decide against keeping the headers on the POST responses?

kippmorris7:
Hey everyone, I just rebased since it's been a while. Is there anything else you'd like me to take another look at before this is ready to merge?

PR #1478, opened 2023-07-18
Customize principal validation using an environment variable

### LINKED ISSUE #1436: Allow SSH usernames to begin with a digit
## Hello!

## Issue details

We have ssh usernames that start with numeric values (0-9). The current `sshUserRegex` precludes anything but a letter (a-z). This inhibits our ability to use the CA without a patch. If possible, the regex should express the first character as `[a-z0-9]`.

## Why is this needed?

The current implementation blocks valid email addresses.

maraino:
### Description

By default, the OIDC user principal must validate the regular expression "^[a-z][-a-z0-9_]*$", but with this commit, a custom regular expression can be defined using the environment variable STEP_SSH_USER_REGEXP.

Fixes #1436

### DIFF
--- a/authority/provisioner/controller.go
+++ b/authority/provisioner/controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/x509"
 	"net/http"
+	"os"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


hslatman (commented):
Code looks OK.

Not very nice to do it in `init`, but this works if it's really required.

An SSH username extracted with a custom regex, for example when starting with a digit, may not work on the system where the user is trying to use it, correct? To me sounds like a mapping in a template in addition to this should maybe be the full solution if we describe this env var?

maraino:
> An SSH username extracted with a custom regex, for example when starting with a digit, may not work on the system where the user is trying to use it, correct? To me sounds like a mapping in a template in addition to this should maybe be the full solution if we describe this env var?

Most of the limitations are usually defined by `adduser` and `useradd` tools, a user starting with a digit will work, but you will need to pass flags like `--allow-badname`, `--allow-all-names`, or `--badname`. So it is possible to create a user that starts with a digit or that contains UTF-8 characters, but as it is not recommended you might find incompatibility issues when using other tools (e.g. samba).

The decision to be made is probably if we want this as an environment variable, as a flag, or as a parameter in the `ca.json` at the authority level. I don't think it makes sense to set it as the provisioner level as it will mean more changes in other projects, e.g. `step ca provisioner add`, `linkedca`, ... Especially when this is a very special case.

hslatman:
I am OK with the env var approach, because this is a very specific case, and wouldn't provide a lot of value compared to the effort needed to add support for it in all places. I don't recall many other options being provided like this, so I understand the concern. Maybe the closest is the `--resolver` flag at CA startup, which could also be an option instead of the env var.

Removing the restriction sounds OK to me, if there's a way to inform/warn users about the potential issue(s) that it brings along. That could be just documentation, but a more contextual warning would be great. Users could create a template to map a problematic principal to a non-problematic one, in that case.

tashian:
I'm open to either removing the restriction, or adding this as a flag.

I think a flag would be a slightly better option than an environment variable. It makes invocation a little bit easier in different contexts like systemd or docker if you don’t have to mix both flags and env vars. And it self-documents.

PR #1481, opened 2023-07-19
Remove OIDC user regexp check

### LINKED ISSUE #1436: Allow SSH usernames to begin with a digit

maraino:
### Description 

This commit removes the regular expression check on OIDC usernames. Although using any character in a username is not recommended, creating and using them is possible. The tool `useradd` has the flag `--badname`, and `adduser` has `--allow-badname` and `--allow-all-names` to create new users with any character.

Moreover, it is possible to create any username with the rest of provisioners.

Fixes #1436

### DIFF
--- a/authority/provisioner/controller.go
+++ b/authority/provisioner/controller.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/x509"
 	"net/http"
-	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


maraino:
@hslatman @tashian This PR can replace the one with the custom regular expression, https://github.com/smallstep/certificates/pull/1478. What do you think?

tashian (approved):

hslatman:
Shouldn't this be behind a start up flag? It seems the check is now completely removed/disabled.

maraino:
> Shouldn't this be behind a start up flag? It seems the check is now completely removed/disabled.

At this moment, I don't see the point of this check. Something to take into account is that principals != usernames. You can configure SSH login a user with a principal `ʕ•ᴥ•ʔ` as any "valid" username.

hslatman (approved):

PR #1484, opened 2023-07-20
Add support for the disableSmallstepExtensions claim

### LINKED ISSUE #620: Sign certificates without the step provisioner extension
### Discussed in https://github.com/smallstep/certificates/discussions/615

<div type='discussions-op-text'>

<sup>Originally posted by **Kieren** June 22, 2021</sup>
In order to use leaf certificates issued by step ca for mTLS in a particular IoT application the certificates cannot have extraneous extensions like X509v3 Step Provisioner. Is there a way to configure step ca to not issue leaf certificates with this extension?

For example the Arduino MKR Wifi 1010 board has an ATECC508A cryptographic authentication chip which has hardware based certificate storage for client certificates. However due to the storage limitations of the chip the certificates are deconstructed into a compressed form for storage, where known values like subject fields are stripped and stored separately. Certificates issued via step ca with the Step Provisioner extension cannot be used as a drop-in in this application.</div>

maraino:
### Description

This commit adds a new claim to exclude the Smallstep provisioner extension from the generated certificates.

Fixes #620


authority/provisioner/sign_options.go
```diff
@@ -443,7 +444,18 @@ func newProvisionerExtensionOption(typ Type, name, credentialID string, keyValue
 	}
 }
 
+// WithControllerOptions returns the provisionerExtensionOption options from the
+// controller. Currently only the claim DisableSmallstepExtensions is used.
```
hslatman, line 448:
```suggestion
// WithControllerOptions updates the provisionerExtensionOption with options from the
// controller. Currently only the DisableSmallstepExtensions claim is used.
```

authority/provisioner/claims.go
```diff
@@ -110,6 +115,15 @@ func (c *Claimer) IsDisableRenewal() bool {
...
 
+// IsDisableSmallstepExtensions returns if the Smallstep extensions, like the
+// provisioner extension, should be excluded from the certificate.
+func (c *Claimer) IsDisableSmallstepExtensions() bool {
+	if c.claims == nil || c.claims.DisableSmallstepExtensions == nil {
+		return *c.global.DisableSmallstepExtensions
+	}
+	return *c.claims.DisableSmallstepExtensions
+}
+
```
hslatman, line 126:
```suggestion
// IsDisableSmallstepExtensions returns whether Smallstep extensions, such as
// the provisioner extension, should be excluded from the certificate.
func (c *Claimer) IsDisableSmallstepExtensions() bool {
	if c.claims == nil || c.claims.DisableSmallstepExtensions == nil {
		return *c.global.DisableSmallstepExtensions
	}
	return *c.claims.DisableSmallstepExtensions
}

```
Generally I prefer `ShouldX` instead of `IsX` for "active function names". `IsX` would be better if it were `AreSmallstepExtensionsDisabled`.

authority/config/config.go
```diff
@@ -35,6 +35,9 @@ var (
 	// DefaultEnableSSHCA enable SSH CA features per provisioner or globally
 	// for all provisioners.
 	DefaultEnableSSHCA = false
+	// DefaultDisableSmallstepExtensions disables the Smallstep extensions in
+	// the certificate.
+	DefaultDisableSmallstepExtensions = false
```
hslatman, line 40:
```suggestion
	// DefaultDisableSmallstepExtensions is the default value for the DisableSmallstepExtensions claim. 
	DefaultDisableSmallstepExtensions = false
```

authority/provisioner/sign_options.go
hslatman: It can be a bit confusing when both options and claims are used to refer to the same / similar thing(s). I know both have been around for a while, but maybe we need to (start) set(ting) things straight now. 

The `DisableSmallstepExtensions` boolean sounds like a plain option to me. Claims, to me, are things that end up in JWTs. Is this the proper distinction to make?

maraino: Fixed with 3790bd01, I've changed "claim" with "provisioner claim" in that comment.

authority/provisioner/claims.go
maraino: Fixed with https://github.com/smallstep/certificates/commit/3790bd01a496452bf1e13b534357fd9f215f2840

authority/config/config.go
maraino: Fixed with https://github.com/smallstep/certificates/commit/3790bd01a496452bf1e13b534357fd9f215f2840

hslatman (approved):

PR #1495, opened 2023-07-28
Return more detailed errors to ACME clients using `device-attest-01` 

hslatman:


acme/challenge.go
maraino, line 1: By the way, badAttestationStatement was added to the draft a few days ago. See https://github.com/brandonweeks/draft-bweeks-acme-device-attest/commit/fdc5aac747c4ea3b0b2ca8d01fc74fe3f6f5f9d9

hslatman: That's great 😄 

We have a slightly different generic description, but I think that's fine, as client are expected to understand the error URIs; not the (standardized) descriptions. 

I included that generic description for the failed attestations, because I think that makes things clearer when errors are inspected manually on the client side.

acme/challenge.go
```diff
@@ -411,7 +418,12 @@ func deviceAttest01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose
 		//
 		// Note: We might want to use an external service for this.
 		if data.UDID != ch.Value && data.SerialNumber != ch.Value {
-			return storeError(ctx, db, ch, true, NewError(ErrorBadAttestationStatementType, "permanent identifier does not match"))
+			subproblem := NewSubproblemWithIdentifier(
+				ErrorRejectedIdentifierType,
+				Identifier{Type: "permanent-identifier", Value: ch.Value},
+				"challenge identifier %q doesn't match any of the attested hardware identifiers %s", ch.Value, []string{data.UDID, data.SerialNumber},
```
maraino, line 424: If you use %q, you can get those as `["udid" "serialNumber"]`

maraino: You can also do %q or %q

acme/challenge.go
```diff
@@ -411,7 +418,12 @@ func deviceAttest01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose
...
+			)
+			return storeError(ctx, db, ch, true, NewError(ErrorBadAttestationStatementType, "permanent identifier does not match").WithAdditionalErrorDetail().AddSubproblems(subproblem))
```
maraino, line 426: Instead of `WithAddinionalErrorDetails()` why not a new constructor like `NewDetailedError()`

acme/errors.go
```diff
@@ -293,16 +293,31 @@ type Subproblem struct {
 	Identifier *Identifier `json:"identifier,omitempty"`
 }
 
+// NewError creates a new Error.
+func NewError(pt ProblemType, msg string, args ...interface{}) *Error {
+	return newError(pt, errors.Errorf(msg, args...))
+}
```
maraino, line 299: Should this create a detailed error by default? We can also have a second constructor instead of WithAdditionalErrorDetail.

maraino (changes requested):

acme/challenge.go
hslatman: It just was my first idea and did fit in with the already existing fluent pattern for `AddSubproblems`. It also works for `Errors` from both `NewError` and `NewErrorISE`, although the additional detail is explicitly blocked for the latter, as a bit of an additional safety net.

acme/errors.go
hslatman: Maybe it makes sense to do so, but we would have to check all paths that return an `Error` to not return sensitive details. Not impossible, but not the thing I was trying to solve in this PR.

maraino: Something like NewDetailedError will be a new function. We can get replace just the ones with `WithAdditionalErrorDetail`

acme/challenge.go
maraino: `AddSubproblems` make more sense because the NewError has a fmt-like interface.

maraino (approved):
I'll approve this, but I think we should use NewDetailedError instead.

hslatman:
Had to fix merge conflict, so also updated to `NewDetailedError` in c952e9fc9de1f2d2f172871bb89cf567d733fde8.

maraino (approved):

PR #1502, opened 2023-08-04
Initial support for SCEP Polling

### LINKED ISSUE #1170: Does step-ca support SCEP manual mode?
## Hello!

## Issue details
Does step-ca support SCEP manual mode?

According to the SCEP RFC (https://www.rfc-editor.org/rfc/rfc8894), section 2.4 (https://www.rfc-editor.org/rfc/rfc8894#name-enrolment-authorisation) SCEP servers can optionally put SCEP clients in a pending mode:

_To perform the authorisation in manual mode, the client's request is placed in the PENDING state until the CA operator authorises or rejects it. Manual authorisation is used when the client has only a self-signed certificate that hasn't been previously authenticated by the CA and/or a challengePassword is not available. The SCEP CA MAY either reject unauthorised requests or mark them for manual authorisation according to CA policy._

## Why is this needed?
We're looking at a few options for CA servers, and some of them have implemented this functionality. It helps in situations where you have some trust, but not perfect trust in the end device's environment, and validation of the requests by a human is required. As per RFC 8894, a challenge password (as step-ca seems to use at the moment) can be paired with pending/manual mode.

adrian-alexander:
### **Description:**

This pull request adds SCEP polling mode which can be configured via ca.json. When enabled, the CA will return a PENDING response which will cause the client to poll the CA server. Fixes #1170 .

Two new databases have been created named `x509_csr` and `x509_certs_csr` to save the certificate requests that have been signed.

### DIFF
--- a/authority/config/config.go
+++ b/authority/config/config.go
@@ -82,10 +82,16 @@ type Config struct {
 	Templates        *templates.Templates `json:"templates,omitempty"`
 	CommonName       string               `json:"commonName,omitempty"`
 	CRL              *CRLConfig           `json:"crl,omitempty"`
+	Polling          *PollingConfig       `json:"polling,omitempty"`
 	SkipValidation   bool                 `json:"-"`
 
 	// Keeps record of the filename the Config is read from
 	loadedFromFilepath string
+}
+
+// PollingConfig represents config options for SCEP polling
+type PollingConfig struct {
+	Enabled bool `json:"enabled"`
 }
 
 // CRLConfig represents config options for CRL generation
@@ -95,6 +101,11 @@ type CRLConfig struct {
 	CacheDuration    *provisioner.Duration `json:"cacheDuration,omitempty"`
 	RenewPeriod      *provisioner.Duration `json:"renewPeriod,omitempty"`
 	IDPurl           string                `json:"idpURL,omitempty"`
+}
+
+// IsEnabled returns if polling is enabled.
+func (c *PollingConfig) IsEnabled() bool {
+	return c != nil && c.Enabled
 }
 
 // IsEnabled returns if the CRL is enabled.
--- a/authority/tls.go
+++ b/authority/tls.go
@@ -289,6 +289,14 @@ func (a *Authority) Sign(csr *x509.CertificateRequest, signOpts provisioner.Sign
 		}
 	}
 
+	// Store certificate and certificate request in the db.
+	if err = a.storeCertificateAndRequest(csr, fullchain); err != nil {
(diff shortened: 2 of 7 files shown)


hslatman:
Hey @adrian-alexander,

Thank you for this contribution. I'll have a look at it. We'll also discuss it in our next open source triage meeting.

Do you have a specific use case for which you need `CertPoll` support? Or some specific SCEP client that requires it? So far our certificate signing operations have been synchronous, in-line with the HTTP request, meaning that we always just sent `SUCCESS` or `FAILURE` immediately. This works because the signed certificate is either available, or not, because the signing operation doesn't happen asynchronously / in the background.

adrian-alexander:
> Hey @adrian-alexander,
> 
> Thank you for this contribution. I'll have a look at it. We'll also discuss it in our next open source triage meeting.
> 
> Do you have a specific use case for which you need `CertPoll` support? Or maybe some specific SCEP client that requires it? So far our certificate signing operations have been synchronous, in-line with the HTTP request, meaning that we always just sent `SUCCESS` or `FAILURE` immediately. This works because the signed certificate is either available, or not, because the signing operation doesn't happen asynchronously / in the background.

Hey @hslatman,

My team has been looking for a CA with SCEP polling to replace our in-house CA. Our use case is untrusted networks and deployments. We need an admin to manually approve cert requests from technicians at the sites.

hslatman:
Sounds like a legit use case 🙂 

I haven't looked into the code in detail yet, but from a quick skim I don't see changes to how the `auth.SignCSR` method is called, especially in regards to asynchronously signing the certificate. For your use case, there's a need to provide some way to manually approve the CSR that you stored in the table, and then turn it into a certificate, storing that state in the the table too. What do you expect that integration/workflow work/look like?

Since you referred to https://github.com/smallstep/certificates/issues/1170, I guess you're aware of the fact that we have support for (manual) approval workflows in our product. Ideally the flow you contributed here would have to be interoperable with that part of our stack. Besides that, it might be an option to provide another method. One way could be to make it work with webhooks.

Are you open to talk more about your use case with a colleague of mine?

adrian-alexander:
> What do you expect that integration/workflow work/look like?

For testing, I used `sscep` to send the request to the CA. While it was polling, I signed the request with `step ca sign`, which signs the CSR as well as adds the certificate and the matching CSR to the database. I went with this route so whether the CSR were signed with the basic-client or through the CLI, the certificate and CSR would still get added to the database.

> Are you open to talk more about your use case with a colleague of mine?

I'm actually part of the same team as the guy in #1170 . What he discussed in his meeting with you is the same thing as my use case 😃

PR #1516, opened 2023-08-25
Include webhook response in error message.

joshdrake:

Not having visibility in to errors during sign requests when a webhook is used.

Helps with troubleshooting eg: device attestation workflows.

💔Thank you!


areed (approved):

maraino (approved):
lgtm, it won't look as expected if the response is not text.

PR #1535, opened 2023-09-15
Skip challenge feature

kadirgun:
#### Name of feature:
Skip challenge validation if provisioner challenges empty

DNS configuration was required for personal use on local machines. With this feature, projects can be run over https using only step-ca and traefik.

💔Thank you!


hslatman:
@kadirgun I'm going to close this PR, as we're not going to merge it as-is. A mode in which challenge validation is skipped is still something I can see being used in some specific use cases, so I'll keep the issue open.

PR #1542, opened 2023-09-19
Propagate request ID when webhook requests are made

hslatman:
This PR adds support for propagating the request ID to the webhook receiver. The request ID is sent in the `X-Request-ID` header and uses the value that is originally created by and sent by the client in the `X-Smallstep-ID` header. If it's not set, it'll be created by the CA.

The request ID is propagated through a `context.Context`. This required adding the `SignWithContext` method to the (TLS) authority and interfaces related to the authority. By adding the method, we can deprecate the existing `Sign` method, and soon default to only using the context-aware method. I think the latter makes sense, because there are already some methods that take `ctx` as the first argument, so we can make that the standard way.

~Technically, the request ID is used more like a tracing ID in this case, so I'm fine with using a different header name. `X-Smallstep-ID` could be used for that? I chose `X-Request-ID` because the value is reported as `request-id` in the logs, so it's easier to ascertain that these are the same values and can be used for correlating/tracing requests to the CA and onwards to the webhook receiver.~

Currently this relies on the existing logic for setting the request ID based on the `logging` package, which takes the ID to forward from the `X-Smallstep-Id` request header. Will follow up with a more end-to-end implementation and testing for the `X-Request-ID` header.


dopey (approved):
love this

PR #1544, opened 2023-09-21
Add SCEP issuance notification webhook

hslatman:


authority/admin/api/webhook.go
```diff
@@ -57,7 +57,7 @@ func validateWebhook(webhook *linkedca.Webhook) error {
 
 	// kind
 	switch webhook.Kind {
-	case linkedca.Webhook_ENRICHING, linkedca.Webhook_AUTHORIZING, linkedca.Webhook_SCEPCHALLENGE:
+	case linkedca.Webhook_ENRICHING, linkedca.Webhook_AUTHORIZING, linkedca.Webhook_SCEPCHALLENGE, linkedca.Webhook_NOTIFYING:
```
maraino, line 60: We could add automatic support for future types adding something like:

```go
if _, ok := linkedca.Webhook_Kind_name[int32(webhook.Kind)]; !ok || webhook.Kind == 0 {
        return admin.NewError(admin.ErrorBadRequestType, "webhook kind %q is invalid", webhook.Kind)
}
```

authority/provisioner/scep.go
```diff
@@ -163,6 +165,71 @@ func (c *challengeValidationController) Validate(ctx context.Context, csr *x509.
...
+		}
+		req.X509Certificate.Raw = cert.Raw // adding the full certificate DER bytes
+		req.SCEPTransactionID = transactionID
+		resp, err := wh.DoWithContext(ctx, c.client, req, nil)
+		if err != nil {
+			return fmt.Errorf("failed executing webhook request: %w", err)
+		}
+		if !resp.Allow { // TODO(hs): different response for notifying?
+			return ErrSCEPNotificationFailed // return early
+		}
```
maraino, line 206: Ok, but the use of this Allow is quite weird. I don't see a use case for a webhook returning 2xx and Allow set to false.

scep/api/api.go
```diff
@@ -333,9 +333,21 @@ func PKIOperation(ctx context.Context, req request) (Response, error) {
 
 	certRep, err := auth.SignCSR(ctx, csr, msg)
 	if err != nil {
+		// default to ERROR_INTERNAL_ERROR: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/18d8fbe8-a967-4f1c-ae50-99ca8e491d2d
+		errorCode := 0x0000054F
```
maraino, line 337: This is very specific, I wonder if it makes more sense to use a more specific webhook type.

scep/context.go
```diff
@@ -0,0 +1 @@
+package scep
```
maraino, line 1: Empty file?

maraino (commented):
Looks ok. There's one almost "empty" file committed. And I wonder if the use of Win32 error codes is too specific but the notifying type looks quite generic.

scep/context.go
hslatman: Removed in ba72710e2d11295287a9816a6d3dbd2b544ff075.

scep/api/api.go
hslatman: I've set this to 0 in ba72710e2d11295287a9816a6d3dbd2b544ff075 for now to not overcomplicate things. A more specific webhook type can work, but having this more generic type is more suitable at the moment. Because the error code can be context specific, I make it the responsibility of the webhook receiver to handle those based on the error message (for now).

authority/admin/api/webhook.go
hslatman: Done in ba72710e2d11295287a9816a6d3dbd2b544ff075.

authority/provisioner/scep.go
hslatman: I've changed this to not rely on the response body at all. Any OK-ish response will now result in the notification succeeding, based on the existing webhook execution logic.

maraino (approved):
lgtm.

TIL, wrapping twice implements `interface{ Unwrap() []error }`, it does not implements `interface{ Unwrap() error }`. 

See https://go.dev/play/p/gCm8j14Jf79

PR #1553, opened 2023-09-25
Change SCEP authority initialization

hslatman:


authority/options.go
```diff
@@ -205,6 +206,23 @@ func WithX509SignerFunc(fn func() ([]*x509.Certificate, crypto.Signer, error)) O
...
 }
 
+// func WithSCEPOptions(crt *x509.Certificate, s crypto.Signer, d crypto.Decrypter) Option {
+// 	return func(a *Authority) error {
+// 		a.scepCertificate = crt
+// 		a.scepSigner = s
+// 		a.scepDecrypter = d
+// 		return nil
+// 	}
+// }
```
maraino, line 216: Remove me.

authority/options.go
```diff
@@ -205,6 +206,23 @@ func WithX509SignerFunc(fn func() ([]*x509.Certificate, crypto.Signer, error)) O
...
+
+func WithFullSCEPOptions(options *scep.Options) Option {
```
maraino, line 218: Add comment and add something like:
```go
// This feature is EXPERIMENTAL and might change at any time.
```

authority/authority.go
```diff
@@ -671,54 +674,57 @@ func (a *Authority) init() error {
...
+					// only pass the decrypter down when it was successfully created,
+					// meaning it's an RSA key, and `CreateDecrypter` did not fail.
+					options.Decrypter = decrypter
+					options.DecrypterCert = options.Intermediates[0]
+				}
+			}
+
+			// provide the current SCEP provisioner names, so that the provisioners
+			// can be validated when the CA is started.
+			options.SCEPProvisionerNames = a.getSCEPProvisionerNames()
```
maraino, line 712: I would set this to `scepOptions` too if not provided.

maraino (changes requested):

authority/authority.go
hslatman: Fixed in c0fbace88227fcfeb5f92415fa6158e4859638b3. Is that what you had in mind?

authority/options.go
hslatman: Fixed in c0fbace88227fcfeb5f92415fa6158e4859638b3.

authority/options.go
hslatman: Fixed in c0fbace88227fcfeb5f92415fa6158e4859638b3.

maraino (approved):

PR #1566, opened 2023-10-02
Fix `step ca init` with duplicate provisioner name (`--provisioner acme --acme`)

hslatman:
This PR fixes https://github.com/smallstep/cli/issues/1018 (after CLI is updated with the commits from this branch).

@tashian I simply suffixed `acme` with `-1` to stay close to the original name. Any other suggestions?


maraino (approved):
I guess it is ok, but I wonder why one somebody would want the JWK provisioner to be named acme or sshpop.

hslatman:
Agree, but it did happen 😅 

It was a bit of an unfortunate choice, likely stemming from going through the docs, and maybe reading a bit about the option to name your first provisioner with the `--provisioner` option. If you do this for the first time, you might not know the first provisioner is actually a JWK one, and that `--provisioner` is only used to name that one; not the `--acme` one. 

That said, maybe the docs aren't entirely clear about what the options are used for, so maybe clarifying there is also needed.

tashian:
As a new user, one might have the mental model of "If I supply `--acme` and `--provisioner-name`, then I will be naming the acme provisioner that gets created."

It's a reasonable mental model for a first time user to have, and I think this is a good start for a fix.

I don't totally understand the value of the `--provisioner-name` flag, so one idea is to remove or hide `--provisioner-name` from the help text, and just make it `admin` always.

And if they want to change it, they can use `step ca provisioner update --name` after init.

But if the flag is important to keep around, my next suggestion would be to change the flag to `--admin-provisioner-name` for clarity.

maraino:
By default we're still defaulting to use the ca.json for configuration, there's not a real "admin" in that case.

PR #1570, opened 2023-10-04
Improve SCEP request handling

hslatman:
This PR:

*  Improves the logic for handling SCEP requests
* Adds a workaround for macOS SCEP clients sending weird GET PKIOperation messages

Apparently the macOS SCEP client sends a SCEP message in the query
that's not fully escaped. Only the base64 padding is escaped, the
'+' and '/' characters aren't.

This is a bit of a special case, because the macOS SCEP client
will default to using HTTP POST for the PKIOperation. But if the
CA is configured without the POSTPKIOperation capability, the
macOS SCEP client will use HTTP GET instead. This behavior might
be the same on iOS.

### DIFF
--- a/scep/api/api.go
+++ b/scep/api/api.go
@@ -171,10 +171,7 @@ func decodeRequest(r *http.Request) (request, error) {
 			}, nil
 		case opnPKIOperation:
 			message := query.Get("message")
-			if message == "" {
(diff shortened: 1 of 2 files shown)


maraino (approved):
lgtm

PR #1572, opened 2023-10-06
Allow for identity certificate signing (in sshSign) by skipping validators

dopey:
- skip urisValidator for identity certificate signing. Implemented by building the validator with the context in a hacky way.


maraino (approved):
lgtm, but we need to refactor this, adding something like an AuthirizeIdentity method in the provisioners that can be used to verify and create the identity certificate.

PR #1609, opened 2023-10-31
Change CommonName validator in JWK

maraino:
### Description

This commit changes the common name validator in the JWK provisioner to accept either the token subject or any of the sans in the token.

This PR allows admin tokens on RA to work with the changes in https://github.com/smallstep/certificates/pull/1608


hslatman (approved):
It does result in more values being accepted as the CN, but they've been validated before, so I guess it's OK.

PR #1640, opened 2023-12-05
Update Dockerfile.hsm to use Debian bookworm

tashian:
I've tested that `step-ca` builds and starts up properly using the `debian:bookworm` base images for `step` and `step-kms-plugin`, and the `golang:bookworm` builder image

Note: https://github.com/smallstep/step-kms-plugin/pull/121 and https://github.com/smallstep/cli/pull/1080 will need to be merged and published to Docker Hub before this can go out.


maraino (approved):
lgtm 📖 🪱

maraino (approved):

PR #1666, opened 2024-01-08
Wire ACME extensions

hslatman:


acme/challenge.go
```diff
@@ -342,6 +354,258 @@ func dns01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSONWebK
...
+	// Update and store the challenge.
+	ch.Status = StatusValid
+	ch.Error = nil
+	ch.ValidatedAt = clock.Now().Format(time.RFC3339)
+
+	if err = db.UpdateChallenge(ctx, ch); err != nil {
+		return WrapErrorISE(err, "error updating challenge")
+	}
+
+	parsedIDToken, err := jwt.ParseSigned(wireChallengePayload.IDToken)
```
maraino, line 426:
```suggestion
	parsedIDToken, err := jose.ParseSigned(wireChallengePayload.IDToken)
```

acme/challenge.go
```diff
@@ -342,6 +354,258 @@ func dns01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSONWebK
...
+func wireDPOP01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSONWebKey, payload []byte) error {
+	prov, ok := ProvisionerFromContext(ctx)
+	if !ok {
+		return NewErrorISE("missing provisioner")
+	}
+
+	rawKid, thumbprintErr := jwk.Thumbprint(crypto.SHA256)
+	if thumbprintErr != nil {
+		return storeError(ctx, db, ch, false, WrapError(ErrorServerInternalType, thumbprintErr, "failed to compute JWK thumbprint"))
+	}
```
maraino, line 462: It's optional, but this looks more natural.
```suggestion
	rawKid, err := jwk.Thumbprint(crypto.SHA256)
	if err != nil {
		return storeError(ctx, db, ch, false, WrapError(ErrorServerInternalType, err, "failed to compute JWK thumbprint"))
	}
```

acme/challenge.go
```diff
@@ -342,6 +354,258 @@ func dns01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSONWebK
...
+
+	kid := base64.RawURLEncoding.EncodeToString(rawKid)
+
+	dpopOptions := prov.GetOptions().GetDPOPOptions()
+	key := dpopOptions.GetSigningKey()
+
+	var wireChallengePayload WireChallengePayload
+	err := json.Unmarshal(payload, &wireChallengePayload)
+	if err != nil {
```
maraino, line 471: Per before:
```suggestion
	if err := json.Unmarshal(payload, &wireChallengePayload); err != nil {
```

or 

```go
err = json.Unmarshal(payload, &wireChallengePayload)
```

acme/challenge.go
```diff
@@ -342,6 +354,258 @@ func dns01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSONWebK
...
+		return storeError(ctx, db, ch, false, WrapError(ErrorRejectedIdentifierType, err,
+			"error unmarshalling Wire challenge payload"))
+	}
+
+	file, err := os.CreateTemp(os.TempDir(), "acme-validate-challenge-pubkey-")
+	if err != nil {
+		return WrapErrorISE(err, "temporary file could not be created")
+	}
+	defer file.Close()
+	defer os.Remove(file.Name())
```
maraino, line 481: You want this the other way around:
```suggestion
	defer os.Remove(file.Name())
	defer file.Close()
```

This program will print `b` and then `a`
```go
package main

import "fmt"

func a() {
	fmt.Println("a")
}

func b() {
	fmt.Println("b")
}

func main() {
	defer a()
	defer b()
}
```

acme/challenge.go
```diff
@@ -342,6 +354,258 @@ func dns01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSONWebK
...
+
+	buf := bytes.NewBuffer(nil)
+	buf.WriteString(key)
+
+	n, err := file.Write(buf.Bytes())
```
maraino, line 486: Seems unnecessary:
```suggestion
	buf := bytes.NewBuffer(nil)
	buf.WriteString(key)

	n, err := file.Write([]byte(key))
```

Btw this also exists `bytes.NewBufferString("foo")`

acme/challenge.go
```diff
@@ -342,6 +354,258 @@ func dns01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSONWebK
...
+	if err != nil {
+		return WrapErrorISE(err, "Failed writing signature key to temp file")
+	}
+	if n != buf.Len() {
+		return WrapErrorISE(err, "expected to write %d characters to the key file, got %d", buf.Len(), n)
+	}
```
maraino, line 492: If you commit the above suggestion this will be:
```suggestion
	if s := len(key); n != s {
		return WrapErrorISE(err, "expected to write %d characters to the key file, got %d", s, n)
	}
```

maraino (changes requested):
WIP review

authority/provisioner/dpop_options.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+type DPOPOptions struct {
+	// ValidationExecPath is the name of the executable to call for DPOP
+	// validation.
+	ValidationExecPath string `json:"validation-exec-path,omitempty"`
+	// Backend signing key for DPoP access token
+	SigningKey string `json:"key"`
+	// URI template acme client must call to fetch the DPoP challenge proof (an access token from wire-server)
+	DpopTarget string `json:"dpop-target"`
```
maraino, line 17: Here and before, we usually use cammelCase for JSON properties.

authority/provisioner/dpop_options.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	return o.DpopTarget
+}
+
+func (o *DPOPOptions) GetTarget(deviceID string) (string, error) {
+	if o == nil {
+		return "", errors.New("misconfigured target template configuration")
+	}
+	targetTemplate := o.GetDPOPTarget()
+	tmpl, err := template.New("DeviceId").Parse(targetTemplate)
```
maraino, line 46: I suppose sprig is not required here.

authority/provisioner/oidc_options.go
```diff
@@ -0,0 +1,90 @@
...
+
+type ConfigJSON struct {
+	ClientID                   string           `json:"client-id,omitempty"`
+	SupportedSigningAlgs       []string         `json:"support-signing-algs,omitempty"`
+	SkipClientIDCheck          bool             `json:"-"`
+	SkipExpiryCheck            bool             `json:"-"`
+	SkipIssuerCheck            bool             `json:"-"`
+	Now                        func() time.Time `json:"-"`
+	InsecureSkipSignatureCheck bool             `json:"-"`
+}
```
maraino, line 32: This looks like configuration options that a user should set; we generally use cammelCase for this.

authority/provisioner/oidc_options.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type ProviderJSON struct {
+	IssuerURL   string   `json:"issuer,omitempty"`
+	AuthURL     string   `json:"authorization_endpoint,omitempty"`
+	TokenURL    string   `json:"token_endpoint,omitempty"`
+	JWKSURL     string   `json:"jwks_uri,omitempty"`
+	UserInfoURL string   `json:"userinfo_endpoint,omitempty"`
+	Algorithms  []string `json:"id_token_signing_alg_values_supported,omitempty"`
+}
```
maraino, line 22: It would be nice if this is available in something like a `/.well-known/openid-configuration`

wire/id.go
```diff
@@ -0,0 +1,49 @@
...
+	parts := strings.SplitN(fullUsername, "!", 2)
+	if len(parts) != 2 {
+		return ClientID{}, fmt.Errorf("invalid clientID %q", fullUsername)
+	}
+	return ClientID{
+		Username: parts[0],
+		DeviceID: parts[1],
+		Domain:   clientIDURI.Host,
+	}, nil
+}
```
maraino, line 49: We are not validating the schema here, but we do it in `acme/api/order.go`

maraino (changes requested):

go.mod
```diff
@@ -146,5 +147,7 @@ require (
 	google.golang.org/genproto v0.0.0-20231211222908-989df2bf70f3 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20231211222908-989df2bf70f3 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20231212172506-995d672761c0 // indirect
-	gopkg.in/yaml.v3 v3.0.1 // indirect
+	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
+	gopkg.in/square/go-jose.v2 v2.6.0
```
maraino, line 151: We should not include go-jose.v2, we should use `go.step.sm/crypto` if possible, alternativelly we can also use `github.com/go-jose/go-jose/v3`

maraino (changes requested):

go.mod
hslatman: Not relying on it directly anymore, but it's still indirect through https://github.com/coreos/go-oidc/.

wire/id.go
hslatman: Good suggestion; I've moved the ClientID parsing to here in 70a2f431fa896624b6dafa375a47f454b3bde1a6

authority/provisioner/oidc_options.go
hslatman: Agree, but that's something Wire would have to implement/provide, in that case, I think? It could be an (optional) improvement after discussing it with them?

authority/provisioner/oidc_options.go
hslatman: Yes, agree. I've held off on changing these so far because that would break the e2e testing and would need changes on their side and documentation. Will be in a follow-up.

authority/provisioner/dpop_options.go
hslatman: Will update this in follow up. Didn't do so to not break the e2e test.

authority/provisioner/dpop_options.go
```diff
@@ -0,0 +1,55 @@
...
+	"bytes"
+	"errors"
+	"fmt"
+	"text/template"
+)
+
+type DPOPOptions struct {
+	// ValidationExecPath is the name of the executable to call for DPOP
+	// validation.
+	ValidationExecPath string `json:"validation-exec-path,omitempty"`
```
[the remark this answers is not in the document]
hslatman: Yes, in progress here: https://github.com/smallstep/certificates/pull/1670.

authority/provisioner/dpop_options.go
```diff
@@ -0,0 +1,55 @@
...
+	// Backend signing key for DPoP access token
+	SigningKey string `json:"key"`
```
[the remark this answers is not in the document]
hslatman: It's currently a plain PEM string. We could propose base64 encoding that, similar to what we do for certs in other cases. Would be a breaking change in their current configuration utility, but I don't think a big issue.

I've suggested a JWK in chat. They chose this as the simplest solution.

acme/db/nosql/challenge.go
```diff
@@ -19,6 +19,7 @@ type dbChallenge struct {
 	Status      acme.Status        `json:"status"`
 	Token       string             `json:"token"`
 	Value       string             `json:"value"`
+	Target      string             `json:"target,omitempty"`
```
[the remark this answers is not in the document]
hslatman: The `Target` is returned to the ACME client (if it's set) in the `GetChallenge` call. Haven't looked into what the Wire ACME client does with it specifically (yet).

acme/db.go
```diff
@@ -52,7 +52,13 @@ type DB interface {
 	CreateOrder(ctx context.Context, o *Order) error
 	GetOrder(ctx context.Context, id string) (*Order, error)
 	GetOrdersByAccountID(ctx context.Context, accountID string) ([]string, error)
+	GetAllOrdersByAccountID(ctx context.Context, accountID string) ([]string, error)
 	UpdateOrder(ctx context.Context, o *Order) error
+
+	CreateDpopToken(ctx context.Context, orderID string, dpop map[string]interface{}) error
+	GetDpopToken(ctx context.Context, orderID string) (map[string]interface{}, error)
+	CreateOidcToken(ctx context.Context, orderID string, idToken map[string]interface{}) error
+	GetOidcToken(ctx context.Context, orderID string) (map[string]interface{}, error)
```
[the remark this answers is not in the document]
hslatman: I've made this a TODO. Should it be a different interface, or the same interface, but dependent on a build tag? If we create a new interface and call that `WireDB`, for example, that `WireDB` would be used in all places, so that didn't look like a nice solution to me.

acme/challenge.go
hslatman: This will be gone with https://github.com/smallstep/certificates/pull/1670

acme/challenge.go
hslatman: This will be gone with https://github.com/smallstep/certificates/pull/1670

acme/challenge.go
hslatman: This will be gone with https://github.com/smallstep/certificates/pull/1670

acme/challenge.go
hslatman: Fixed in https://github.com/smallstep/certificates/pull/1666/commits/70a2f431fa896624b6dafa375a47f454b3bde1a6

acme/challenge.go
hslatman: Fixed in https://github.com/smallstep/certificates/pull/1666/commits/70a2f431fa896624b6dafa375a47f454b3bde1a6.

acme/challenge.go
hslatman: Fixed in https://github.com/smallstep/certificates/pull/1666/commits/70a2f431fa896624b6dafa375a47f454b3bde1a6

acme/challenge.go
```diff
@@ -20,23 +21,25 @@ import (
...
-	"golang.org/x/exp/slices"
-
 	"github.com/smallstep/go-attestation/attest"
-
 	"go.step.sm/crypto/jose"
 	"go.step.sm/crypto/keyutil"
 	"go.step.sm/crypto/pemutil"
 	"go.step.sm/crypto/x509util"
+	"golang.org/x/exp/slices"
+	"gopkg.in/square/go-jose.v2/jwt"
```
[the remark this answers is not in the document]
hslatman: Fixed in https://github.com/smallstep/certificates/pull/1666/commits/70a2f431fa896624b6dafa375a47f454b3bde1a6

acme/challenge.go
```diff
@@ -20,23 +21,25 @@ import (
...
 	"github.com/google/go-tpm/legacy/tpm2"
-	"golang.org/x/exp/slices"
-
 	"github.com/smallstep/go-attestation/attest"
-
 	"go.step.sm/crypto/jose"
 	"go.step.sm/crypto/keyutil"
 	"go.step.sm/crypto/pemutil"
 	"go.step.sm/crypto/x509util"
+	"golang.org/x/exp/slices"
```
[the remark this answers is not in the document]
hslatman: Yes, I used the std one before, but that failed in CI, so had to revert that change.

acme/api/order.go
```diff
@@ -48,13 +51,26 @@ func (n *NewOrderRequest) Validate() error {
...
+			if err != nil {
+				return acme.NewError(acme.ErrorMalformedType, "ID cannot be parsed")
+			}
+			clientIDURI, err := uri.Parse(orderValue.ClientID)
+			if err != nil {
+				return acme.NewError(acme.ErrorMalformedType, "invalid client ID, it's supposed to be a valid URI")
+			}
+			if clientIDURI.Scheme != "wireapp" {
+				return acme.NewError(acme.ErrorMalformedType, "invalid client ID scheme")
+			}
```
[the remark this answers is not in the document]
hslatman: I've moved some logic in https://github.com/smallstep/certificates/pull/1666/commits/70a2f431fa896624b6dafa375a47f454b3bde1a6

authority/provisioner/dpop_options.go
hslatman: Is this just a general remark? Doing something with the templates is one of the things that needs a bit more work, and using `sprig` could be useful to them. The goal is to dynamically get some information from the OIDC/DPoP token, and to possibly apply some transformation(s). So `sprig` support may need to be added here, I think.

acme/db.go
```diff
@@ -52,7 +52,13 @@ type DB interface {
 	CreateOrder(ctx context.Context, o *Order) error
 	GetOrder(ctx context.Context, id string) (*Order, error)
 	GetOrdersByAccountID(ctx context.Context, accountID string) ([]string, error)
+	GetAllOrdersByAccountID(ctx context.Context, accountID string) ([]string, error)
```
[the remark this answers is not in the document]
hslatman: See above.

authority/provisioner/oidc_options.go
hslatman: Changed this in https://github.com/smallstep/certificates/pull/1671.

authority/provisioner/dpop_options.go
hslatman: Changed this in https://github.com/smallstep/certificates/pull/1671

maraino (approved):
This looks good, but I added a few comments about:
1. One related to error messages
2. Using `net/url` instead of kms/uri
3. Split the db interface on two

I think it would be nice to fix 2 and 3 soon.

hslatman:
@maraino about the DB interface: I tried it before, but without changing a lot more, you'll end up with `WireDB` being used as the type in every struct, function and method that has to operate on the current `DB` interface, which I considered worse than just having these methods defined in the interface, and would basically result in what exists today.

This would've been a nice solution, but isn't allowed in Go (today, at least; [ref](https://github.com/golang/go/issues/45346#issuecomment-862505803)):

```golang
type DB interface {
    // DB methods
}
type WireDB interface {
    DB 
    // Wire methods
}
type NewDB interface {
    DB | WireDB
}
```

Alternatives would be:

- Have the top level `DB` interface provide all methods, as it does now, and check at runtime that all(/only) the required methods are available, which _could_ be based on some conditional logic taking into account the enabled challenge types. But that way it would still take the implementer to at least define the method in their implementation, so it's not a great solution.
- Make the methods in the `DB` interface depend on a build tag.
- Others ...?

It's not the cleanest solution, but implementers could always return some "not implemented" error. It's not as strong of a guarantee as having an interface, but it works.

acme/challenge.go
```diff
@@ -377,6 +387,387 @@ func dns01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSONWebK
...
+			"error verifying ID token signature"))
+	}
+
+	var claims struct {
+		Name         string `json:"preferred_username,omitempty"`
+		Handle       string `json:"name"`
+		Issuer       string `json:"iss,omitempty"`
+		GivenName    string `json:"given_name,omitempty"`
+		KeyAuth      string `json:"keyauth"`
+		ACMEAudience string `json:"acme_aud,omitempty"`
```
maraino, line 437: `omitempty` tags are not used on unmarshaling.

acme/challenge.go
```diff
@@ -377,6 +387,387 @@ func dns01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSONWebK
...
+		return WrapErrorISE(err, "could not retrieve current order by account id")
+	}
+	if len(orders) == 0 {
+		return NewErrorISE("there are not enough orders for this account for this custom OIDC challenge")
+	}
+
+	order := orders[len(orders)-1]
+	if err := db.CreateOidcToken(ctx, order, transformedIDToken); err != nil {
+		return WrapErrorISE(err, "failed storing OIDC id token")
+	}
```
maraino, line 486: Although we already discussed this, these methods are problematic. Without implementing them, we won't be able to upgrade the certificates' version.

We should split this into a different interface and return an error if the database doesn't implement it.

We can change this later, but not too late. This should not be hard to change as this method is probably called from `Challenge.Validate`

acme/challenge.go
```diff
@@ -377,6 +387,387 @@ func dns01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSONWebK
...
+	orders, err := db.GetAllOrdersByAccountID(ctx, ch.AccountID)
+	if err != nil {
+		return WrapErrorISE(err, "could not find current order by account id")
+	}
+	if len(orders) == 0 {
+		return NewErrorISE("there are not enough orders for this account for this custom OIDC challenge")
+	}
+
+	order := orders[len(orders)-1]
+	if err := db.CreateDpopToken(ctx, order, map[string]any(*dpop)); err != nil {
```
maraino, line 598: Same issue here.

acme/wire/id.go
```diff
@@ -0,0 +1,91 @@
...
+	Domain string `json:"domain,omitempty"`
+	Handle string `json:"handle,omitempty"`
+}
+
+type DeviceID struct {
+	Name     string `json:"name,omitempty"`
+	Domain   string `json:"domain,omitempty"`
+	ClientID string `json:"client-id,omitempty"`
+	Handle   string `json:"handle,omitempty"`
+}
```
maraino, line 22: It's weird having `omitempty` tags when most of those values are required later.

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	cloud.google.com/go/longrunning v0.5.11
 	cloud.google.com/go/security v1.17.4
 	github.com/Masterminds/sprig/v3 v3.2.3
+	github.com/coreos/go-oidc/v3 v3.4.0
```
maraino, line 9: It is not the latest version, but I guess ok.

maraino (approved):
I've added a few comments, but it is generally ok. We need to figure out the WireDB thing sooner rather than later.

PR #1669, opened 2024-01-10
Implementation of the Prometheus endpoint

azazeal:
This PR implements the `http://{metricsAddress}/metrics` endpoint.


authority/authority.go
```diff
@@ -958,3 +963,40 @@ func (a *Authority) startCRLGenerator() error {
 
 	return nil
 }
+
+//nolint:gocritic // used in defered statements
+func (a *Authority) incrProvisionerCounter(prov *provisioner.Interface, err *error, count func(Meter, string, bool)) {
```
maraino, line 968: This is an interface
```suggestion
func (a *Authority) incrProvisionerCounter(prov provisioner.Interface, err *error, count func(Meter, string, bool)) {
```

azazeal: There's pointers there as this is called only via `defer` statements. I understand it's an interface.

authority/config/config.go
```diff
@@ -327,6 +328,12 @@ func (c *Config) Validate() error {
 		return errors.Errorf("invalid address %s", c.Address)
 	}
 
+	if addr := c.MetricsAddress; addr != "" {
+		if _, _, err := net.SplitHostPort(addr); err != nil {
+			return errors.Errorf("invalid metrics address %s", c.Address)
```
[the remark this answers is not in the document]
azazeal:
```go
// Validate address (a port is required)
if _, _, err := net.SplitHostPort(c.Address); err != nil {
	return errors.Errorf("invalid address %s", c.Address)
}
```

:question:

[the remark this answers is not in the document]
hslatman: It is generally nicer to have user input demarcated.

authority/ssh.go
```diff
@@ -146,22 +146,24 @@ func (a *Authority) GetSSHBastion(ctx context.Context, user, hostname string) (*
 }
 
 // SignSSH creates a signed SSH certificate with the given public key and options.
-func (a *Authority) SignSSH(_ context.Context, key ssh.PublicKey, opts provisioner.SignSSHOptions, signOpts ...provisioner.SignOption) (*ssh.Certificate, error) {
+func (a *Authority) SignSSH(_ context.Context, key ssh.PublicKey, opts provisioner.SignSSHOptions, signOpts ...provisioner.SignOption) (cert *ssh.Certificate, err error) {
```
maraino, line 149: After looking at this package's dependents, I am willing to accept a breaking change that accepts a provisioner as a parameter or `Sign` and `SignSSH`. We can add the provisioner as a return value in `authority.Authorize`, and just pass it.

I'm not worried about breaking changes, but an option that won't cause less breaking changes would be to store the provisioner in the context, although we will need to add a context to `Sign` or just create the `SignContext` method. 

In any case, I'm ok to clean these, and have always a context and a provisioner parameter.

internal/metrix/meter.go
```diff
@@ -0,0 +1,190 @@
+// Package metrix implements stats-related functionality.
+package metrix
+
+import (
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
```
maraino, line 10: Would using OpenTelemetry be more flexible and will allow us to add more things in the future?

internal/metrix/meter.go
```diff
@@ -0,0 +1,190 @@
...
+			"success",
+		),
+		authorized: newCounterVec(subsystem, "authorized_total", "Number of authorizing webhooks called",
+			"provisioner",
+			"success",
+		),
+		enriched: newCounterVec(subsystem, "enriched_total", "Number of enriching webhooks called",
+			"provisioner",
+			"success",
+		),
```
maraino, line 168: I would rename those counters. The `authorized_total` especially can be confusing because the CA also has an authorization process, and the webhook authorization usage is very niche. Just prepending `webhook_` would be enough.

authority/ssh.go
```diff
@@ -146,22 +146,24 @@ func (a *Authority) GetSSHBastion(ctx context.Context, user, hostname string) (*
...
+	var prov provisioner.Interface
+	defer a.incrProvisionerCounter(&prov, &err, Meter.SSHSigned)
```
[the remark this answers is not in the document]
maraino: We can also clean this up by moving the signing code to internal methods and getting the error from the return values.

authority/ssh.go
maraino: If we do something like this we can kill a few issues at once:
```go
func (a *Authority) SignSSH(_ context.Context, key ssh.PublicKey, opts provisioner.SignSSHOptions, signOpts ...provisioner.SignOption (*ssg.Certificate, err error) {
   certs, prov, err = a.signSSH(...)
   a.incrProvisionerCounter(prov, err, Meter.SSHSigned) 
   return certs, err
}
```

We can keep the current interface, although, as I said, I'm not opposed to changing it. But especially, it will be harder to cause bugs in the future because a PR adds something like `if err := foo(); err != nil { ... }`, well and we won't need the pointer to an interface.

maraino (changes requested):
I've added a few comments.

We should replace the current signing code with internal methods and do a minimal wrapper on those with the metrics. See an example in one of my comments.

For Sign and SignSSH we do not consider errors in `authority.Authorize` as errors.

internal/metrix/meter.go
azazeal: It was discussed and decided against.

internal/metrix/meter.go
azazeal: They've been renamed.

authority/authority.go
```diff
@@ -958,3 +972,50 @@ func (a *Authority) startCRLGenerator() error {
...
+	var name string
+	if p := *prov; p != nil {
+		name = p.GetName()
+	}
+
+	count(a.meter, name, *err == nil)
+}
```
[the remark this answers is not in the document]
azazeal: These would be authorization errors and not errors that happen while signing the certificate. Happy to consider adding more instruments in the future or reviewing a PR that adds them.

authority/ssh.go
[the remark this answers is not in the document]
azazeal: We're already setting the error to `nil` when `storeCertificate` fails and the `errors.Is(err,  db.ErrNotImplemented)`.

If we base our decision to instrument based on whether the returned chain is nil, then we lose the ability (in other places, outside the scope of open source) to instrument in more detail (perhaps according to the underlying type of error).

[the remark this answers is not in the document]
azazeal: Whether `(*Authority).storeCertificate` should return a `nil` error, where it now returns `db.ErrNotImplemented` (so that we can avoid checking for the particular exception), is something to consider, for the future.

authority/ssh.go
azazeal: Cleaning up sounds nice and I hope that we will. There's quite a few things that could benefit from being worked on in many places.

Having said that, named returns and 'defer' statements aren't uncommon when dealing with instrumented code and I don't see how adding yet another level of indirection is helpful.

If we instrument `SignSSH` and not `signSSH` then a caller of `signSSH` would need to remember to instrument the call.

The pointer to a pointer (interface), could be solved by refactoring the defer statement to be:

```go
defer func() { incrProvisionerCounter(prov, err, a.Meter.SSHSigned) }()
```

instead of 

```go
defer a.incrProvisionerCounter(prov, err, Meter.SSHSigned)
```

Other than verbosity, there's not a lot of difference to these signatures.

authority/ssh.go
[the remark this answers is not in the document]
azazeal: In fact, I've changed the signature of meter functions to directly accept the provisioner & the error.

authority/ssh.go
hslatman: @maraino Adding a `SignContext` sounds good. In fact, I have that in https://github.com/smallstep/certificates/pull/1542. I should have a look at that again, sometime soon 😅

dopey:
Want to chime in and say that, while I agree with the sentiment of doing a deeper rewrite of certificates to enable easier instrumentation, I think it's outside the scope of this feature. Let's set aside some time in the future to plan and execute some deeper changes that will make the APIs easier and more intuitive.

dopey (dismissed):

PR #1670, opened 2024-01-10
Remove `rusty-jwt-cli` 

hslatman:


acme/challenge.go
```diff
@@ -446,158 +441,178 @@ func wireOIDC01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
 	return nil
 }
 
-func wireDPOP01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSONWebKey, payload []byte) error {
+type wireDpopPayload struct {
+	// AccessToken is the token generated by wire-server
+	AccessToken string `json:"access_token,omitempty"`
```
maraino, line 446: Ok, but we can skip `omitempty` on structs used for unmarshaling only.

acme/challenge.go
```diff
@@ -446,158 +441,178 @@ func wireOIDC01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
-		kid,
-		"--key",
-		file.Name(),
-	)
-
-	stdin, err := cmd.StdinPipe()
+	params := verifyParams{
+		token:      dpopPayload.AccessToken,
+		key:        dpopOptions.GetSigningKey(),
+		accountJWK: accountJWK,
```
maraino, line 480: I commented on the other PR something about this, about the names, and that you can just pass the public key, accountJWK.Key

acme/challenge.go
```diff
@@ -446,158 +441,178 @@ func wireOIDC01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
+	key        string
+	issuer     string
+	accountJWK *jose.JSONWebKey
+	wireID     wire.ID
+	challenge  *Challenge
+	t          time.Time
+}
+
+func parseAndVerifyWireAccessToken(v verifyParams) (*wireAccessToken, *wireDpopToken, error) {
+	k, err := pemutil.Parse([]byte(v.key)) // TODO(hs): move this to earlier in the configuration process? Do it once?
```
maraino, line 543: Ok, I think v.key will a crypto.PublicKey now.

acme/challenge.go
```diff
@@ -446,158 +441,178 @@ func wireOIDC01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
-	ch.Status = StatusValid
-	ch.Error = nil
-	ch.ValidatedAt = clock.Now().Format(time.RFC3339)
-
-	if err = db.UpdateChallenge(ctx, ch); err != nil {
-		return WrapErrorISE(err, "error updating challenge")
+	pk, ok := k.(ed25519.PublicKey) // TODO(hs): allow more key types
+	if !ok {
+		return nil, nil, fmt.Errorf("unexpected type: %T", k)
 	}
```
maraino, line 551: Why don't just fail on jwt.Claims? Wire can change this at any time.

acme/challenge.go
```diff
@@ -446,158 +441,178 @@ func wireOIDC01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
+		return nil, nil, fmt.Errorf("failed validating Wire DPoP token claims: %w", err)
 	}
 
-	rawDpop, ok := access["proof"].(string)
-	if !ok {
-		return WrapErrorISE(err, "invalid dpop proof format in access token")
+	if err := accessToken.ValidateWithLeeway(jose.Expected{
+		Time:   v.t,
+		Issuer: v.issuer,
+	}, 360*time.Second); err != nil {
```
maraino, line 566: 6 minutes? We usually use 1m.

acme/challenge.go
```diff
@@ -446,158 +441,178 @@ func wireOIDC01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
+		return nil, nil, fmt.Errorf("failed validation: %w", err)
 	}
 
-	parsedDpopToken, err := jose.ParseSigned(rawDpop)
+	rawKid, err := v.accountJWK.Thumbprint(crypto.SHA256)
 	if err != nil {
-		return WrapErrorISE(err, "invalid DPoP token")
+		return nil, nil, fmt.Errorf("failed to compute JWK thumbprint")
+	}
```
maraino, line 573: Although it's probably the same, if the account has a kid `accountJWK.KeyID` we should use this one.

acme/challenge.go
```diff
@@ -446,158 +441,178 @@ func wireOIDC01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
+	jose.Claims
+	Challenge  string `json:"chal,omitempty"`
+	Cnf        *cnf   `json:"cnf,omitempty"`
+	Proof      string `json:"proof,omitempty"`
+	ClientID   string `json:"client_id,omitempty"`
+	APIVersion int    `json:"api_version,omitempty"`
+	Scope      string `json:"scope,omitempty"`
+}
+
+type wireDpopToken map[string]any
```
maraino, line 530: Can we use a struct here?

acme/challenge.go
```diff
@@ -446,158 +441,178 @@ func wireOIDC01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
-		return WrapErrorISE(err, "failed parsing dpop token")
+	if accessToken.Cnf.Kid != kid {
+		return nil, nil, fmt.Errorf("expected kid %q; got %q", kid, accessToken.Cnf.Kid)
+	}
+	if accessToken.ClientID != v.wireID.ClientID {
+		return nil, nil, fmt.Errorf("invalid Wire client ID %q", accessToken.ClientID)
+	}
+	if accessToken.Expiry.Time().After(v.t.Add(time.Hour * 24 * 365)) {
+		return nil, nil, fmt.Errorf("'exp' %s is too far into the future", accessToken.Expiry.Time().String())
 	}
```
maraino, line 587: It would be weird to have tokens with a long expiration time, but why do we allow ~ one year?

acme/challenge.go
```diff
@@ -446,158 +441,178 @@ func wireOIDC01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
+	if err = jwt.Claims(pk, &accessToken); err != nil {
+		return nil, nil, fmt.Errorf("failed validating Wire DPoP token claims: %w", err)
 	}
 
-	rawDpop, ok := access["proof"].(string)
-	if !ok {
-		return WrapErrorISE(err, "invalid dpop proof format in access token")
+	if err := accessToken.ValidateWithLeeway(jose.Expected{
+		Time:   v.t,
+		Issuer: v.issuer,
```
maraino, line 565: Do we need to validate the audience?

maraino (changes requested):

acme/challenge.go
hslatman: I took the value from Wire's implementation. The reason it exists is because they want to protect against clients trying to use a token that's valid for _far_ too long.

acme/challenge.go
hslatman: After discussion with Wire I think the `map[string]any` suits their use case best.

acme/challenge.go
hslatman: This value was taken from Wire's CLI invocation

acme/challenge.go
hslatman: Will add this in a follow up after checking with Wire.

acme/challenge.go
```diff
@@ -559,45 +532,103 @@ func wireDPOP01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
 
-	orders, err := db.GetAllOrdersByAccountID(ctx, ch.AccountID)
-	if err != nil {
-		return WrapErrorISE(err, "could not find current order by account id")
+	if err := accessToken.ValidateWithLeeway(jose.Expected{
+		Time:   v.t,
+		Issuer: v.issuer,
+	}, 360*time.Second); err != nil {
+		return nil, nil, fmt.Errorf("failed validation: %w", err)
 	}
```
beltram, line 595: :Likewise an access token has a JWK in its header that must match the provided verifying key (v.tokenKey)

acme/challenge.go
```diff
@@ -559,45 +532,103 @@ func wireDPOP01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
+		return nil, nil, fmt.Errorf("'exp' %s is too far into the future", accessToken.Expiry.Time().String())
 	}
 
-	order := orders[len(orders)-1]
+	dpopJWT, err := jose.ParseSigned(accessToken.Proof)
+	if err != nil {
+		return nil, nil, fmt.Errorf("invalid Wire DPoP token: %w", err)
+	}
+	var dpopToken wireDpopToken
+	if err := dpopJWT.Claims(v.dpopKey, &dpopToken); err != nil {
```
beltram, line 611: The DPoP token must also be validated (ValidateWithLeeway) with max expiry and the same leeway

acme/challenge.go
```diff
@@ -559,45 +532,103 @@ func wireDPOP01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
+	challenge, ok := dpopToken["chal"].(string)
+	if !ok {
+		return nil, nil, fmt.Errorf("invalid challenge in Wire DPoP token")
+	}
+	if challenge != v.chToken {
+		return nil, nil, fmt.Errorf("invalid Wire DPoP challenge %q", challenge)
 	}
 
-	return nil
+	handle, ok := dpopToken["handle"].(string)
```
beltram, line 623: The DPoP token must have the same ClientId (v.wireID.ClientID) as the access token

acme/challenge.go
```diff
@@ -559,45 +532,103 @@ func wireDPOP01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
 
-	if err := db.CreateDpopToken(ctx, order, dpop); err != nil {
-		return WrapErrorISE(err, "failed storing DPoP token")
+	challenge, ok := dpopToken["chal"].(string)
+	if !ok {
+		return nil, nil, fmt.Errorf("invalid challenge in Wire DPoP token")
+	}
+	if challenge != v.chToken {
+		return nil, nil, fmt.Errorf("invalid Wire DPoP challenge %q", challenge)
 	}
```
beltram, line 622: The DPoP token must have the same "nonce" as the access token

beltram, line 622: The DPoP token must have the same issuer as the access token

authority/provisioner/wire/oidc_options.go
```diff
@@ -0,0 +1,155 @@
...
+	AuthURL     string   `json:"authorization_endpoint,omitempty"`
+	TokenURL    string   `json:"token_endpoint,omitempty"`
+	JWKSURL     string   `json:"jwks_uri,omitempty"`
+	UserInfoURL string   `json:"userinfo_endpoint,omitempty"`
+	Algorithms  []string `json:"id_token_signing_alg_values_supported,omitempty"`
+}
+
+type Config struct {
+	ClientID                   string           `json:"clientId,omitempty"`
+	SignatureAlgorithms        []string         `json:"signatureAlgorithms,omitempty"`
```
maraino, line 28: I don't think this is necessary anymore. It's also a property that is usually in the JWKs and on the `id_token_signing_alg_values_supported` property of the openid-configuration. And you have those in the provider.

maraino: That leaves us with only the ClientID. Assuming the rest of fields are for testing, having something like this makes more sense to me:
```go
type OIDCOptions struct {
	ClientID          string    `json:"clientId,omitempty"` // omitempty?
	Provider          *Provider `json:"provider,omitempty"`
	TransformTemplate string    `json:"transform,omitempty"`
}
```

authority/provisioner/wire/wire_options.go
```diff
@@ -0,0 +1,62 @@
...
+// GetDPOPOptions returns the DPoP options.
+func (o *Options) GetDPOPOptions() *DPOPOptions {
+	if o == nil {
+		return nil
+	}
+	return o.DPOP
+}
+
+func (o *Options) Validate() error {
+	o.validateOnce.Do(
```
maraino, line 35: See previous comment. Check if changing the provisioner using the admin API or HUP will reset the full provisioner and, consequently, this once, too.

maraino (changes requested):

acme/challenge.go
hslatman: Fixed in https://github.com/smallstep/certificates/pull/1681

authority/provisioner/wire/oidc_options.go
```diff
@@ -0,0 +1,155 @@
...
+	SkipClientIDCheck          bool             `json:"-"`
+	SkipExpiryCheck            bool             `json:"-"`
+	SkipIssuerCheck            bool             `json:"-"`
```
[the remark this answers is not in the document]
beltram: Agree you can remove them all

authority/provisioner/wire/wire_options.go
hslatman: I've removed this for now. I think it would work as expected with the `ReloadAdminResources` doing a new `Init` for every provisioner and replacing the provisioners, but I didn't want to block on this right now. Also, I think it should actually be part of the `Init` (or just after that), instead of it being triggered inside of the actual handler, so let's do it the right way (later), instead of doing it now

authority/provisioner/options.go
```diff
@@ -53,20 +54,18 @@ func (o *Options) GetSSHOptions() *SSHOptions {
...
-// GetDPOPOptions returns the OIDC options.
-func (o *Options) GetDPOPOptions() *DPOPOptions {
-	if o == nil {
-		return nil
+	if o.Wire == nil {
+		return nil, errors.New("no Wire options available")
+	}
+	if err := o.Wire.Validate(); err != nil {
+		return nil, fmt.Errorf("failed validating Wire options: %w", err)
 	}
```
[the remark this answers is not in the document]
hslatman: See comment below. Decided to remove it for now.

acme/challenge.go
hslatman: This is done now.

beltram (approved):

authority/provisioner/wire/oidc_options.go
```diff
@@ -0,0 +1,157 @@
...
+	AuthURL     string   `json:"authorization_endpoint,omitempty"`
+	TokenURL    string   `json:"token_endpoint,omitempty"`
+	JWKSURL     string   `json:"jwks_uri,omitempty"`
+	UserInfoURL string   `json:"userinfo_endpoint,omitempty"`
+	Algorithms  []string `json:"id_token_signing_alg_values_supported,omitempty"`
+}
+
+type Config struct {
+	ClientID            string   `json:"clientId,omitempty"`
+	SignatureAlgorithms []string `json:"signatureAlgorithms,omitempty"`
```
maraino, line 28: Some notes about the `go-oidc` implementation:

* If signature algorithms are not set, then the algorithms used in the provider are used. The use case for it is to either restrict or set the algorithms if the provider doesn't have them.
* The client id is matched with a `string.Contains` to the token audience.
* As it is, the `go-oidc` provider will download the JWKs on each request. The provider has a mechanism to cache and re-download the keys if the keyId does not match, they don't use any cache header, this will be ok in most cases, the exception would be if the implementation uses Key IDs that are not fingerprints. It would make more sense to cache the `*oidc.Provider` rather than the provider config.

maraino (approved):
lgtm, but I still don't like how the OIDCOptions work. See my notes.

acme/challenge.go
```diff
@@ -353,63 +352,78 @@ func dns01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSONWebK
...
+	}
+
+	var claims struct {
+		Name         string `json:"preferred_username,omitempty"`
+		Handle       string `json:"name"`
+		Issuer       string `json:"iss,omitempty"`
+		GivenName    string `json:"given_name,omitempty"`
+		KeyAuth      string `json:"keyauth"`
+		ACMEAudience string `json:"acme_aud,omitempty"`
+	}
```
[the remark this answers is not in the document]
hslatman: Will have a look at those. There's some more places I can remove them; tests primarily

PR #1672, opened 2024-01-12
Use a struct for the Wire DPoP token

hslatman:
@maraino this works, and I guess it _can_ be used for DPoPs in practice, because their format should be relatively stable. For OIDC ID tokens it may not be as nice, because those would depend on the IdP, and could contain custom claims too. The thing is that Wire would like to use data from the OIDC ID token (and possibly DPoP token) to put in the template, so you would end up with something like this:

```golang
{{
    "subject": {{
        "organization": "Org",
        "commonName": {{ toJson .Oidc.preferred_username }}
    }},
    "uris": [{{ toJson .Oidc.name }}, {{ toJson .Dpop.Subject }}],
    "keyUsage": ["digitalSignature"],
    "extKeyUsage": ["clientAuth"]
}}
```

Note the difference in capitalization: because for the DPoP we have a struct, it works with `.Dpop.Subject`; for the OIDC claims they would start with a lowercase character. But I think it's better to be consistent, given that they're both tokens, and to always use the lowercase names. Should make things slightly less confusing, I think.

Before this PR, the DPoP URI would be retrieved as `.Dpop.sub`, from the JWT `"sub"` claim. That's actually not that different from what we support with the `.Token`, so maybe it's OK to keep it as a `map[string]any`?

We could also choose to operate on structs for both OIDC and DPoP tokens internally, but to then set them in the template as `map[string]any`?

### DIFF
--- a/acme/api/wire_integration_test.go
+++ b/acme/api/wire_integration_test.go
@@ -370,7 +370,7 @@ MCowBQYDK2VwAyEA5c+4NKZSNQcR1T8qN6SjwgdPZQ0Ge12Ylx/YeGAJ35k=
 		require.NoError(t, err)
 
 		// TODO(hs): move these to a more appropriate place and/or provide more realistic value
-		err = db.CreateDpopToken(ctx, order.ID, map[string]any{"fake-dpop": "dpop-value"})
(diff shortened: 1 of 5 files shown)


hslatman:
Closing this after discussion with Wire. They preferred a more flexible object, so that more/other data _could_ be added if needed.

PR #1673, opened 2024-01-12
Add OIDC token template transformation

hslatman:
This PR adds a new `transform` configuration property that allows for transforming OIDC tokens, so that claims can be mapped to names in a template. The original claims will be added to the new map if they don't exist yet. These values can then be validated.


acme/challenge.go
```diff
@@ -421,31 +425,51 @@ func wireOIDC01Validate(ctx context.Context, ch *Challenge, db DB, jwk *jose.JSO
...
+		return nil, fmt.Errorf("invalid 'name' %q after transformation", name)
+	}
+
+	handle, ok := transformed["handle"]
+	if !ok {
+		return nil, fmt.Errorf("transformed OIDC ID token does not contain 'handle'")
+	}
+	if wireID.Handle != handle {
+		return nil, fmt.Errorf("invalid 'handle' %q after transformation", handle)
+	}
```
maraino, line 468: Check with wire the case sensitivity of these conditions.

authority/provisioner/wire/oidc_options.go
```diff
@@ -62,6 +66,8 @@ func (o *OIDCOptions) GetConfig() *oidc.Config {
 	}
 }
 
+const defaultTemplate = `{"name": "{{ .name }}", "handle": "{{ .preferred_username }}"}`
```
maraino, line 69: Should we care about quotes in those variables?

Something like this will take care of it:
```
const defaultTemplate =  `{{ printf "{\"name\": %q, \"handle\": %q}"  .name .preferred_username }}`
```

maraino (approved):
looks good, I added some comments to consider.

acme/challenge.go
```diff
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/coreos/go-oidc/v3/oidc"
```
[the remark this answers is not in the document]
hslatman: Mostly convenience at this this time, as it was already implemented by Wire, and didn't want to rip it out before having sufficient test coverage. It's a layer on top of `go-jose` itself, so replacing that with our own thing shouldn't be hard, if we want to. If you think the dynamic OIDC configuration loading can be done better/smaller and make this unused, feel free to rip it out. I might get to it before you, if I decide to refactor some of the OIDC validation.

acme/challenge.go
```diff
@@ -360,25 +362,40 @@ type wireOidcPayload struct {
...
 	}
 
+	// TODO(hs): move this into validation below?
+	expectedKeyAuth, err := KeyAuthorization(ch.Token, jwk)
+	if err != nil {
+		return WrapErrorISE(err, "error determining key authorization")
+	}
+	if expectedKeyAuth != oidcPayload.KeyAuth {
+		return storeError(ctx, db, ch, true, NewError(ErrorRejectedIdentifierType,
+			"keyAuthorization does not match; expected %q, but got %q", expectedKeyAuth, oidcPayload.KeyAuth))
```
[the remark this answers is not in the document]
hslatman: In this case it's stored with the challenge in the DB, and will be returned to the client in case it's not nil. The behavior is the same as for HTTP validation. I don't think just the key authorization will be sufficient to do any sensitive without having access to the account private key, but I can change the error message slightly to not return the right value?

acme/challenge.go
hslatman: @beltram? Now, the transformations _can_ make them use the right case, but then you would also have these values available in the template. Do we need to support case-sensitivity here?

beltram: Nah let's not bother with case-sensitivity. Those claim names will already be canonicalized by the template. And since we tend to adhere to the standards we will have them snake_case (lower case by consequence). So what you have here is fine.

PR #1688, opened 2024-01-22
KMS instrumentation

azazeal:
This PR adds KMS-related instrumentation.


maraino (approved):
This lgtm, assuming that the zero or nil Meter is ok.

PR #1708, opened 2024-02-08
Add `Expires` header to CRL endpoint

hslatman:
This implements the change from https://github.com/wireapp/smallstep-certificates/commit/659117c329da240d6799461917a33b2bd1290c82.

Besides that, it improves the test (coverage) for the CRL HTTP handler.

### DIFF
--- a/api/crl.go
+++ b/api/crl.go
@@ -18,7 +18,7 @@ func CRL(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if crlInfo == nil {
-		render.Error(w, errs.New(http.StatusInternalServerError, "no CRL available"))
+		render.Error(w, errs.New(http.StatusNotFound, "no CRL available"))
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


api/crl.go
```diff
@@ -3,30 +3,44 @@ package api
...
+	crlInfo, err := mustAuthority(r.Context()).GetCertificateRevocationList()
 	if err != nil {
 		render.Error(w, err)
 		return
 	}
 
+	if crlInfo == nil {
+		render.Error(w, errs.New(http.StatusInternalServerError, "no CRL available"))
+		return
+	}
```
maraino, line 23: Perhaps a 404 is a better response if a CA doesn't have the CRL configured.

maraino (changes requested):
I think 404 might be a better response.

api/crl.go
hslatman: I've changed this to a 404 in 3dbb4aad3dc338f3d838f2c153019a84064c7eab.

The 404 case was already returned one layer down, and in general `crlInfo` won't ever be `nil` in the handler, but better safe than sorry 🙂

maraino (approved):

PR #1718, opened 2024-02-13
Support executable provisioner password files in RA install script

yarcod:
#### Name of feature:
Fixing install source of step-ca in setup script

Following the guide on: https://smallstep.com/docs/registration-authorities/acme-for-certificate-manager/index.html?fingerprint=41363fbbc3b4c23a38b1ad6120da34a7876e0480bebd8aa72c12d7a7dbe75d29&caUrl=https://home.edholm.ca.smallstep.com&#1-create-your-ra leads me to download a setup script. Most of the script works as intended, but after the extraction of `step-ca`, the install looks for the `step-ca` in the wrong directory. This PR adjusts the source dir to where `step-ca` actually ends up.

### DIFF
--- a/scripts/install-step-ra.sh
+++ b/scripts/install-step-ra.sh
@@ -188,7 +188,7 @@ CA_VERSION=$(curl -s https://api.github.com/repos/smallstep/certificates/release
 
 curl -sLO https://github.com/smallstep/certificates/releases/download/$CA_VERSION/step-ca_linux_${CA_VERSION:1}_$arch.tar.gz
 tar -xf step-ca_linux_${CA_VERSION:1}_$arch.tar.gz
-install -m 0755 -t /usr/bin step-ca_${CA_VERSION:1}/step-ca
+install -m 0755 -t /usr/bin ./step-ca
 setcap CAP_NET_BIND_SERVICE=+eip $(which step-ca)
 rm step-ca_linux_${CA_VERSION:1}_$arch.tar.gz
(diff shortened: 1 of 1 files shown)


yarcod:
Apologies, I intended to push this next part as a separate PR. 

The idea behind the last part is that if a runnable executable is provided as the password file, it will be run in order to get the passphrase from some other tool. My use case for it is to retrieve an item from 1password (`op`), but it could likewise be used to generate and store a new password in a safe place. 

If this last feature is acceptable, the documentation on the previously mentioned website would need updating. Both with the fact that there *is* such an option, and that it can be provided as an executable.

tashian:
Hi @yarcod,
Thanks for pointing out the issue with the RA installer, and for your contribution. I created #1719 to address that with a newer URL.
Could you please update this PR only cover the password executable? Thanks.

yarcod:
@tashian I reverted the initial commit and kept the second one, as well as edited the PR title to reflect the actual changes.

I realised that the `--provisioner-password-file` is used in other places as well. For consistency, should this flag be updated for, e.g., `step` as well?

tashian:
Hi @yarcod,

We took a look at this PR during our open source review. Unfortunately, having an executable for fetching the password is not a feature that we're able to support at the moment. It's a bit confusing to have a "password file" that is actually an executable, maybe if it were a separate flag we could make it work. But, my suggestion would be to make a copy of this script for yourself and add what you need for supporting the password manager you use. It's not a script that changes very often. Hope this helps.

PR #1743, opened 2024-02-28
Improve end-to-end request ID propagation

hslatman:
The CA will now propagate a request ID from the `X-Request-Id` header from its entrypoint all the way back to the client. The CA clients have been adjusted to generate a request ID (if not yet set) and set this on each request. When an error occurs, the request ID is read from the response header.


internal/userid/userid.go
```diff
@@ -0,0 +1,20 @@
+package userid
+
+import "context"
+
+type userIDKey struct{}
```
azazeal, line 5: You're already in a `userid` package, so `contextKey` or `key` would be great!

logging/handler.go
```diff
@@ -9,6 +9,8 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"github.com/smallstep/certificates/internal/requestid"
```
azazeal, line 12: Maybe separate the `certificates` imports from 3rd party ones?

monitoring/monitoring.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/newrelic/go-agent/v3/newrelic"
 	"github.com/pkg/errors"
+	"github.com/smallstep/certificates/internal/requestid"
```
azazeal, line 12: I'd separate local from 3rd party imports

test/e2e/requestid_test.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"path/filepath"
+	"sync"
+	"testing"
+
+	"github.com/smallstep/certificates/authority/config"
```
azazeal, line 12: Would also separate local from 3rd party imports, if at all possible.

test/e2e/requestid_test.go
```diff
@@ -0,0 +1,132 @@
...
+	)
+	require.NoError(t, err)
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
+		defer wg.Done()
+		err = c.Run()
+		require.Error(t, err) // expect error when server is stopped
```
azazeal, line 89: Which specific error are you expecting here? If there's a sentinel error you're expecting, you could opt for `require.ErrorIs`.

test/e2e/requestid_test.go
```diff
@@ -0,0 +1,132 @@
...
+	healthResponse, err := caClient.HealthWithContext(ctx)
+	require.NoError(t, err)
+	if assert.NotNil(t, healthResponse) {
+		require.Equal(t, "ok", healthResponse.Status)
+	}
+
+	// expect an error when retrieving an invalid root
+	rootResponse, err := caClient.RootWithContext(ctx, "invalid")
+	if assert.Error(t, err) {
+		apiErr := &errs.Error{}
```
azazeal, line 103: `var apiErr *errs.Error` is just enough here. There's no need to allocate an error.

test/e2e/requestid_test.go
```diff
@@ -0,0 +1,132 @@
...
+			// TODO: include the below error in the JSON? It's currently only output to the CA logs. Also see https://github.com/smallstep/certificates/pull/759
+			//assert.Equal(t, "/root/invalid was not found: certificate with fingerprint invalid was not found", apiErr.Msg)
+		}
+	}
+	assert.Nil(t, rootResponse)
+
+	// expect an error when retrieving an invalid root and provided request ID
+	rootResponse, err = caClient.RootWithContext(client.NewRequestIDContext(ctx, "reqID"), "invalid")
+	if assert.Error(t, err) {
+		apiErr := &errs.Error{}
```
azazeal, line 118: Similar allocation suggestion as above.

internal/requestid/requestid_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func newRequest(t *testing.T) *http.Request {
+	t.Helper()
+	r, err := http.NewRequest(http.MethodGet, "https://example.com", http.NoBody)
+	require.NoError(t, err)
+	return r
+}
+
+func Test_Middleware(t *testing.T) {
+	requestWithID := newRequest(t)
```
azazeal, line 20: It'd be great if you'd add the occasional blank line here and there in this block. It'd make it easier to read. As an example, you can group `requestWithID` statements from `requestWithoutID` statements, etc. Whitespace is really strong!

ca/client.go
```diff
@@ -97,12 +98,43 @@ func (c *uaClient) PostWithContext(ctx context.Context, u, contentType string, b
...
+// requestIDHeader is the header name used for propagating request IDs from
+// the CA client to the CA and back again.
+const requestIDHeader = "X-Request-Id"
+
+// newRequestID generates a new random UUIDv4 request ID. If it fails,
+// the request ID will be the empty string.
+func newRequestID() string {
+	requestID, err := randutil.UUIDv4()
+	if err != nil {
+		return ""
```
azazeal, line 113: Is it possible for `randutil.UUIDv4` to fail? If yes, then there's something wrong with the entropy of the underlying system and that's something we should take note of.

authority/provisioner/webhook_test.go
```diff
@@ -101,10 +101,10 @@ func TestWebhookController_isCertTypeOK(t *testing.T) {
 	}
 }
 
-// withRequestID is a helper that calls into [logging.WithRequestID] and returns
-// a new context with the requestID added to the provided context.
+// withRequestID is a helper that calls into [requestid.NewContext] and returns
+// a new context with the requestID added.
 func withRequestID(ctx context.Context, requestID string) context.Context {
```
azazeal, line 106: I think we should be injecting an ID in the context if none is present and, from what I understand, `withRequestID` is only called in tests atm.

The reason I'm mentioning this is that it'd be great if an outgoing call (to a webhook) contained a request ID in its logs, even though there's none attached to it from the caller.

azazeal (changes requested):
Nitpicking here and there

ca/client.go
hslatman: It is possible to fail, which is why an error is returned. It can fail on a client system, but there's not much we can do, unless we want an alternative random ID to be generated that doesn't rely on the failing code path. 

Given that the server side will generate a request ID if the client doesn't provide one, IMO we don't need that additional method for generating a unique ID.

authority/provisioner/webhook_test.go
hslatman: `withRequestID` is intended to be used in testing only. Have made that stronger by making it an actual helper. 

The request ID was already forwarded to all webhooks. It is guaranteed to exist and to be set, due to ordering of the CA middleware. I've added an additional integration test covering this case.

Issue #1094, opened 2022-10-06
configurable Certificate duration for the ACME Provisioner


marcohald:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

It would be useful to have a config option for the Certificate duration for the ACME Provisioner as discussed here https://github.com/smallstep/certificates/discussions/1049
## Why is this needed?
We monitor the validity period of our SSL Certificates and like to have a duration of 30 Days and renew them after 15 Days.  
So that we have enough time to troubleshoot any issues with renewing the certificates.

marcohald:
@hslatman yes you are correct.
I think it would be better if the lifetime and the renew time could be specified in hours, for the most users that need it customized

maraino:
I don't understand the need for this, the CA will automatically renew its own certificate, and if it's unable to do it, it won't be able to issue new certificates to anyone. And it's always possible to monitor that the CA always has a valid certificate just doing a request to it.

sshipway:
We have a default 90d lifetime on the Provisioner, but the RA is still generating a 1d certificate.  Autorenewals are working; however, we have also configured Slack integration to notify us of expired certificates, and -- apparently because of the short lifetime -- we're getting daily slack notifications from Smallstep that the RA certificate has expired, even though it autorenews.
Being able to configure the RA to use a 90d certificate would allow renewal to happen comfortably outside the notification window so that we would not be receiving these daily spurious notifications.

maraino:
@sshipway with this PR https://github.com/smallstep/certificates/pull/1764, I'm letting the CA decide which lifetime the RA should have.

PR #1764, opened 2024-03-12
Let the CA determine the RA lifetime

maraino:
When the RA mode with StepCAS is used, let the CA decide which lifetime the RA should get instead of always requiring 24h.

This commit also fixes linter warnings.

Related to #1094

### DIFF
--- a/authority/tls.go
+++ b/authority/tls.go
@@ -913,7 +913,10 @@ func (a *Authority) GetTLSCertificate() (*tls.Certificate, error) {
 		return fatal(err)
 	}
 
-	// For StepCAS RA let the lifetime to the provisioner used by the CA.
+	// Set the cert lifetime as follows:
+	//   i) If the CA is not a StepCAS RA use 24h, else
+	//  ii) if the CA is a StepCAS RA, leave the lifetime empty and
+	//      let the provisioner of the CA decide the lifetime of the RA cert.
 	var lifetime time.Duration
 	if casapi.TypeOf(a.x509CAService) != casapi.StepCAS {
 		lifetime = 24 * time.Hour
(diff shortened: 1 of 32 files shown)


cas/stepcas/stepcas.go
```diff
@@ -73,8 +78,8 @@ func (s *StepCAS) CreateCertificate(req *apiv1.CreateCertificateRequest) (*apiv1
 		return nil, errors.New("createCertificateRequest `csr` cannot be nil")
 	case req.Template == nil:
 		return nil, errors.New("createCertificateRequest `template` cannot be nil")
-	case req.Lifetime == 0:
-		return nil, errors.New("createCertificateRequest `lifetime` cannot be 0")
+	case req.Lifetime < 0:
+		return nil, errors.New("createCertificateRequest `lifetime` cannot less than 0")
```
dopey, line 82:
```suggestion
		return nil, errors.New("createCertificateRequest `lifetime` cannot be less than 0")
```

authority/tls.go
```diff
@@ -913,10 +913,16 @@ func (a *Authority) GetTLSCertificate() (*tls.Certificate, error) {
 		return fatal(err)
 	}
 
+	// For StepCAS RA let the lifetime to the provisioner used by the CA.
```
dopey, line 916:
```suggestion
	// For StepCAS RA, set the lifetime to the provisioner used by the CA.
```

maraino: Fixed with 0ac90235

cas/stepcas/stepcas.go
maraino: Fixed with 0ac90235

dopey (approved):

Issue #1757, opened 2024-03-11
[Bug]: wrong oidc provisioner config prevents startup, can't remove provisioner


PeterGrace:
### Steps to Reproduce

- setup an oidc provisioner with keycloak using legacy url format (https://keycloak-server/auth/realms/realm/...)
- upgrade keycloak to newer version with updated route url (https://keycloak-server/realms/realm/...)
- step-ca fails to start with `error parsing https://keycloak-server/auth/realms/gfpd/.well-known/openid-configuration: issuer cannot be empty`

### Your Environment

* OS - Docker container
* `step-ca` Version - 0.25.2

### Expected Behavior

I'd like to be able to start step-ca so that I can remove or repair the faulty provisioner config

### Actual Behavior

step-ca errors on start, no way to change the provisioner config

### Additional Context

I attempted to remediate by turning off remoteAdmin which allows step-ca to start, but it also appears to not read the database, so the provisioner list is `[]` when remoteAdmin is false, leaving me no ability to delete or edit the faulty provisioner.

### Contributing

Vote on this issue by adding a 👍 reaction.
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

maraino:
Fixed by https://github.com/smallstep/certificates/pull/1765

PR #1765, opened 2024-03-13
Do not fail if a provisioner cannot be initialized

maraino:
This commit will mark a provisioner as disabled if it fails to initialize. The provisioner will be visible, but authorizing a token with a disabled provisioner will always fail.

Fixes: #589, #1757


maraino:
@dopey @hslatman, I still need to make sure this works as expected in all cases. Should I write the reason in the provisioner list, as I'm doing right now?

PeterGrace:
@maraino thanks for taking up this issue.  Would it be possible for you to push a docker image or link me to a binary for x86_64 that I can use temporarily to fix my CA while this PR is under consideration?  I haven't been able to sign certs for days and its starting to hurt.  Thanks!

hslatman:
@maraino the approach makes sense to me. 

Only thing I can think of now is that it maybe needs to be a more specific state, such as `Uninitialized` instead of `Disabled`, or more generic than it is now, like an actual `State`. That way there could be other reasons than initialization errors for provisioners to not be available/usable in the future, such as when a user decides to disable a provisioner after migrating to a different provisioner with a period of them being both being available, for example. Keeping the old provisioner around but with some specific state indicating unavailability could then result in a different error message when there's an attempt to use it, instead of having to handle the case that the provisioner is not found.

Admittedly, that's a different thing than this issue, but the behavior is related to what's in this PR. It would probably make things slightly more complex because the state of the provisioner would have to be taken into account in more places, but I don't think that's a bad thing per se.

As long as the current changes are mostly internal, and our JSON outputs not being documented in formal API docs, I think we can change/break the property names and/or uses in the future too, so I don't view the above as a blocker for the current state to go in.

dopey (approved):

PR #1795, opened 2024-04-10
Prevent exposing any internal details in SCEP failure message

hslatman:
To be on the safe side, block errors from signing operations from being returned to the client. We should revisit, and make it return a more informative error, but with high assurance that no sensitive information is added to the message.


hslatman:
A quick follow-up, @dopey 😅

dopey (approved):

PR #1834, opened 2024-05-08
Enable cookies on acme client

### LINKED ISSUE #1833: Add an in-memory cookie jar to http.Client returned by NewClient
## Hello!

## Issue details

I am trying to use an ACME server that is behind an AWS ELB. The ELB uses cookies to ensure that an http session is routed back to the same ACME server using cookies.

Without cookies, caddy will just be connected by the ELB to random servers.

This results in the ACME server rejecting nonces that were generated by another ACME server.

## Why is this needed?

Interfacing with ACME servers behind load balancers correct.

arontsang:
Fixes #1833

### DIFF
--- a/acme/client.go
+++ b/acme/client.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"net"
 	"net/http"
+	"net/http/cookiejar"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


hslatman:
As discussed in #1833, this is not the right fix.

PR #1844, opened 2024-05-14
Verify provisioner with id if available

maraino:
This commit allows verifying the account provisioner with the request provisioner using the ID instead of the name. When the provisioner are stored in the DB, the provisioner ID is not mutable but the name is.


acme/api/middleware.go
```diff
@@ -334,15 +334,22 @@ func lookupJWK(next nextHTTP) nextHTTP {
...
-					// Provisioner in the URL must match the provisioner with
-					// which the account was created.
-					render.Error(w, acme.NewError(acme.ErrorUnauthorizedType,
-						"account provisioner does not match requested provisioner; account provisioner = %s, requested provisioner = %s",
-						accProvName, reqProvName))
-					return
+				if acc.ProvisionerID == "" || reqProv.GetID() != acc.ProvisionerID {
+					reqProvisioner := reqProv.GetName()
+					accProvisioner := acc.ProvisionerName
+					if reqProvisioner != accProvisioner {
```
hslatman, line 340: There's an unlikely (I think) case in which the IDs don't match, but the account has the old provisioner name, and another provisioner was renamed to that old name, which would now be considered OK, and thus not return an error.

Maybe it's easier to follow when the ID check happens in a different case from the one that checks the names.

maraino: Fixed with 8673818

hslatman (approved):

dopey (approved):

hslatman (commented):
lgtm 😄

PR #1849, opened 2024-05-15
Log errors using slog.Logger

maraino:
This commit allows logging errors in a `slog.Logger` injected in the context. This logger type is not currently used directly in step-ca, but this will change in the future.

The main change is in `API/log/log.go`; the rest is adapting the code to allow this. 

I've considered using `func Error(ctx context.Context, w http.ResponseWriter, err error)`, but this won't allow us to add extra things from the request to the log if we need to. 

I've also had to change the render methods because they can call `Error`.

cc: @hslatman

### DIFF
--- a/api/log/log.go
+++ b/api/log/log.go
@@ -2,12 +2,30 @@
 package log
 
 import (
+	"context"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"os"
 
 	"github.com/pkg/errors"
 )
+
+// ErrorKey is the logging attribute key for error values.
+const ErrorKey = "error"
+
+type loggerKey struct{}
+
+// NewContext creates a new context with the given slog.Logger.
+func NewContext(ctx context.Context, logger *slog.Logger) context.Context {
+	return context.WithValue(ctx, loggerKey{}, logger)
+}
+
+// FromContext returns the logger from the given context.
+func FromContext(ctx context.Context) (l *slog.Logger, ok bool) {
+	l, ok = ctx.Value(loggerKey{}).(*slog.Logger)
+	return
+}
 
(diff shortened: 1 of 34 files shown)


api/log/log.go
```diff
@@ -2,13 +2,31 @@
...
 	"github.com/pkg/errors"
 )
 
+// ErrorKey is the logging attribute key for error values.
+const ErrorKey = "error"
+
+type loggerKey struct{}
+
+// NewContext creates a new context with the given slog.Logger.
+func NewContext(ctx context.Context, logger *slog.Logger) context.Context {
```
azazeal, line 20: I think that instead of attaching a `slog.Logger` reference to the `context.Context`, we may benefit more by attaching a `func`. This way we can inject this concern in more elastic ways in the future.

Thoughts?

```go
type HandlerErrorLogger func(http.ResponseWriter, *http.Request, error)

func (fn HandlerErrorLogger) call(w http.ResponseWriter, r *http.Request, err error) {
	if fn == nil {
		return // noop or use a sensible default (print to stderr via slog, etc)
	}

	fn(w, r, err)
}

func WithHandlerErrorLogger(ctx context.Context, fn HandlerErrorLogger) context.Context {
	// ...
}

func HandlerErrorLoggerFromContext(ctx context.Context) HandlerErrorLogger {
	// ...
}

func Error(rw http.ResponseWriter, r *http.Request, err error) {
	HandlerErrorLoggerFromContext(r.Context()).call(rw, r, err)
	// ...
}
```

Thoughts?

maraino: Fixed with c0b7c33a

azazeal (approved):

PR #1850, opened 2024-05-16
Add GetX509Signer method

maraino:
This commit adds a method to the Authority type that returns the signer used to sign X509 certificates.

### DIFF
--- a/authority/tls_test.go
+++ b/authority/tls_test.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


authority/tls_test.go
```diff
@@ -418,7 +428,7 @@ ZYtQ9Ot36qc=
 			require.NoError(t, err)
 			testAuthority.db = &db.MockAuthDB{
 				MStoreCertificate: func(crt *x509.Certificate) error {
-					sassert.Equals(t, crt.Subject.CommonName, "smallstep test")
+					assert.Equal(t, crt.Subject.CommonName, "smallstep test")
```
hslatman, line 431: It's better to reverse the order of arguments for expected and actual, but not a blocker. I've been changing those whenever I saw them happening and felt like they didn't muddy the PRs I was working at the time.

authority/tls_test.go
```diff
@@ -223,6 +224,15 @@ func generateSubjectKeyID(pub crypto.PublicKey) ([]byte, error) {
 	return hash[:], nil
 }
 
+func assertHasPrefix(t *testing.T, s, p string) bool {
+	if strings.HasPrefix(s, p) {
+		return true
+	}
+	t.Helper()
+	t.Errorf("'%s' is not a prefix of '%s'", p, s)
```
hslatman, line 232:
```suggestion
	t.Errorf("%q is not a prefix of %q", p, s)
```

authority/tls_test.go
maraino: I know, but I've just find and replace everything with sassert 🙂

Fixed with 812ffd3

hslatman (approved):

PR #1894, opened 2024-06-20
allow relative paths for root, crt and key in ca.conf

### LINKED ISSUE #372: Use `StepAbs` for all files in the ca.json
### Description

Currently, some files in the `ca.json`, like the certificate templates, can be configured relative to the `STEPPATH`, for example, if a template file is set in the ca.json to `"templates/x509/leaf.tpl"` the location of the file would be relative to the STEPPATH, it will be `$(step path)/templates/x509/leaf.tpl`.

We should make this functionality available across all the files in the ca.json. That way, we can have things like:

```json
{
   "root": "certs/root_ca.crt",
   "crt": "certs/intermediate_ca.crt",
   "key": "secrets/intermeidate_ca_key",
   "...": {}
}
```

marten-seemann:
Fixes #372.

Do we have any end-to-end tests for this?

### DIFF
--- a/commands/app.go
+++ b/commands/app.go
@@ -183,6 +183,13 @@ func appAction(ctx *cli.Context) error {
 		}
 	}
 
+	// allow paths relative to STEPPATH for root, crt and key
+	for i, root := range cfg.Root {
+		cfg.Root[i] = step.Abs(root)
+	}
+	cfg.IntermediateCert = step.Abs(cfg.IntermediateCert)
+	cfg.IntermediateKey = step.Abs(cfg.IntermediateKey)
+
 	if cfg.AuthorityConfig != nil {
 		if token == "" && strings.EqualFold(cfg.AuthorityConfig.DeploymentType, pki.LinkedDeployment.String()) {
 			return errors.New(`'step-ca' requires the '--token' flag for linked deploy type.


commands/app.go
```diff
@@ -183,6 +183,13 @@ func appAction(ctx *cli.Context) error {
 		}
 	}
 
+	// allow paths relative to STEPPATH for root, crt and key
+	for i, root := range cfg.Root {
+		cfg.Root[i] = step.Abs(root)
+	}
+	cfg.IntermediateCert = step.Abs(cfg.IntermediateCert)
+	cfg.IntermediateKey = step.Abs(cfg.IntermediateKey)
```
hslatman, line 191: Maybe this logic can/should be part of `config.LoadConfiguration`. That should also allow for testing just this logic as part of the `config` [package tests](https://github.com/smallstep/certificates/blob/master/authority/config/config_test.go). 

Changing the path in this way most likely results in a different value being marshaled and persisted at a different time. I think we should prevent that from happening. I think with how the `config.Config` struct is currently used, it'll likely require additional non-exported properties for the updated fields and a getter. 

There are some "integration-like" tests that touch the config in https://github.com/smallstep/certificates/blob/master/authority/authority_test.go. They don't test the new behavior, but they can help in adding new tests for this. If you add new tests that operate on the `appCommand`, you can put them in a new sub dir of `test/integration`.

Issue #1875, opened 2024-06-10
Feature request: Environment variables for database setup in container


mikaelparkefelt:
### Discussed in https://github.com/smallstep/certificates/discussions/1428

<div type='discussions-op-text'>

<sup>Originally posted by **btrepp** June 11, 2023</sup>
Hi there,

First of, this tool is amazing, an ACME provider that I can easily manage private certificates, its a great missing piece for hobby labs!.

I've been integrating it with my k8s setup, and made some of the manifests/kustomizations https://git.sr.ht/~btrepp/environment/tree/main/item/dist/step-ca/latest/base

One of the issues I've encountered, is there doesn't seem to be an easy way to container the docker container to use a database instead of the badger DB. My motivation to do this is that I already have some postgres bases that use rsnapshot to ensure I have backups, at the moment with a StatefulSet/PVC provisioner it works okay, but its fairly easy to 'remove' your pvc and potentially lose your data :).

It would be great if the docker container could be pointed to databases as env vars (I've tried to look at the entry point, and it doesn't look like this is supported yet). That way it's easy to keep the data backed up if needed, plus also easy to run the step-ca as a deployment that could scale, and leave postgres as a stateful set for persistence.</div>

mikaelparkefelt:
Last comment in the discussion:
`
And, you're welcome to open a feature request issue on this repo for this, if it's something you'd like us to consider adding.
`

jdoss:
Hey @mikaelparkefelt thanks for opening this feature request. We can look into adding features to the step-ca container entrypoint to configure the CA to use PostgreSQL but in the meantime, I actually configure my step-ca container after the initial setup with this bash snipit from my automation scripts. 

It should be a stopgap for users that want to automate launching step-ca with PostgreSQL support. You will need `jq` and `sponge` for it to work. I bootstrap my step-ca container and let it configure itself with BadgerDB and then I set it up to use PostgreSQL after the fact. You will want to adjust the if statements below to fit your needs and change `${SELFHOST_APP_DATA_DIR}/stepca/config/ca.json` to the path to the `ca.json` file after it has been created.

You need to restart the step-ca container once the `ca.json` is adjusted to use PostgreSQL. It should be safe to remove the BadgerDB directory after everything has been setup to use PostgreSQL. I have those lines commented out in my example below.

Please note that this creates a fresh database in PostgreSQL and there is no path to import your existing BadgerDB currently.

```bash
[...]

Issue #807, opened 2022-01-24
Programmatically set ca.json configuration for containers


smarsh-tim:
### What would you like to be added
I would like to be able to programmatically set ca.json configuration settings, such as `maxTLSCertDuration` via the `step ca init` script.  Support for a DOCKER_ environment variable would be ideal, much like how values like `DOCKER_STEPCA_INIT_NAME` are being passed.

### Why this is needed
This will allow me to configure additional fields that aren't currently supported during `step ca init`. A specific example is `maxTLSCertDuration`, which defaults to 24 hours. I would like to increase this settings to a greater value. 

Editing the ca.json manually and restarting is not preferred in a container orchestrated setup.

So we're left to doing something a little hacky like this in the [entrypoint.sh](https://github.com/smallstep/certificates/blob/master/docker/entrypoint.sh) script for the [Dockerfile](https://github.com/smallstep/certificates/blob/master/docker/Dockerfile.step-ca):

```sh
    function set_custom_vars () {
            if grep -Fq "maxTLSCertDuration" ${STEPPATH}/config/ca.json
            then
                echo "'maxTLSCertDuration' already set."
            else
                echo "Did not find 'maxTLSCertDuration' in ca.json. Adding value before start."
                mkdir -p ${STEPPATH}/config/backups
                cp ${STEPPATH}/config/ca.json ${STEPPATH}/config/backups/ca.json_$(date +%s)
                awk 'NR==1{print; print "        \"maxTLSCertDuration\":\"48\","} NR!=1' ${STEPPATH}/config/ca.json > /tmp/ca.json && mv /tmp/ca.json ${STEPPATH}/config/ca.json
            fi
        }
```

* Note that `jq` isn't available on the Alpine image, and an init container wouldn't do much since the step ca init occurs first, then we make the changes, then start up.

dopey:
Hey @smarsh-tim thanks for opening the issue! 

This is definitely a problem. Unfortunately we're not quite sure what the answer should be, and we don't want to implement something half-baked. We've been discussing for a while having some sort of import / export functionality for CA config - this issue might get rolled into that.

Our short term suggested fix is to generate your ca.json separately and then mount it as a volume. There's code in the entrypoint.sh file to bypass `step ca init` if a config file is already in the right location. Let us know if you're having trouble getting that working.

PR #1895, opened 2024-06-24
Update entrypoint.sh to add /docker-entrypoint-initdb.d

wizhippo:
#### Name of feature:

Add /docker-entrypoint-initdb.d support like most other images.

This will only run these on init and allows adding scripts that modify the setup just after first init, for example you could use it to modify ca.json and add/change datasource.

Instead of having to create/override the default entrypoint this allows to extend functionality like most other docker images

May help with items like https://github.com/smallstep/certificates/issues/1875, https://github.com/smallstep/certificates/issues/807 as these could be added as init scripts.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -76,6 +76,17 @@ function step_ca_init () {
     echo "🤫 This will only be displayed once."
     shred -u $STEPPATH/provisioner_password
     mv $STEPPATH/password $PWDPATH
+
+    # Run scripts in /docker-entrypoint-initdb.d
+    if [ -d /docker-entrypoint-initdb.d ]; then
+        for f in /docker-entrypoint-initdb.d/*; do
+            case "$f" in
+                *.sh)  echo "$0: running $f"; . "$f" ;;
+                *)     echo "$0: ignoring $f" ;;
+            esac
+            echo
+        done
+    fi
 }
 
 if [ -f /usr/sbin/pcscd ]; then


tashian:
Hi @wizhippo thank you for the PR.

In the case of Postgres and some other database images that use it, the `/docker-entrypoint-initdb.d` script directory allows people to run SQL and bash scripts at a point in the execution flow that can't easily be accessed any other way.

With `step-ca`, there are good options for injecting init scripts that don't require any additional logic in the entrypoint.sh:
- You can build your own Dockerfile, adding a layer that runs an init script via `CMD`.
- You can inject an init script on the command line, eg:

   ```
   docker run \
      -p 9000:9000 \
      -v step:/home/step \
      -v ./init.d:/init.d \
      -e "DOCKER_STEPCA_INIT_NAME=Smallstep" \
      -e "DOCKER_STEPCA_INIT_DNS_NAMES=localhost" \
      -e "DOCKER_STEPCA_INIT_REMOTE_MANAGEMENT=true" \
      smallstep/step-ca \
      sh -c '[ ! -f "${STEPPATH}/config/ca.json" ] && sh /init.d/script.sh ; /usr/local/bin/step-ca --password-file /home/step/secrets/password /home/step/config/ca.json'
   ```

- A third option is to init and start up the CA, make whatever changes you need (to `ca.json`, for example) while the CA is running, and then restart the container to pick up the changes.

None of these options require overriding or changing the entrypoint script.
So, given the possibilities that already exist, I'm going to close this issue.

PR #1910, opened 2024-07-02
Do strict DNS lookup on ACME

maraino:
This commit changes the ACME challenges to perform a strict DNS lookup without taking into account the search list in a resolv.conf


dopey (approved):

acme/challenge_test.go
```diff
@@ -457,7 +457,7 @@ func TestChallenge_Validate(t *testing.T) {
 						assert.Equal(t, "zap.internal", updch.Value)
 						assert.Equal(t, StatusPending, updch.Status)
 
-						err := NewError(ErrorConnectionType, "error doing http GET for url http://zap.internal/.well-known/acme-challenge/%s: force", ch.Token)
+						err := NewError(ErrorConnectionType, "error doing http GET for url http://zap.internal./.well-known/acme-challenge/%s: force", ch.Token)
```
hslatman, line 460: I think we shouldn't report the dot in the error messages, as it's different from what the user requested (most likely). Is it possible to keep the rooted name strictly internally?

test/integration/scep/common_test.go
```diff
@@ -39,6 +39,9 @@ func newCAClient(t *testing.T, caURL, rootFilepath string) *ca.Client {
 }
 
 func requireHealthyCA(t *testing.T, caClient *ca.Client) {
+	// Wait for CA
+	time.Sleep(time.Second)
```
hslatman, line 43: I see I forgot this before:
```suggestion
func requireHealthyCA(t *testing.T, caClient *ca.Client) {
        t.Helper()
	// Wait for CA
	time.Sleep(time.Second)
```

acme/challenge_test.go
dopey: Agreed. Maybe could be an internal field on the Challenge?

maraino: Fixed with 2b30ae50

dopey (approved):

hslatman (approved):
Great 😄

PR #1913, opened 2024-07-05
Fix HTTP internal server error when bad attestation object is provided

### LINKED ISSUE #1902: [Bug]: Device attestation validation returns 500 for invalid CBOR payload
### Steps to Reproduce

- **Fake reproducer**: Configure empty JSON in the payload part of the JWK for `/acme/acme/challenge/...` HTTP call.
- **Real reproducer**: Try to generate device certificate using device attestation on MacOS 15 Beta with INTEL processor with profile configuration attribute `HardwareBound=false` ([link](https://developer.apple.com/documentation/devicemanagement/acmecertificate)).

### Your Environment

* OS - MacOS 15 Beta
* `step-ca` Version - v0.26.2
* INTEL processor
* HardwareBound=false

### Expected Behavior

The server should return a client error -> status `400`.

### Actual Behavior

The server returns internal error -> status `500`.

Unexpected error: `error validating challenge: error unmarshalling CBOR: EOF`
Log message:
```
ERRO[6548] duration=33.057877ms duration-ns=33057877 error="error validating challenge: error unmarshalling CBOR: EOF" fields.time="2024-06-26T11:16:32+02:00" method=POST name=ca nonce=eHpuMGRaMlpMa3BoU0JUcDJwWnZSeVprRE44QmJ3Y1c path=/acme/acme/challenge/6zzEABv1oqdHExiAETjjR0RHVnD4hwVP/P9STwds3QRTt5CQhMCjU3ij4Uxs5cN6W protocol=HTTP/1.1 referer= remote-address=127.0.0.1 request-id=2dfcf8a3-c285-4fd0-80d5-9efc05688c82 response="{\"type\":\"urn:ietf:params:acme:error:serverInternal\",\"detail\":\"The server experienced an internal error\"}" size=105 status=500 user-agent="Apache-HttpClient/4.5.13 (Java/17.0.4)" user-id=
```

### Additional Context

_No response_

### Contributing

hslatman:
Fixes #1902 

Besides checking for an empty attestation object, which can be the case when a client is not able to provide an attestation and which should fail issuance, there's now an additional check for the CBOR data to be well formed. Effectively that check will be performed twice, as it's done during unmarshaling too, but this way it's easier to surface invalid CBOR data errors to the user. The alternative is to catch specific errors from the `cbor` package.

Instead of returning an error to the client immediately, the error is stored with the challenge instead, as the challenge won't be solved using the same request body again.

### DIFF
--- a/acme/challenge.go
+++ b/acme/challenge.go
@@ -1,6 +1,7 @@
 package acme
 
 import (
+	"bytes"
 	"context"
 	"crypto"
 	"crypto/ecdsa"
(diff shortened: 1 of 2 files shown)


dopey (approved):

maraino (approved):
lgtm

PR #1922, opened 2024-07-11
Create identity uri on any provisioner

maraino:
### Description

This commit allows the creation of the identity certificate with the host URI using any provisioner. Before, only the K8SSA provisioner could create an identity certificate with the URI.

### DIFF
--- a/api/ssh.go
+++ b/api/ssh.go
@@ -502,7 +502,7 @@ func SSHBastion(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
-// identityModifier is a custom modifier used to force a fixed duration. and set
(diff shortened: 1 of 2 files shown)


api/ssh.go
```diff
@@ -498,14 +502,42 @@ func SSHBastion(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
-// identityModifier is a custom modifier used to force a fixed duration.
+// identityModifier is a custom modifier used to force a fixed duration. and set
```
dopey, line 505:
```suggestion
// identityModifier is a custom modifier used to force a fixed duration, and set
```

dopey (approved):

PR #1960, opened 2024-08-13
Change `Wire` DB operations into using a runtime type assertion

hslatman:


hslatman:
@dopey this one should go into the next release too.

acme/challenge.go
```diff
@@ -117,9 +117,17 @@ func (ch *Challenge) Validate(ctx context.Context, db DB, jwk *jose.JSONWebKey,
 	case DEVICEATTEST01:
 		return deviceAttest01Validate(ctx, ch, db, jwk, payload)
 	case WIREOIDC01:
-		return wireOIDC01Validate(ctx, ch, db, jwk, payload)
+		wireDB, ok := db.(WireDB)
+		if !ok {
+			return NewErrorISE("db %T is not a WireDB", db)
```
maraino, line 122: Should we return something different, the RFC has an `unsupportedIdentifier` defined as `An identifier is of an unsupported type`

maraino: We can detect this earlier and return an `unsupportedIdentifier`, but at this point is in internal error.

maraino (approved):

Issue #1946, opened 2024-07-30
Vault RA mode supports AWS auth method


leonweecs:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently, the Vault RA mode of `step-ca` only supports `approle` and `kubernetes` auth methods.

We would like to add support for [AWS auth method](https://developer.hashicorp.com/vault/docs/auth/aws).

## Why is this needed?

At the moment, if we are using `step-ca` in Vault RA mode and not hosting it in `kubernetes`, the only way to authenticate to Vault is with `approle`.

`approle` auth method is considerably more complicated to implement right compared to platform identity auth (cloud provider IAM roles, Kubernetes service accounts, etc.).

In fact, Hashicorp recommends going with platform identity whenever possible, and if not, then `approle`.

> Applications authenticating through the platform is best, but not every platform has Vault auth integration. You might be in a cloud without any authentication plugin for Vault, or you might be using bare metal. Fortunately, Vault has an auth method that can give you many of the advantages of platform-based authentication even without native platform integration: AppRole lets you build a trusted broker for your applications easily and effectively. ([Source](https://www.hashicorp.com/blog/how-and-why-to-use-approle-correctly-in-hashicorp-vault))

## Notes

I am working on a PR to implement this.

leonweecs:
Feel free to assign this to me and I will be raising a PR soon.

maraino:
Hi @leonweecs, we are willing to merge AWS or other major cloud authentication in our vault CAS implementation. I've been looking a bit into it, and it looks quite straightforward. 

In general, a proper PR would be implemented in a new package, `github.com/smallstep/certificates/cas/vaultcas/auth/aws`, in a similar way as `kubernetes` and `approle` are. A new `AuthOptions` struct will contain specific configuration options for Vault, like `role`, and for AWS, like `region`.  Although you can set some of the AWS credentials like `accessKeyId` or `secretAccessKey` as environment variables, it probably makes sense to be able to use the custom `authOptions` to encode those. You should add unit tests, too, as much as possible.

Then, you should write some documentation in the PR so we can integrate it in https://github.com/smallstep/docs, although you can instead add a PR there.

I'll assign this to you. Let me know if you have any questions.

leonweecs:
[...]

PR #1976, opened 2024-08-28
Add AWS auth method for Vault RA mode

leonweecs:
Implements #1946

#### Name of feature:

Add AWS auth method for Vault RA mode.

The available `authOptions` maps directly to the Vault AWS auth [package](https://github.com/hashicorp/vault/blob/main/api/auth/aws/aws.go).

Please refer to #1946.

N/A

Yes, a separate PR will be raised for the docs repo.

https://github.com/smallstep/docs/pull/346

- AWS
- Vault RA mode

N/A

- #1946
- https://developer.hashicorp.com/vault/docs/auth/aws

💔Thank you!


cas/vaultcas/auth/aws/aws.go
```diff
@@ -0,0 +1,84 @@
...
+		default:
+			return nil, fmt.Errorf("unknown SignatureType type: %s, valid options are 'pkcs7', 'identity' and 'rsa2048'", opts.SignatureType)
+		}
+
+		if opts.Nonce != "" {
+			loginOptions = append(loginOptions, aws.WithNonce(opts.Nonce))
+		}
+	default:
+		return nil, fmt.Errorf("unknown awsAuthType: %s, valid options are 'iam' and 'ec2'", opts.AwsAuthType)
+	}
```
hslatman, line 76:
```suggestion
		return nil, fmt.Errorf("unknown awsAuthType %q; valid options are 'iam' and 'ec2'", opts.AwsAuthType)
	}
```

cas/vaultcas/auth/aws/aws.go
```diff
@@ -0,0 +1,84 @@
...
+			loginOptions = append(loginOptions, aws.WithPKCS7Signature())
+		case "identity":
+			loginOptions = append(loginOptions, aws.WithIdentitySignature())
+		case "rsa2048":
+			loginOptions = append(loginOptions, aws.WithRSA2048Signature())
+		case "":
+			// no-op
+		default:
+			return nil, fmt.Errorf("unknown SignatureType type: %s, valid options are 'pkcs7', 'identity' and 'rsa2048'", opts.SignatureType)
+		}
```
hslatman, line 69:
```suggestion
			return nil, fmt.Errorf("unknown SignatureType type %q; valid options are 'pkcs7', 'identity' and 'rsa2048'", opts.SignatureType)
		}
```

cas/vaultcas/auth/aws/aws_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+	return u, client
+}
+
+func TestAws_LoginMountPaths(t *testing.T) {
+	caURL, _ := testCAHelper(t)
+
+	config := vault.DefaultConfig()
+	config.Address = caURL.String()
+	client, _ := vault.NewClient(config)
```
hslatman, line 65: Is it necessary to create the `client` in both parts? It seems the second time it uses the same inputs, so it looks like the returned `client` would be OK to use too?

hslatman: Looks like it's the same for the other tests. Anyway, seems like the second time is not needed here.

hslatman:
Don't forget to fix the conflicts by rebasing or some other way

cas/vaultcas/auth/aws/aws_test.go
leonweecs: Yeah agree that it is redundant to create another `client`, we can use the one from the helper directly. I have updated accordingly.

hslatman (approved):

PR #2042, opened 2024-10-23
Allow Tokens for Renewing Certificates

hunoz:
None that I am aware of.

Any environments where this usage is planned or expected.

None

#### Supporting links/other PRs/issues: N/A

💔Thank you!

### DIFF
--- a/ca/provisioner.go
+++ b/ca/provisioner.go
@@ -118,6 +118,7 @@ func (p *Provisioner) Token(subject string, sans ...string) (string, error) {
 	return tok.SignedString(p.jwk.Algorithm, p.jwk.Key)
 }
 
+// RenewalToken generates a token for use with renewing certificates
 func (p *Provisioner) RenewalToken(subject string, sans ...string) (string, error) {
 	oldAudience := p.audience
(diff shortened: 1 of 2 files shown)


ca/provisioner.go
```diff
@@ -85,6 +85,7 @@ func (p *Provisioner) SetFingerprint(sum string) {
 
 // Token generates a bootstrap token for a subject.
 func (p *Provisioner) Token(subject string, sans ...string) (string, error) {
+	p.endpoint.ResolveReference(&url.URL{Path: "/1.0/renew"})
```
hunoz, line 88: Need to amend PR, forgot to remove this.

hunoz: Removed.

maraino (changes requested):
I don't really know what you're trying to do here. 

Without taking into account the rekeys, `step-ca` provides two different ways to renew an X.509 certificate:
* Using an mTLS connection with the certificate to renew.
* Making a request with an authorization token that contains the certificate embedded in the token and is signed by the private key of the certificate.

The token you are creating is a regular sign token with the audience changed. I would be surprised if that works.

PR #2044, opened 2024-10-25
Use dnsNamesSubsetValidator for IID provisioners

dopey:
... when disableCustomSANs is set to 'true'.

The DNS names in the certificate request must be a subset of the authorized set of DNS names (from the IID token). The previous functionality required that the DNS names in the certificate request exactly matched the authorized DNS names.


maraino (commented):
Herman's comments make sense, fix them and I'll approve it

maraino (approved):

PR #2065, opened 2024-11-12
Add data support on SCEPCHALLENGE webhooks

maraino:
### Description

This commit adds support for using template data from SCEPCHALLENGE webhooks.

This is a WIP

### DIFF
--- a/authority/provisioner/scep.go
+++ b/authority/provisioner/scep.go
@@ -15,6 +15,7 @@ import (
 
 	"go.step.sm/crypto/kms"
 	kmsapi "go.step.sm/crypto/kms/apiv1"
+	"go.step.sm/crypto/x509util"
 	"go.step.sm/linkedca"
 
 	"github.com/smallstep/certificates/webhook"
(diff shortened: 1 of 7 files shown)


authority/provisioner/sign_options.go
```diff
@@ -545,3 +545,28 @@ func (s csrFingerprintValidator) Valid(cr *x509.CertificateRequest) error {
 	}
 	return nil
 }
+
+// SignCSROption is the interface used to collect extra option in the SignCSR
+// method of the SCEP authority.
+type SignCSROption interface{}
```
hslatman, line 551:
```suggestion
// SignCSROption is the interface used to collect extra options in the SignCSR
// method of the SCEP authority.
type SignCSROption any
```

authority/provisioner/sign_options.go
```diff
@@ -545,3 +545,28 @@ func (s csrFingerprintValidator) Valid(cr *x509.CertificateRequest) error {
...
+
+// TemplateDataModifier in an interface that allows to modify template data.
+type TemplateDataModifier interface {
```
hslatman, line 554:
```suggestion
// TemplateDataModifier is an interface that allows to modify template data.
type TemplateDataModifier interface {
```

authority/provisioner/scep.go
```diff
@@ -145,25 +146,33 @@ var (
...
+			opts = append(opts, TemplateDataModifierFunc(func(data x509util.TemplateData) {
+				data.SetWebhook(wh.Name, resp.Data)
+			}))
 		}
 	}
 
-	return ErrSCEPChallengeInvalid
+	if len(opts) == 0 {
+		return nil, ErrSCEPChallengeInvalid
+	}
```
hslatman, line 173: Haven't checked out the implementation of `SetWebhook`, but I assume that it'll accept an empty `resp.Data`, and it'll effectively turn into a noop? It would still continue, as the `opts` now contains a noop modifier?

maraino: `data.SetWebhook(name, nil)` will create this:

```go
data["Webhooks"][name] = nil
```

The template could check if that webhook was called and assume the response was `{"allow": true}`.

authority/provisioner/sign_options.go
maraino: Fixed with ff37bf1

authority/provisioner/sign_options.go
maraino: Fixed with ff37bf1

hslatman (approved):

PR #2085, opened 2024-11-28
fix: allow active revocation on http revoke endpoint

VonRehberg:


hslatman:
Hey @VonRehberg, thank you for the PR.

Can you remove the test at https://github.com/smallstep/certificates/blob/af2e58ab28f3405456aa4276fe5e943ba0f272bc/api/revoke_test.go#L46-L53?

From the looks of it the change looks OK. It seems to stem from the time we only supported passive revocation, but since then support for active revocation was added. In practice the way that is configured and works does seem to make the `Passive` boolean obsolete, and it seems to be just informational at the moment. I think there's improvements that can be done, but that would require some more work to feed it through, and doesn't need to be in this PR.

tacerus:
Hi, as this seemed to be stale I continued in https://github.com/smallstep/certificates/pull/2327.

PR #2098, opened 2024-12-10
internal/httptransport: initial implementation of the package

azazeal:
Currently, there's extensive cloning of what's stored in `http.DefaultTransport`, under the assumption that what's stored there is an `http.Transport` reference.

There are many reasons why that might not be the case, especially since the type of `http.DefaultTransport` is `http.RoundTripper` and _not_ `http.Transport`.

To that end, this PR introduces the `internal/httptransport` package and refactors the `authority`, `ca` & `test` packages to use it.

### DIFF
--- a/authority/provisioner/webhook.go
+++ b/authority/provisioner/webhook.go
@@ -15,10 +15,13 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+
+	"go.step.sm/linkedca"
+
+	"github.com/smallstep/certificates/internal/httptransport"
 	"github.com/smallstep/certificates/middleware/requestid"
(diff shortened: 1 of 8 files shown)


authority/provisioner/webhook.go
```diff
@@ -15,10 +15,12 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"go.step.sm/linkedca"
```
hslatman, line 18: nit
```suggestion
	"github.com/pkg/errors"
	
	"go.step.sm/linkedca"
```

areed (approved):

PR #2103, opened 2024-12-12
Transport wrappers

azazeal:
This PR adds references to a transport wrapper that callers may pass in order to inject their own dependecies to references of HTTP clients that the internal packages generate.

### DIFF
--- a/authority/options.go
+++ b/authority/options.go
@@ -18,6 +18,7 @@ import (
 	"github.com/smallstep/certificates/cas"
 	casapi "github.com/smallstep/certificates/cas/apiv1"
 	"github.com/smallstep/certificates/db"
+	"github.com/smallstep/certificates/internal/httptransport"
 	"github.com/smallstep/certificates/scep"
 )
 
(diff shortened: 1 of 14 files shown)


authority/provisioner/webhook.go
```diff
@@ -141,7 +142,11 @@ type Webhook struct {
 	} `json:"-"`
 }
 
-func (w *Webhook) DoWithContext(ctx context.Context, client *http.Client, reqBody *webhook.RequestBody, data any) (*webhook.ResponseBody, error) {
+// TransportWrapper wraps the set of functions mapping [http.Transport] references to
+// [http.RoundTripper].
+type TransportWrapper = httptransport.Wrapper
```
areed, line 147: What is this for?

authority/provisioner/provisioner.go
```diff
@@ -264,6 +264,9 @@ type Config struct {
 	// HTTPClient is an HTTP client that trusts the system cert pool and the CA
 	// roots.
 	HTTPClient *http.Client
+	// WrapTransport references the function that should wrap any [http.Transport] initialized
+	// down the Config's chain.
+	WrapTransport func(*http.Transport) http.RoundTripper
```
areed, line 269: Why not `httptransport.Wrapper`?

authority/options.go
```diff
@@ -103,6 +104,22 @@ func WithWebhookClient(c *http.Client) Option {
 	}
 }
 
+// Wrapper wraps the set of functions mapping [http.Transport] references to [http.RoundTripper].
+type TransportWrapper = httptransport.Wrapper
```
areed, line 108: What's this for?

areed (approved):

authority/provisioner/webhook.go
azazeal: The `httptransport` package is an internal one, to avoid duplication from package to package. Since the function's signature is inside function definitions I thought to compress the signature by declaring a type alias.

authority/provisioner/provisioner.go
azazeal: Well, it'll be a type alias but it's coming right up.

authority/options.go
azazeal: See above the above.

authority/provisioner/provisioner.go
azazeal: 06936791

dopey (approved):

PR #2131, opened 2025-01-07
Wait and retry connection to test CA server instead of failing (immediately)

hslatman:
The integration test checking for request IDs to be reflected depends on a test CA server to be started and running. At times (see [this case](https://github.com/smallstep/certificates/actions/runs/12642283670/job/35226696954?pr=2129) with a few attempts) the CA healthcheck would fail the test, presumably because the CA server was not running yet. This PR checks the connection a few times before failing the test.

### DIFF
--- a/test/integration/requestid_test.go
+++ b/test/integration/requestid_test.go
@@ -50,6 +50,8 @@ func reservePort(t *testing.T) (host, port string) {
 }
 
 func Test_reflectRequestID(t *testing.T) {
+	ctx := context.Background()
+
 	dir := t.TempDir()
 	m, err := minica.New(minica.WithName("Step E2E"))
 	require.NoError(t, err)
@@ -133,8 +135,11 @@ func Test_reflectRequestID(t *testing.T) {
 		require.ErrorIs(t, err, http.ErrServerClosed)
 	}()
 
+	// require the CA server to be available within 10 seconds,
+	// failing the test if it doesn't.
+	requireCAServerToBeAvailable(t, net.JoinHostPort("localhost", port), 10*time.Second)
+
 	// require OK health response as the baseline
-	ctx := context.Background()
 	healthResponse, err := caClient.HealthWithContext(ctx)
 	require.NoError(t, err)
 	if assert.NotNil(t, healthResponse) {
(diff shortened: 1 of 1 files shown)


test/integration/requestid_test.go
```diff
@@ -287,3 +292,30 @@ func newAuthorizingServer(t *testing.T, mca *minica.CA) *httptest.Server {
...
+func canConnect(ctx context.Context, address string) bool {
+	d := net.Dialer{Timeout: 5 * time.Second}
+	conn, err := d.DialContext(ctx, "tcp", address)
+	if err != nil {
+		return false
+	}
+
+	conn.Close()
+
+	return true
```
marten-seemann, line 320: It's faster if you just re-dial every 100ms, since you avoid running into TCP's exponential backoff.

Also, if the server isn't bound to the port yet, the kernel might send you a RST in response to the SYN immediately, which would make `DialContext` fail immediately.

test/integration/requestid_test.go
```diff
@@ -287,3 +292,30 @@ func newAuthorizingServer(t *testing.T, mca *minica.CA) *httptest.Server {
...
+		select {
+		case <-ctx.Done():
+			require.FailNow(t, fmt.Sprintf("CA server failed to start at https://%s within %s", address, timeout.String()))
+		case <-time.After(100 * time.Millisecond):
+		}
+	}
+}
+
+func canConnect(ctx context.Context, address string) bool {
+	d := net.Dialer{Timeout: 5 * time.Second}
```
marten-seemann, line 312: Why hardcode a timeout, if you're dialing with a context anyway?

test/integration/requestid_test.go
```diff
@@ -287,3 +292,30 @@ func newAuthorizingServer(t *testing.T, mca *minica.CA) *httptest.Server {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	for !canConnect(ctx, address) {
+		select {
+		case <-ctx.Done():
+			require.FailNow(t, fmt.Sprintf("CA server failed to start at https://%s within %s", address, timeout.String()))
+		case <-time.After(100 * time.Millisecond):
+		}
```
marten-seemann, line 307: Not sure I understand this `select` statement. Why sleep for 100ms and then return?

test/integration/requestid_test.go
hslatman: Fair point. Added it out of habit. b3fb92756fe66cdd6b19267c4718b8eb59cf9850.

test/integration/requestid_test.go
hslatman: Returning an error immediately is OK, and kind of what I want. The CA server is started (or not) in a different goroutine. The goal is to wait for it to become available before continuing the test, and a failure to connect here indicates it's not yet available.

test/integration/requestid_test.go
hslatman: After 100ms the connection is retried. If it succeeds, the `for` loop is exited, and we continue the test. If it fails to connect at least once before the `ctx` times out, the entire test is failed.

marten-seemann (approved):

PR #2133, opened 2025-01-10
feat(gcp): enable organization validation

ericnorris:
Hey all, I'm submitting a PR to enable validating that a project is a part of a GCP organization, rather than a static list of project IDs. As I mention in the commit message, I tried to strike the right balance between production-ready and proof-of-concept, so feel free to leave as much feedback as possible since I'm open to changing anything.

I'm going to share the first commit message below:

---

Before this commit, users could specify a hardcoded list of project IDs to restrict access to the GCP provisioner. While this works, it can be both toilsome to the team maintaining the Smallstep installation and unintuitive to the internal infrastructure users that may encounter errors as a result of their project not being added.

This commit is a rough attempt at adding support for validating that a GCP project belongs to a given GCP organization. It does this by using the `projects.getAncestry` call in the Cloud Resource Manager API. If a token's project claim does not have the given organization ID as its topmost ancestor, the token is rejected. This will require the `resourcemanager.projects.get` IAM permission on the organization.

The new `OrganizationID` configuration directive is compatible with the existing `ProjectIDs` configuration. If `ProjectIDs` is non-empty, it will take precedence over the `OrganizationID` and act as it did before, with the minor difference that if `OrganizationID` is also non-empty, the provisioner will check the project's ancestry before rejecting the token.

There are a couple outstanding questions and tasks after this commit. I tried to strike the right balance between production-ready and proof-of-concept here, so I'm open to any suggestions.

- Is the `authority/provisioner/gcp` package the right place for adding this functionality? Is the new struct the right approach?
- We should add tests for validating the organization ID.
- How should users configure the authentication for the Cloud Resource Manager client? I expect this would be similar to the Cloud KMS integration.
- Does Smallstep Professional run in an environment that will be able to authenticate with Google? We would need to either grant permissions to a Smallstep-owned Google service account if it's run in GCP, or set up something like Google's Workload Identity Federation to handle a K8s, AWS, or Azure deployment.

### DIFF
--- a/authority/provisioner/gcp/projectvalidator.go
+++ b/authority/provisioner/gcp/projectvalidator.go
@@ -1,0 +1,78 @@
+package gcp
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/smallstep/certificates/errs"
+
+	"google.golang.org/api/cloudresourcemanager/v1"
+)
+
+type ProjectValidator struct {
+	ProjectIDs []string
+}
+
+func (p *ProjectValidator) ValidateProject(_ context.Context, projectID string) error {
+	if len(p.ProjectIDs) == 0 {
+		return nil
+	}
+
+	for _, pi := range p.ProjectIDs {
+		if pi == projectID {
+			return nil
+		}
+	}
+
+	return errs.Unauthorized("gcp.authorizeToken; invalid gcp token - invalid project id")
+}
+
+type OrganizationValidator struct {
+	*ProjectValidator
+
+	OrganizationID  string
+	projectsService *cloudresourcemanager.ProjectsService
+}
+
+func NewOrganizationValidator(ctx context.Context, projectIDs []string, organizationID string) (*OrganizationValidator, error) {
+	crm, err := cloudresourcemanager.NewService(ctx)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &OrganizationValidator{
+		&ProjectValidator{projectIDs},
+		organizationID,
+		crm.Projects,
+	}, nil
+}
+
+func (p *OrganizationValidator) ValidateProject(ctx context.Context, projectID string) error {
+	err := p.ProjectValidator.ValidateProject(ctx, projectID)
+
+	if p.OrganizationID == "" {
+		return err
+	}
+
+	ancestry, err := p.projectsService.
+		GetAncestry(projectID, &cloudresourcemanager.GetAncestryRequest{}).
+		Context(ctx).
+		Do()
+
+	if err != nil {
+		return errs.Wrap(http.StatusInternalServerError, err, "gcp.authorizeToken")
+	}
+
+	if len(ancestry.Ancestor) < 1 {
+		return errs.InternalServer("gcp.authorizeToken; getAncestry response malformed")
+	}
+
+	progenitor := ancestry.Ancestor[len(ancestry.Ancestor)-1]
+
+	if progenitor.ResourceId.Type != "organization" || progenitor.ResourceId.Id != p.OrganizationID {
+		return errs.Unauthorized("gcp.authorizeToken; invalid gcp token - project does not belong to organization")
+	}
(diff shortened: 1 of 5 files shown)


authority/provisioner/gcp/projectvalidator.go
```diff
@@ -0,0 +1,78 @@
...
+		crm.Projects,
+	}, nil
+}
+
+func (p *OrganizationValidator) ValidateProject(ctx context.Context, projectID string) error {
+	err := p.ProjectValidator.ValidateProject(ctx, projectID)
+
+	if p.OrganizationID == "" {
+		return err
+	}
```
maraino, line 56: Why is the project validation only enforced when the organizationID is not set?

maraino (commented):
Hi, @ericnorris I have a question I would like to clarify

authority/provisioner/gcp/projectvalidator.go
maraino: It makes more sense to return an error even if the organization id is set. 

I don't know if they are mutually exclusive, but if the project id is in the token and it doesn't match the ones in my configuration, I would expect an error.

maraino: Can you make that change?

ericnorris: I can make that change, but let me explain - if you set the `projectIDs` _and_ `organizationID`, would you not be confused that it was rejected when a token was in fact inside the organization?

That's what I meant by them being mutually exclusive, if it worked the way you described, you could set either `projectIDs` or `organizationID`, but not both. I was thinking about how if someone had set `projectIDs` today, and wanted to try out the `organizationID` setting, rather than completely disabling `projectIDs` and enabling `organizationID`, they could enable `organizationID`, try it out with a few projects, and _then_ remove `projectIDs`.

Or maybe they want to allow their organization and one or a few projects from a separate organization? This would also require them to be complementary.

Again though, this is not something I personally feel strongly about, but my hunch is that this behavior is more user-friendly.

maraino: Hi @ericnorris, I'm not convinced, I still think both must match if they are set. To try the organization setting, you can create a new provisioner without project IDs. 

> Or maybe they want to allow their organization and one or a few projects from a separate organization? This would also require them to be complementary.

If they want to allow some projects from one organization, but all the ones in another organization. They should create two provisioners, one with the projects with or without organization, and a second one with just the second organization.

If we want to allow two organizations, with your change we will need two provisioners, I'm ok with that. At some point, we can also change the settings to be something like the `multiString` in https://github.com/smallstep/certificates/blob/cf1ef2a77dfab39389423bfa453cd637e8cd1e0c/authority/config/types.go#L11

A type that accepts both string and []string. So it can be backward compatible.

ericnorris: > Hi @ericnorris, I'm not convinced, I still think both must match if they are set. To try the organization setting, you can create a new provisioner without project IDs.

Fair, but then should we make this option officially mutually exclusive with `projectIDs` then? Why even set the organization ID if you've set a project ID list? The former is what actually matters, and allowing `organizationID` to be set at the same time is misleading since it's redundant.

maraino: As the project ID is globally unique, yes it makes sense to not use both at the same time. But I don't mind to check both. I leave that decission to you.

ericnorris: done!

ericnorris:
Hey! I appreciate seeing activity on this pull request - one thing to note is that it might be worth adding caching to the `OrganizationValidator`, since once you verify a project ID you can pretty much assume it will never move. Without caching, I believe there's a chance to hit API rate limits depending on the volume of provisioner requests.

I didn't add this when I wrote it because I wanted to get feedback before committing more effort.

PR #2167, opened 2025-02-11
Fix new `golangci-lint` issues

hslatman:


internal/cast/cast_test.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestUintConvertsValues(t *testing.T) {
+	require.Equal(t, uint(0), Uint(0))
+	require.Equal(t, uint(math.MaxInt64), Uint(math.MaxInt64))
```
azazeal, line 12: This will break on non-x64 platforms. You should probably swap `math.MaxInt64` for `math.MaxInt`.

azazeal (changes requested):

azazeal (approved):

PR #2187, opened 2025-02-24
Fix CI

hslatman:
Replaces #2185.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/smallstep/certificates
 
 go 1.22.12
+toolchain go1.23.6
 
 require (
 	cloud.google.com/go/longrunning v0.6.4
@@ -38,7 +39,7 @@ require (
 	github.com/urfave/cli v1.22.16
 	go.step.sm/crypto v0.58.0
 	go.uber.org/mock v0.5.0
-	golang.org/x/crypto v0.33.0
+	golang.org/x/crypto v0.34.0
 	golang.org/x/exp v0.0.0-20240318143956-a85f2c67cd81
 	golang.org/x/net v0.35.0
 	google.golang.org/api v0.221.0
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,6 @@
 module github.com/smallstep/certificates
 
-go 1.22.12
-toolchain go1.23.6
+go 1.23.6
 
 require (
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/smallstep/certificates
 
-go 1.22.12
+go 1.23.6
```
azazeal, line 3: Won't `1.23` just do?

hslatman: Only if `toolchain go1.23.5` is added, which happens with `go mod tidy`. But that is not sufficient, as we need `.6` to not use a Go version with a vulnerability, so it should be `toolchain go1.23.6`. 

Using `go 1.23.6` is just the simplest.

azazeal: The only thing bumped in this file is `x/crypto` and I see that the version this needs ([ref](https://cs.opensource.google/go/x/crypto/+/master:go.mod)) is `1.23.0`.

Where's this bump coming from? Any idea?

hslatman: Well, I made it `1.23.6`.

Also, as mentioned:

```console
$ make lint
=== Symbol Results ===

Vulnerability #1: GO-2025-3447
    Timing sidechannel for P-256 on ppc64le in crypto/internal/nistec
  More info: https://pkg.go.dev/vuln/GO-2025-3447
  Standard library
    Found in: crypto/internal/nistec@go1.23.5
    Fixed in: crypto/internal/nistec@go1.23.6
    Platforms: ppc64le
    Example traces found:
      #1: ca/client.go:1419:30: ca.CreateSignRequest calls ecdsa.GenerateKey, which eventually calls nistec.P256Point.ScalarBaseMult
      #2: acme/challenge.go:1474:23: acme.doStepAttestationFormat calls ecdsa.VerifyASN1, which eventually calls nistec.P256Point.ScalarMult
      #3: test/integration/scep/internal/x509/parser.go:279:29: x509.parsePublicKey calls elliptic.Unmarshal, which eventually calls nistec.P256Point.SetBytes

Your code is affected by 1 vulnerability from the Go standard library.
This scan also found 0 vulnerabilities in packages you import and 2
vulnerabilities in modules you require, but your code doesn't appear to call
these vulnerabilities.
Use '-show verbose' for more details.
```

azazeal (approved):

PR #2191, opened 2025-02-25
Add option to specify a client timeout

### LINKED ISSUE #2176: ca.NewClient don't allow setting a timeout for the http client
## Hello!

## Issue details

ca.NewClient don't allow setting a timeout for the http client.

## Why is this needed?

If an http client don't have a timeout, it can potentially wait forever, setting one is a common best practice.

But the client in the `ca` don't don't provide any way of setting it.

maraino:
### Description 

This commit adds a new option to specify the client timeout to the CA client.

Fixes #2176

### DIFF
--- a/ca/client.go
+++ b/ca/client.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"golang.org/x/net/http2"
@@ -53,10 +54,11 @@ type uaClient struct {
 	Client *http.Client
 }
 
-func newClient(transport http.RoundTripper) *uaClient {
+func newClient(transport http.RoundTripper, timeout time.Duration) *uaClient {
 	return &uaClient{
 		Client: &http.Client{
 			Transport: transport,
+			Timeout:   timeout,
 		},
 	}
 }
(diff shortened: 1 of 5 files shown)


ca/client.go
```diff
@@ -568,9 +582,10 @@ func NewClient(endpoint string, opts ...ClientOption) (*Client, error) {
...
 
 	return &Client{
-		client:    newClient(tr),
+		client:    newClient(tr, o.timeout),
 		endpoint:  u,
 		retryFunc: o.retryFunc,
+		timeout:   o.timeout,
 		opts:      opts,
 	}, nil
 }
```
hslatman, line 591: Can you also set a default timeout? I believe without a default timeout set, it will just wait until the server disconnects based on its timeouts (if any). See the output for different cases below. I know this client will be used with our CA, and we're in control over those timeouts, but it's generally a good practice to set a timeout in any `http.Client` that is created. 

The `WithContext` methods can also be used, and they seem to behave as expected. However, we're not setting timeouts through the context in places where we use those methods. 

```console
# timeout on connection triggered by server; no (default) timeout in the HTTP client or context:
$ go run cmd/step/main.go ca health 
client GET https://127.0.0.1:8443/health failed: stream error: stream ID 1; INTERNAL_ERROR; received from peer

# timeout on the context from the client side:
$ go run cmd/step/main.go ca health
client GET https://127.0.0.1:8443/health failed: context deadline exceeded
exit status 1

# timeout on the client:
$ go run cmd/step/main.go ca health
client GET https://127.0.0.1:8443/health failed: context deadline exceeded (Client.Timeout exceeded while awaiting headers)
exit status 1
```

hslatman (approved):

dopey (approved):

PR #2271, opened 2025-05-15
Use container secrets securely

### LINKED ISSUE #2270: Use container secrets securely
## Hello!

## Issue details

When running a container with `DOCKER_STEPCA_INIT_PASSWORD_FILE` set the Docker `entrypoint.sh` [`cat`'s the password file](https://github.com/smallstep/certificates/blob/30e79a2b80c8ab9aecb9d98972812c8a5ffc0a26/docker/entrypoint.sh#L50) to `/home/step/secrets/password` & the file is world readable:
```
stepca:~$ ls -l secrets
total 516
-rw-------    1 step     step           314 May 14 00:03 intermediate_ca_key
-rw-r--r--    1 step     step        511875 May 14 00:02 password
-rw-------    1 step     step           314 May 14 00:03 root_ca_key
-rw-------    1 step     step           314 May 14 00:03 ssh_host_ca_key
-rw-------    1 step     step           314 May 14 00:03 ssh_user_ca_key
```

## Why is this needed?

* For containers at least, a `symlink` is a better solution (which also works perfectly) - as the `secret` will usually be mounted in the container under `/run/secrets/xxx` with explicit permissions (possible in both `docker` & `podman`). This will work with the Dockerfile expecting to find the hardcoded `/home/step/secrets/password` & a container `secret` named anything.

* Moving the symlink creation out of `init` to run every time `DOCKER_STEPCA_INIT_PASSWORD_FILE`  is set is probably also recommended for cases where the container is run with a different password file location.

* Am willing to make a PR if you agree with this approach. 

---

* Running `stepca` with a `378,000` char `podman` secret works ok - [example](https://docs.podman.io/en/latest/markdown/podman-secret-create.1.html#examples). 

* I can also add to the PR a podman `quadlet` example `.container` file & `run` command.

itoffshore:
* increases the default password length from `40` to `256` characters
* adds a new `ENV` variable `DOCKER_STEPCA_PASSWORD_FILE` so the password file location can be changed / set on every `run`
* adds `set_password_files()` to `entrypoint.sh` so `/home/step/secrets/password` becomes a `symlink` in containers pointing to `DOCKER_STEPCA_INIT_PASSWORD_FILE` & also `DOCKER_STEPCA_PASSWORD_FILE` so secret file permissions are retained
* adds podman example quadlet / run command with a `378,000` character secret
* small update to `README.md` for new `podman` examples / `docker` examples

Fixes #2270

---

#### Supporting links/other PRs/issues: #2270

💔Thank you!


itoffshore:
did a bit more testing:

- `378,000` character `base64` string from `openssl` doesn't work (I don't think the input expects so many newlines) 
- successfully tested & changed the examples to `8096` character `hex` string from `openssl`

May also add an example for creating a cert & key for `postfix` after a bit more testing

itoffshore:
added an example for a Quantum Resistant CA (16384 bit RSA)

docker/entrypoint.sh
```diff
@@ -27,11 +27,18 @@ function init_if_possible () {
 
 function generate_password () {
     set +o pipefail
-    < /dev/urandom tr -dc A-Za-z0-9 | head -c40
+    < /dev/urandom tr -dc A-Za-z0-9 | head -c256
```
jdoss, line 30: Let's not increase the password length here. 40 characters is long enough.

docker/entrypoint.sh
```diff
@@ -27,11 +27,18 @@ function init_if_possible () {
...
     echo
     set -o pipefail
 }
 
+function set_password_files () {
+    local file_path="$1"
```
jdoss, line 36: Can you please make your bash variables in the following format:

```bash
FOO=bar
echo "${FOO}"
```

Declare them in all caps and then when referencing them enclose them in quotations and squiggly brackets.

docker/entrypoint.sh
```diff
@@ -27,11 +27,18 @@ function init_if_possible () {
...
+
+    ln -sf "$file_path" "${STEPPATH}/password"
```
jdoss, line 38: Update variable formatting.

docker/entrypoint.sh
```diff
@@ -27,11 +27,18 @@ function init_if_possible () {
...
+    ln -sf "$file_path" "${STEPPATH}/provisioner_password"
```
jdoss, line 39: Update variable formatting.

examples/podman/stepca.container.md
```diff
@@ -0,0 +1,45 @@
...
+DNSOption=~custom.domain
+
+[Service]
+Restart=always
+
+[Install]
+WantedBy=default.target
+```
+
+* Contributed by: [Stuart Cardall](https://github.com/itoffshore)
```
jdoss, line 45: Please remove this from documentation. The attribution of your work is in the commit.

examples/podman/stepca.run.md
```diff
@@ -0,0 +1,47 @@
...
+    --env "DOCKER_STEPCA_PASSWORD_FILE=/run/secrets/$ctr" \
+    --cap-drop ALL \
+    --restart always \
+    --privileged \
+    --label "io.containers.autoupdate=registry" \
+    -v ${HOME}/volumes/$ctr/config:/home/step:Z \
+$repo
+```
+
+* Contributed by: [Stuart Cardall](https://github.com/itoffshore)
```
jdoss, line 47: Please remove this from documentation. The attribution of your work is in the commit.

examples/podman/stepca.container.md
```diff
@@ -0,0 +1,45 @@
...
+Environment=TZ="Europe/London"
+Environment="DOCKER_STEPCA_INIT_NAME=Example CA"
+Environment=DOCKER_STEPCA_INIT_DNS_NAMES=ca.custom.domain,10.89.0.10,localhost,127.0.0.1
+Environment=DOCKER_STEPCA_INIT_PROVISIONER_NAME=admin@custom.domain
+Environment=DOCKER_STEPCA_INIT_SSH=true
+Environment=DOCKER_STEPCA_INIT_ACME=true
+Environment=DOCKER_STEPCA_INIT_PASSWORD_FILE=/run/secrets/stepca
+HostName=stepca
+PodmanArgs=--privileged
+# Alpine image
```
jdoss, line 27: Let's just remove the Alpine and TPM container example comments. If a user is going to customize what image they are going to use they will do so and the examples we are giving don't give any context on why they would use something different. 

For example, if they want to use a TPM they need to pass in the TPM device as well and also configure step ca to use it. Providing an alternative to the container image here could lead to confusion and/or a broken setup.

examples/podman/stepca.container.md
```diff
@@ -0,0 +1,45 @@
...
+After=network-online.target
+
+[Container]
+PodmanArgs=--memory 25m --cpus 0.20
+PidsLimit=60
+DropCapability=ALL
+AutoUpdate=registry
+ContainerName=stepca
+DropCapability=ALL
+Environment=TZ="Europe/London"
```
jdoss, line 18: Let's use `TZ="UTC"` as a default.

examples/podman/stepca.container.md
```diff
@@ -0,0 +1,45 @@
...
+Environment="DOCKER_STEPCA_INIT_NAME=Example CA"
+Environment=DOCKER_STEPCA_INIT_DNS_NAMES=ca.custom.domain,10.89.0.10,localhost,127.0.0.1
+Environment=DOCKER_STEPCA_INIT_PROVISIONER_NAME=admin@custom.domain
+Environment=DOCKER_STEPCA_INIT_SSH=true
+Environment=DOCKER_STEPCA_INIT_ACME=true
+Environment=DOCKER_STEPCA_INIT_PASSWORD_FILE=/run/secrets/stepca
+HostName=stepca
+PodmanArgs=--privileged
```
jdoss, line 26: Remove this. You shouldn't need to step ca with the `--privileged` flag unless you are trying to access a TPM and your rootless user isn't in the `tss` group.

jdoss (changes requested):
Thank you so much for this PR @itoffshore. You get bonus points for including Podman examples. 👏🏼👏🏼  I use Podman with Quadlets for all of my homelab container deployments, including step ca.

I have left a detailed review. If you can make the requested changes that would be great, and we can get this merged.

examples/podman/stepca.container.md
itoffshore: I need `--privileged` to start a **rootful** podman container (just tested it again) - without it I see on the Alpine image:
```
May 29 01:02:30 endeavour systemd[1]: Started Smallstep Certificate Authority.
May 29 01:02:30 endeavour stepca[2043603]: 0a072b7f7061811ccde258488e80069834f50c55f782d52c848e900c13d80144
May 29 01:02:30 endeavour stepca[2043692]: /bin/sh: exec: line 0: /usr/local/bin/step-ca: Operation not permitted
May 29 01:02:30 endeavour podman[2043752]: 2025-05-29 01:02:30.248589806 +0100 BST m=+0.022680466 container died 0a072b7f7061811ccde258488e80069834f50c55f782d52c848e900c13d80144 (image=docker.io/smallstep/step-ca:latest, name=stepca, io.containers.autoupdate=registry, PODMAN_SYSTEMD_UNIT=stepca.service)
```
I'm using a podman secret inside `/root` with an `8192` char password (this seemed about the same security as using a TPM & can be done automated)

docker/entrypoint.sh
itoffshore: force pushed fixing everything

docker/entrypoint.sh
itoffshore: force pushed fixing everything

docker/entrypoint.sh
itoffshore: fixed these too ;o)

examples/podman/stepca.container.md
itoffshore: switched to UTC

itoffshore:
> Thank you so much for this PR @itoffshore. You get bonus points for including Podman examples. 👏🏼👏🏼 I use Podman with Quadlets for all of my homelab container deployments, including step ca.
> 
> I have left a detailed review. If you can make the requested changes that would be great, and we can get this merged.

all the changes should be there

Quadlets are great - [on to pods next](https://thkukuk.de/blog/prometheus_grafana/)

if you love `podman` check out [MicroOS](https://microos.opensuse.org/) as a container host - I switched to it on my stuff about a year ago. Rolling release with self correcting upgrades (btrfs snapshots) 

Making customised iso images with "[kiwi](https://github.com/OSInside/kiwi)" is great. RKE2 can be run on it very easily.

---

* Possibly the issue with the Alpine image when not run with `--privileged`:

alpine-stepca

* perhaps put the image binaries under `/usr/bin` (where Alpine expects them to be) - in Alpine ordinary users are quite locked down (no `ping`), something doesn't like `/usr/local/bin` executing early in startup.

examples/podman/stepca.run.md
```diff
@@ -0,0 +1,45 @@
+## Example creation of a Podman container & secret
+
+* Using a [cryptographically strong secret from `openssl` with an `8192` character `hex` string](https://docs.podman.io/en/latest/markdown/podman-secret-create.1.html#examples)
+
+see also:
+
+- [Create a "quadlet"](https://github.com/containers/podlet)
+- [Netbird VPN](https://github.com/netbirdio/netbird)
```
hslatman, line 8: Why is this here? There's only a comment using `netbird` below, but it doesn't seem to be required for this example.

hslatman: My comment was specifically for Netbird. I understand that running the CA on a VPN interface is a nice way to add additional security, but IMO the link to `netbird` isn't required.

itoffshore: ok will remove the link after work & resolve

itoffshore: added a `mkdir -p` to the `run` example for persistent storage

itoffshore: & a link back to your [TPM Blog](https://smallstep.com/blog/trusted-platform-modules-tpms/) in a note about `--privileged`

itoffshore: & `ReloadSignal=SIGHUP` to the quadlet

reload

itoffshore: * also removed the `finger.print` file that got accidentally added to the PR

docker/entrypoint.sh
```diff
@@ -46,9 +53,8 @@ function step_ca_init () {
         --provisioner-password-file "${STEPPATH}/provisioner_password"
         --address "${DOCKER_STEPCA_INIT_ADDRESS}"
     )
-    if [ -n "${DOCKER_STEPCA_INIT_PASSWORD_FILE}" ]; then
-        cat < "${DOCKER_STEPCA_INIT_PASSWORD_FILE}" > "${STEPPATH}/password"
-        cat < "${DOCKER_STEPCA_INIT_PASSWORD_FILE}" > "${STEPPATH}/provisioner_password"
+    if [ -f "${DOCKER_STEPCA_INIT_PASSWORD_FILE}" ]; then
+	set_password_files "${DOCKER_STEPCA_INIT_PASSWORD_FILE}"
```
tashian:
```suggestion
	    set_password_files "${DOCKER_STEPCA_INIT_PASSWORD_FILE}"
```

PR #2284, opened 2025-05-26
Update TLS duration settings and add tests for provisioner storage er…

marioja:
…rors

#### Name of feature:
Added test for the StoreProvisioner method

💔Thank you!


dopey:
Hey @marioja 👋. Thanks for opening the issue!

I understand the additional test coverage for the StoreProvisioner method - thanks for adding that!

Can you please explain the changes in the testAuthority function? https://github.com/smallstep/certificates/pull/2284/files#diff-efd742a0f56eda5c2aba7508dc0dbac4ffbbe8b66bee36875715a1dcf035d33cR77-R78 -- I believe the `min` and `max` were purposely set to invalid values so that the provisioner would not be initialized.

PR #2322, opened 2025-06-30
Add android-key attestation format support

jbpin:
Add a new attestationFormat for ACME device-attest-01 challenge to support Android attestation (android-key) as defined by WebAuthn and modify by RFC (sig use key authorization).

The implementation involve adding a CRL. ACME provider support a new configuration key called RootCRLs (rootCRLs in json). When 'android-key' is specified in attestationFormat and the list is not provided by the configuration, the list will be populated and updated
automatically based on the validation implementation procedure. Other ACME challenge could use IsRootRevoked and RootCRLs in the future independantly to android-key or device-attest-01 challenge.

#### Name of feature: 
android-key attestation support for device-attest-01 challenge

All to use device identifier certificate issuing on Android devices

The project already support Apple, so supporting Android is a plus. 

This PR only provide a new attestationFormat called android-key 

ACME provisionner with device_attest_01 and attestationFormat android-key. The device need to be managed by a MDM and an ACME client for android should be deployed

💔Thank you!


jbpin:
Hi, 
This PR can be test with a provisioner config like this : 
```
{
            "attestationFormats": [
              "android-key"
            ],
            "challenges": [
              "device-attest-01"
            ],
            "claims": {
              "defaultTLSCertDuration": "2h",
              "maxTLSCertDuration": "8h"
            },
            "name": "android-acme",
            "type": "ACME"
          },
```

However, this PR will require an update to linkedca for two reason: 
- `android-key` is not supported on attestionFormat [in the proto].(https://github.com/smallstep/linkedca/blob/339b487a925e0cf176a55133a755a1a40e602c35/spec/linkedca/provisioners.proto#L149).
- the new RootCRLs can't be included in ether ca.json nor step cli, so we can't override the default behavior that consist of loading the Android CRL Json

The CRL List is cached for 24h and refreshed when android-key attestation validation occurred passed this delay. 

I will be happy to reply to any comments or suggestions.

acme/challenge.go
```diff
@@ -1367,6 +1399,195 @@ func doAppleAttestationFormat(_ context.Context, prov Provisioner, _ *Challenge,
...
+			return nil, WrapError(ErrorBadAttestationStatementType, err, "error parsing attestation")
+		}
+		data.Attestation = keyDesc
+		break
+	}
+
+	// validate challenge
+	if string(data.Attestation.AttestationChallenge) != keyAuth {
+		return nil, NewDetailedError(ErrorBadAttestationStatementType, "Challenge mismatach: "+string(data.Attestation.AttestationChallenge))
+	}
```
hslatman, line 1586:
```suggestion
	// validate challenge
	if string(data.Attestation.AttestationChallenge) != keyAuth {
		return nil, NewDetailedError(ErrorBadAttestationStatementType, fmt.Sprintf("challenge mismatch; expected %q, got %q", keyAuth, string(data.Attestation.AttestationChallenge))
	}
```

acme/challenge.go
```diff
@@ -1367,6 +1399,195 @@ func doAppleAttestationFormat(_ context.Context, prov Provisioner, _ *Challenge,
...
+		keyDesc, err := attestation.ParseExtension(ext.Value)
+		if err != nil {
+			return nil, WrapError(ErrorBadAttestationStatementType, err, "error parsing attestation")
+		}
+		data.Attestation = keyDesc
+		break
+	}
+
+	// validate challenge
+	if string(data.Attestation.AttestationChallenge) != keyAuth {
```
hslatman, line 1584: Consider using `subtle.ConstantTimeCompare` to compare these.

acme/challenge.go
```diff
@@ -1367,6 +1399,195 @@ func doAppleAttestationFormat(_ context.Context, prov Provisioner, _ *Challenge,
...
+
+	block, _ := pem.Decode([]byte(AndroidRootCAPubKey))
+	trustedPubKey, err := x509.ParsePKIXPublicKey(block.Bytes)
+	switch root.PublicKey.(type) {
+	case *rsa.PublicKey:
+		if !root.PublicKey.(*rsa.PublicKey).Equal(trustedPubKey) {
+			return nil, NewDetailedError(ErrorBadAttestationStatementType, "Root certificate not signed by Android")
+		}
+	default:
+		return nil, NewDetailedError(ErrorBadAttestationStatementType, "Invalid root certificate signature algorithm")
```
hslatman, line 1508:
```suggestion
		return nil, NewDetailedError(ErrorBadAttestationStatementType, "invalid root certificate key type")
```

authority/provisioner/acme.go
```diff
@@ -216,10 +226,50 @@ func (p *ACME) Init(config Config) (err error) {
...
+	if err := json.NewDecoder(res.Body).Decode(&CRLResponse); err != nil {
+		return fmt.Errorf("error decoding Android CRL JSON: %w", err)
+	}
+
+	// Extract keys into a slice
+	keys := make([]string, 0, len(CRLResponse.Entries))
+	for k := range CRLResponse.Entries {
+		keys = append(keys, k)
+	}
+	p.RootCRLs = keys
```
hslatman, line 268: This field should be renamed. It contains the revoked keys themselves; not multiple CRLs. Maybe `p.RevokedAndroidAttestationKeySerials`.

authority/provisioner/acme.go
```diff
@@ -120,6 +128,8 @@ type ACME struct {
 	AttestationRoots    []byte   `json:"attestationRoots,omitempty"`
 	Claims              *Claims  `json:"claims,omitempty"`
 	Options             *Options `json:"options,omitempty"`
+	RootCRLs            []string `json:"rootCRLs,omitempty"`
```
hslatman, line 131: It would be preferable to not have to define `RootCRLs` on the provisioner. The existing properties are all meant as configuration, whereas the `RootCRLs` is currently used as state. 

If we don't expect users to have to set a different key revocation status URL, then `RootCRLs` (also see the other comment to rename it) can be unexported, and just be handled as state. Maybe all Android revocation state can be contained in a new struct as further improvement.

authority/provisioner/acme.go
```diff
@@ -216,10 +226,50 @@ func (p *ACME) Init(config Config) (err error) {
...
 	p.ctl, err = NewController(p, p.Claims, config, p.Options)
 	return
 }
 
+const ANDROID_ATTESTATION_STATUS_URL = "https://android.googleapis.com/attestation/status"
+
+// fetch CRL https://android.googleapis.com/attestation/status and build a list of serial number
+func (p *ACME) initializeAndroidCRL() error {
+	log.Printf("Updating Android CRL list for %s ACME provisioner", p.Name)
+	var CRLResponse struct {
```
hslatman, line 242:
```suggestion
	var crlResponse struct {
```

authority/provisioner/acme.go
```diff
@@ -216,10 +226,50 @@ func (p *ACME) Init(config Config) (err error) {
...
+		return fmt.Errorf("client: error making Android CRL request: %s\n", err)
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		bodyBytes, _ := io.ReadAll(res.Body)
+		return fmt.Errorf("unexpected Android CRL response %d: %s", res.StatusCode, string(bodyBytes))
+	}
+
+	if err := json.NewDecoder(res.Body).Decode(&CRLResponse); err != nil {
```
hslatman, line 259:
```suggestion
	if err := json.NewDecoder(res.Body).Decode(&crlResponse); err != nil {
```

authority/provisioner/acme.go
```diff
@@ -216,10 +226,50 @@ func (p *ACME) Init(config Config) (err error) {
...
+		return fmt.Errorf("error decoding Android CRL JSON: %w", err)
+	}
+
+	// Extract keys into a slice
+	keys := make([]string, 0, len(CRLResponse.Entries))
```
hslatman, line 264:
```suggestion
	keys := make([]string, 0, len(crlResponse.Entries))
```

authority/provisioner/acme.go
```diff
@@ -216,10 +226,50 @@ func (p *ACME) Init(config Config) (err error) {
...
+	for k := range CRLResponse.Entries {
```
hslatman, line 265:
```suggestion
	for k := range crlResponse.Entries {
```

authority/provisioner/acme.go
```diff
@@ -216,10 +226,50 @@ func (p *ACME) Init(config Config) (err error) {
...
+	}
+
 	p.ctl, err = NewController(p, p.Claims, config, p.Options)
 	return
 }
 
+const ANDROID_ATTESTATION_STATUS_URL = "https://android.googleapis.com/attestation/status"
+
+// fetch CRL https://android.googleapis.com/attestation/status and build a list of serial number
+func (p *ACME) initializeAndroidCRL() error {
```
hslatman, line 240:
```suggestion
// fetch CRL https://android.googleapis.com/attestation/status and build a list of revoked serial numbers
func (p *ACME) fetchAndroidCRL() error {
```

authority/provisioner/acme.go
```diff
@@ -392,3 +442,12 @@ func (p *ACME) IsAttestationFormatEnabled(_ context.Context, format ACMEAttestat
 func (p *ACME) GetAttestationRoots() (*x509.CertPool, bool) {
 	return p.attestationRootPool, p.attestationRootPool != nil
 }
+
+// IsRootRevoked return a true if the serialNumber is part of the list
+// It will also be in charge of updating the list periodically if no CRL list is provided at configuration.
+func (p *ACME) IsRootRevoked(serialNumber string) bool {
+	if slices.Contains(p.AttestationFormats, "android-key") && !p.androidCRLTimeout.IsZero() && time.Now().After(p.androidCRLTimeout) {
+		p.initializeAndroidCRL()
```
hslatman, line 450: I kinda like how this is done lazily, but I think it would be better if it were performed on a timer and taking into account caching data on the Google side. It might be sufficient initially, though.
```suggestion
		p.fetchAndroidCRL()
```

authority/provisioner/acme.go
```diff
@@ -392,3 +442,12 @@ func (p *ACME) IsAttestationFormatEnabled(_ context.Context, format ACMEAttestat
 func (p *ACME) GetAttestationRoots() (*x509.CertPool, bool) {
 	return p.attestationRootPool, p.attestationRootPool != nil
 }
+
+// IsRootRevoked return a true if the serialNumber is part of the list
+// It will also be in charge of updating the list periodically if no CRL list is provided at configuration.
+func (p *ACME) IsRootRevoked(serialNumber string) bool {
+	if slices.Contains(p.AttestationFormats, "android-key") && !p.androidCRLTimeout.IsZero() && time.Now().After(p.androidCRLTimeout) {
```
hslatman, line 449: Consider an additional early return for clarity

authority/provisioner/acme.go
```diff
@@ -392,3 +442,12 @@ func (p *ACME) IsAttestationFormatEnabled(_ context.Context, format ACMEAttestat
...
+		p.initializeAndroidCRL()
+	}
+	return len(p.RootCRLs) > 0 && slices.Contains(p.RootCRLs, serialNumber)
```
hslatman, line 452: It's sufficient to just check using `Contains`, as it's backed by `range`:
```suggestion
	return slices.Contains(p.RootCRLs, serialNumber)
```

authority/provisioner/acme.go
```diff
@@ -392,3 +442,12 @@ func (p *ACME) IsAttestationFormatEnabled(_ context.Context, format ACMEAttestat
 func (p *ACME) GetAttestationRoots() (*x509.CertPool, bool) {
 	return p.attestationRootPool, p.attestationRootPool != nil
 }
+
+// IsRootRevoked return a true if the serialNumber is part of the list
+// It will also be in charge of updating the list periodically if no CRL list is provided at configuration.
+func (p *ACME) IsRootRevoked(serialNumber string) bool {
```
hslatman, line 448: Also see comment w.r.t. fetching the serials. It would be better to extract that out of this function, to make it responsible for a single thing.
```suggestion
// IsRootRevoked returns true if the provided serialNumber is in the list of revoked 
// certificate serial number.
// It will also be in charge of updating the list periodically if no CRL list is provided at configuration.
func (p *ACME) IsRootRevoked(serialNumber string) bool {
```

authority/provisioner/acme.go
```diff
@@ -216,10 +226,50 @@ func (p *ACME) Init(config Config) (err error) {
...
+	var CRLResponse struct {
+		Entries map[string]struct {
+			Status string `json:"status"`
+			Reason string `json:"reason"`
+		} `json:"entries"`
+	}
+	res, err := http.Get(ANDROID_ATTESTATION_STATUS_URL)
+	if err != nil {
+		return fmt.Errorf("client: error making Android CRL request: %s\n", err)
+	}
```
hslatman, line 251: Can you change this into using the `http.Client` that's available on the `Controller`?

authority/provisioner/acme.go
```diff
@@ -216,10 +226,50 @@ func (p *ACME) Init(config Config) (err error) {
...
+// fetch CRL https://android.googleapis.com/attestation/status and build a list of serial number
+func (p *ACME) initializeAndroidCRL() error {
+	log.Printf("Updating Android CRL list for %s ACME provisioner", p.Name)
+	var CRLResponse struct {
+		Entries map[string]struct {
+			Status string `json:"status"`
+			Reason string `json:"reason"`
+		} `json:"entries"`
+	}
+	res, err := http.Get(ANDROID_ATTESTATION_STATUS_URL)
```
hslatman, line 248:
```suggestion
	res, err := http.Get(androidAttestationStatusURL)
```

authority/provisioner/acme.go
```diff
@@ -216,10 +226,50 @@ func (p *ACME) Init(config Config) (err error) {
 		return fmt.Errorf("failed initializing Wire options: %w", err)
 	}
 
+	if slices.Contains(p.AttestationFormats, "android-key") && len(p.RootCRLs) == 0 {
+		p.initializeAndroidCRL()
```
hslatman, line 230:
```suggestion
		p.fetchAndroidCRL()
```

hslatman (commented):
Hey @jbpin,

Looking quite good overall. Many suggestions are stylistic; feel free to combine them into your own commit(s). 

I'd like to test this myself with a device too to inspect the full X5C that's provided. Or, alternatively, it would be nice to get a chain from your testing. There's some surprising thing in the code and the Google docs, so I just want to make sure things are correct.

Functionally, and kind of expected, the biggest thing that needs some attention/changes is fetching and refreshing the revoked attestation certificate serials. Eventually that logic needs to become "more pluggable" because of the way we integrate changes into our hosted version. Concretely, it would not be necessary to refetch the revocation list for each hosted authority, as it would simply be retrieving the same list for each. I think it'll come down to going with a basic version for now, and then we may do some changes to make it compatible with our platform ourselves, but we'll have to see how things roll.

acme/challenge.go
```diff
@@ -1367,6 +1399,195 @@ func doAppleAttestationFormat(_ context.Context, prov Provisioner, _ *Challenge,
 	return data, nil
 }
 
+// Android Root CA
+// https://developer.android.com/privacy-and-security/security-key-attestation#root_certificate
+const AndroidRootCAPubKey = `-----BEGIN PUBLIC KEY-----
+MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEAr7bHgiuxpwHsK7Qui8xU
```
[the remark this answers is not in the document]
hslatman: Also see e.g. the `apple` format, which relies on the default hardcoded Apple attestation root CA, but allows overriding it by setting the existing `attestationRoots` configuration on the provisioner level.

authority/provisioner/acme.go
jbpin: I think this is based on the comment you made on discord: We may also want to be able to provide revoked keys in a way that doesn't rely on the Google endpoint for those same purposes.

I've certainly misunderstand your point so I made it as a configuration. Can be removed.

acme/challenge.go
```diff
@@ -1367,6 +1399,195 @@ func doAppleAttestationFormat(_ context.Context, prov Provisioner, _ *Challenge,
...
+
+// Attestion oid for Android, encoded as an integer.
+// https://source.android.com/docs/security/features/keystore/attestation#id-attestation
+var oidAndroidAttestation = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 1, 17}
+
+type androidKeyAttestationData struct {
+	Certificate *x509.Certificate
+	Fingerprint string
+	Attestation *attestation.KeyDescription
+}
```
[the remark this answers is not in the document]
jbpin: I imagine it is possible. The difference between Android Attestation and Apple Attestation, if I understand it, is that in Apple each data is a separate OID when Android is a structure in one OID. Parsing this value to the attestation KeyDescription object coming from the lib introduce in this implementation. We may still extract value necessary then discard the whole object.

jbpin:
Hi @hslatman,
I'm back from vacation, I will implement the suggestion and change a bit the code. It looks like there are some changes since this implementation on Google side : https://developer.android.com/privacy-and-security/security-key-attestation 

We will be able to implement the `attestationRoots` instead of key and support ECDSA as well.

acme/challenge.go
```diff
@@ -1367,6 +1399,195 @@ func doAppleAttestationFormat(_ context.Context, prov Provisioner, _ *Challenge,
...
+			root = cert
+		} else {
+			certs = append(certs, cert)
+			intermediates.AddCert(cert)
+		}
+	}
+
+	if root == nil {
+		return nil, NewDetailedError(ErrorBadAttestationStatementType, "missing root certificate in x5c chain")
+	}
```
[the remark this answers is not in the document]
jbpin: Yes the root is part of the chain. 

> check that the attestation certificate chain contains a root certificate that is signed with the Google attestation root key and that the attestationSecurityLevel element within the [key description](https://source.android.com/docs/security/features/keystore/attestation#keydescription-fields) data structure is set to the TrustedEnvironment security level or to the StrongBox security level.

acme/challenge.go
```diff
@@ -36,6 +37,7 @@ import (
 	"go.step.sm/crypto/x509util"
 	"golang.org/x/exp/slices"
 
+	"github.com/mbreban/attestation"
```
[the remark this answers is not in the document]
hslatman: We've forked it under `github.com/smallstep/android-attestation`.

acme/challenge.go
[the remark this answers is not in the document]
hslatman: Addressed in #2749.

acme/challenge.go
```diff
@@ -1367,6 +1399,195 @@ func doAppleAttestationFormat(_ context.Context, prov Provisioner, _ *Challenge,
...
+
+	keyAuth, err := KeyAuthorization(ch.Token, jwk)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse attestation data:
+	// find the attestation certificate
+	attCert, err := findAndroidAttestationCert(certs)
+	if err != nil {
```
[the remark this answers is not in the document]
hslatman: Addressed in #2749 by iterating from the end (root) of the chain to the start (leaf).

authority/provisioner/acme.go
hslatman: Addressed in #2749 by refactoring the CRL check.

authority/provisioner/acme.go
hslatman: Addressed in #2749 by refactoring the CRL configuration.

acme/challenge.go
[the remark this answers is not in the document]
hslatman: Covered in #2749.

hslatman:
Closing this, as its functionality has been merged via #2749.

PR #2325, opened 2025-07-01
Memory improvements

maraino:
This commit replaces the client in provisioners and webhooks with an interface. Then it implements the interface using the new poolhttp package. This package implements the HTTPClient interface but it is backed by a sync.Pool, this improves memory, allowing the GC to clean more memory. It also removes the timer in the keystore to avoid having extra goroutines if a provisioner goes away. This commit avoids creating the templates func multiple times, reducing some memory in the heap.

### DIFF
--- a/authority/authority_test.go
+++ b/authority/authority_test.go
@@ -28,8 +28,8 @@ import (
 
 func TestMain(m *testing.M) {
 	if err := initializeSystemCertPool(); err != nil {
-		fmt.Fprintln(os.Stderr, "failed to initialize system cert pool:", err)
-		fmt.Fprintln(os.Stderr, "See https://pkg.go.dev/github.com/tjfoc/gmsm/x509#SystemCertPool\n", err)
+		fmt.Fprintf(os.Stderr, "failed to initialize system cert pool: %v\n", err)
+		fmt.Fprintln(os.Stderr, "See https://pkg.go.dev/crypto/x509#SystemCertPool")
 		os.Exit(2)
 	}
 
--- a/authority/authority_test.go
+++ b/authority/authority_test.go
@@ -8,6 +8,7 @@ import (
 	"crypto/x509"
 	"encoding/hex"
 	"encoding/pem"
+	"fmt"
 	"net"
 	"os"
 	"path/filepath"
@@ -24,6 +25,16 @@ import (
 	"go.step.sm/crypto/minica"
 	"go.step.sm/crypto/pemutil"
 )
+
+func TestMain(m *testing.M) {
+	if err := initializeSystemCertPool(); err != nil {
+		fmt.Fprintln(os.Stderr, "failed to initialize system cert pool:", err)
+		fmt.Fprintln(os.Stderr, "See https://pkg.go.dev/github.com/tjfoc/gmsm/x509#SystemCertPool\n", err)
+		os.Exit(2)
+	}
+
+	os.Exit(m.Run())
+}
 
 func testAuthority(t *testing.T, opts ...Option) *Authority {
 	maxjwk, err := jose.ReadKey("testdata/secrets/max_pub.jwk")
--- a/authority/http_client.go
+++ b/authority/http_client.go
@@ -3,36 +3,54 @@ package authority
 import (
 	"crypto/tls"
 	"crypto/x509"
-	"fmt"
 	"net/http"
+	"sync/atomic"
 
+	"github.com/smallstep/certificates/authority/poolhttp"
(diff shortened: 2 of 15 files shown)


authority/http_client.go
```diff
@@ -3,36 +3,54 @@ package authority
...
+	if systemCertPool.Load() == nil {
+		pool, err := x509.SystemCertPool()
+		if err != nil {
+			return err
+		}
+		systemCertPool.Store(pool)
+	}
+	return nil
+}
+
```
hslatman, line 30: `x509.SystemCertPool` does use some logic to only execute this once already, and then provides a `Clone` if set, so I believe this may not be necessary? Or am I missing something?

authority/authority_test.go
```diff
@@ -25,6 +26,16 @@ import (
 	"go.step.sm/crypto/pemutil"
 )
 
+func TestMain(m *testing.M) {
+	if err := initializeSystemCertPool(); err != nil {
+		fmt.Fprintln(os.Stderr, "failed to initialize system cert pool:", err)
```
hslatman, line 31:
```suggestion
		fmt.Fprintf(os.Stderr, "failed to initialize system cert pool: %v\n", err)
```

authority/poolhttp/poolhttp.go
```diff
@@ -0,0 +1,80 @@
...
+	"sync"
+
+	"github.com/smallstep/certificates/internal/httptransport"
+)
+
+// Transporter is the implemented by custom HTTP clients with a method that
+// returns an [*http.Transport].
+type Transporter interface {
+	Transport() *http.Transport
+}
```
hslatman, line 14:
```suggestion
// Transporter is implemented by custom HTTP clients with a method that
// returns an [*http.Transport].
type Transporter interface {
	Transport() *http.Transport
}
```

authority/poolhttp/poolhttp.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Client returns an HTTP client that uses a [sync.Pool] to create new HTTP
+// client. It implements the [provisioner.HTTPClient] and [Transporter]
+// interfaces. This is the HTTP client used by the provisioners.
+type Client struct {
+	pool sync.Pool
+}
```
hslatman, line 21:
```suggestion
// Client is an HTTP client that uses a [sync.Pool] to create new and reuse HTTP
// clients. It implements the [provisioner.HTTPClient] and [Transporter]
// interfaces. This is the HTTP client used by the provisioners.
type Client struct {
	pool sync.Pool
}
```

authority/poolhttp/poolhttp.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+// SetNew replaces the inner pool with a new [sync.Pool] with the given New
+// function. This method should not be used concurrently with other methods.
+func (c *Client) SetNew(fn func() *http.Client) {
+	c.pool = sync.Pool{
+		New: func() any { return fn() },
+	}
+}
```
hslatman, line 39: What purpose does this function serve? Wouldn't a call to `New` be sufficient? If it shouldn't be used concurrently, it might need a mutex.

authority/poolhttp/poolhttp.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Get issues a GET to the specified URL. If the response is one of the
+// following redirect codes, Get follows the redirect after calling the
+// [Client.CheckRedirect] function:
+func (c *Client) Get(u string) (resp *http.Response, err error) {
```
hslatman, line 44: Also, it looks like the redirect codes are missing.
```suggestion
// Get issues a GET request to the specified URL. If the response is one of the
// following redirect codes, Get follows the redirect after calling the
// [Client.CheckRedirect] function:
func (c *Client) Get(u string) (resp *http.Response, err error) {
```

authority/poolhttp/poolhttp.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Transport() returns a clone of the http.Client Transport or returns the
+// default transport.
+func (c *Client) Transport() *http.Transport {
+	if hc, ok := c.pool.Get().(*http.Client); ok && hc != nil {
+		tr, ok := hc.Transport.(*http.Transport)
+		c.pool.Put(hc)
+		if ok {
+			return tr.Clone()
+		}
```
hslatman, line 76: In other places I've been relying on a check for the `Clone` method to be present, instead of checking the `*http.Transport` type. It's a bit more flexible, and could work here too.

hslatman (approved):

PR #2338, opened 2025-07-15
Fix process hanging after SIGTERM

maraino:
This commit fixes a problem where step-ca hangs after a shutdown because errors were sent to a channel without enough capacity.


smst-jeff:
@maraino could this be rewritten using [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup)?

PR #2369, opened 2025-08-19
Add support for Yubikey 5.7.4+

### LINKED ISSUE #2355: Update Yubico attestation CA
## Hello!

## Issue details

Yubico is updating its root CA: https://developers.yubico.com/PIV/Introduction/PIV_attestation.html#_yubico_root_ca_update. It looks like there's a new attestation CA root certificate, and multiple intermediate CAs chained off of that, including (at least) one that's meant for PIV attestation. 

We currently rely on [a single attestation root as our default](https://github.com/smallstep/certificates/blob/master/acme/challenge.go#L1370) when verifying the `step` attestation format. We should include at least the new attestation root in our verification logic. We may also need to include the PIV attestation intermediate(s) in the server side validation, or we should ensure that the CLI includes the right intermediates.

In go-piv, the new root and intermediates are already included: https://github.com/go-piv/piv-go/pull/176.

## Why is this needed?

Support `step` attestation format with YubiKeys with firmware 5.7.4+.

maraino:
### Description

This change adds the root and intermediate certificates used on PIV attestation on YubiKeys 5.7.4+

See https://developers.yubico.com/PKI/yubico-ca-certs.txt

Fixes #2355

Note that this hasn't been tested yet with a YubiKey 5.7.4+

@hslatman I think it might make more sense to use `go.step.sm/crypto` and `step kms attest` to get all the intermediates, instead of adding them here.


maraino:
Closed in favor of https://github.com/smallstep/certificates/pull/2370

PR #2379, opened 2025-08-28
Fix sshagentkms

glance-:
#### Name of feature:
Fix sshagentkms

sshagentkms not working after "Implementation of the Prometheus endpoint (#1669)" (dd1ff9c15b0f8ca5920dac6a0ff080b7f0cfb8be).

Currently, sshagentkms doesn't work in anything after v0.25.2

Yes, this feature already exists, and this just makes it work again.

When one already has a ssh-certificate signing infrastructure based on ssh-agent and would like to expose that via step, or when one would like to use a HSM which exposes the access to the key via a ssh-agent interface and would like to expose that via step.

Any old step versions between dd1ff9c15b0f8ca5920dac6a0ff080b7f0cfb8be and master.

None. This is both bug report and PR with fix.

💔Thank you!

### DIFF
--- a/authority/authority.go
+++ b/authority/authority.go
@@ -532,6 +532,13 @@ func (a *Authority) init() error {
 			switch s := signer.(type) {
 			case *sshagentkms.WrappedSSHSigner:
 				a.sshCAHostCertSignKey = s.Signer
+			case *instrumentedKMSSigner:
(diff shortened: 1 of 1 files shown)


hslatman:
Thank you, @glance-!  😄

PR #2382, opened 2025-09-04
Support managed device ID OID for `step` attestation format

hslatman:


acme/challenge.go
```diff
@@ -1524,16 +1529,27 @@ func doStepAttestationFormat(_ context.Context, prov Provisioner, ch *Challenge,
 		return nil, WrapErrorISE(err, "error calculating key fingerprint")
 	}
 	for _, ext := range leaf.Extensions {
-		if !ext.Id.Equal(oidYubicoSerialNumber) {
-			continue
+		if ext.Id.Equal(oidYubicoSerialNumber) {
```
smst-jeff, line 1532: Breaking this logic out into a func made it more quickly understood for me. Something similar to the following.

```suggestion
func doStepAttestationFormat(_ context.Context, prov Provisioner, ch *Challenge, jwk *jose.JSONWebKey, att *attestationObject) (*stepAttestationData, error) {
	...
	// Parse attestation data:
	// TODO(mariano): add support for other extensions.
	data := &stepAttestationData{
		Certificate: leaf,
	}
	if data.Fingerprint, err = keyutil.Fingerprint(leaf.PublicKey); err != nil {
		return nil, WrapErrorISE(err, "error calculating key fingerprint")
	}
	
	if data.SerialNumber, err = findSerialNumber(leaf); err != nil {
		return nil, WrapErrorISE(err, "error finding serial number")
	}

	return data, nil
}

func findSerialNumber(leaf *x509.Certificate) (string, error) {
	for _, ext := range leaf.Extensions {
		if ext.Id.Equal(oidYubicoSerialNumber) {
			var serialNumber int
			rest, err := asn1.Unmarshal(ext.Value, &serialNumber)
			if err != nil || len(rest) > 0 {
				return "", WrapError(ErrorBadAttestationStatementType, err, "error parsing serial number")
			}
			return strconv.Itoa(serialNumber), nil
		}
		if ext.Id.Equal(oidStepManagedDevice) {
			type stepManagedDevice struct {
				DeviceID string
			}
			var md stepManagedDevice
			rest, err := asn1.Unmarshal(ext.Value, &md)
			if err != nil || len(rest) > 0 {
				return "", WrapError(ErrorBadAttestationStatementType, err, "error parsing serial number")
			}
			return md.DeviceID, nil
		}
	}
	
	return "", nil
}
```

hslatman: Yep, better: eb475e0f7ca82e40659e93f6f0f33b1e2d92fa0a.

Issue #477, opened 2021-02-16
Notify systemd of finished startup, reloading, reload failures, ...


mohe2015:
### Subject of the issue
The service should tell systemd whether a reload, startup etc. was successful/finished. Either dbus (see example https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecReload= ) or sd_notify (https://www.freedesktop.org/software/systemd/man/sd_notify.html, e.g https://github.com/coreos/go-systemd) are likely the best options (although I never used any of them)

mohe2015:
As a minimal idea I would love to see startup notification as my acme services fail as they seem to startup faster than step-ca.

tashian:
@mohe2015 I think if this were fixed it would address the service dependency issue you're facing. We discussed this last week in our open source triage meeting. We're committed to it, as long as it's a quick fix and doesn't add too much complexity. Looks like the Go systemd module will make it pretty easy.

provokateurin:
I implemented this: https://github.com/smallstep/certificates/pull/2463

PR #2463, opened 2025-11-09
Add sd_notify support

provokateurin:
#### Name of feature:

sd_notify support

Closes https://github.com/smallstep/certificates/issues/477

Allows a systemd service to use `Type=notify`, so it's only considered ready once all the listeners are up.
Additionally I also added sd_notify Reloading support, so systemd knows when step-ca is reloading the config.
See https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#Options and https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html#Well-known%20assignments.

This allows systemd to be aware of when the service is actually started, so other services that depend on it can be started. Currently this is not the case and other services will fail to connect, if started too early.

Not sure if needed, but probably useful.

systemd

non-systemd

### DIFF
--- a/ca/ca.go
+++ b/ca/ca.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/coreos/go-systemd/v22/daemon"
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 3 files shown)


hslatman:
Thank you, @provokateurin! 🙂

provokateurin:
Thanks for merging! Any idea when a new release is planned? I just saw there are a lot of commits since the last release, so it wouldn't be just for this feature.

hslatman:
It'll likely before / close to end of the month.

provokateurin:
Cool, thanks!

Issue #2480, opened 2025-11-23
Bump docker images to Debian Trixie


udf2457:
Could you please consider bumping the docker images to Debian Trixie ?

Added bonus is you could address #2469 too.

maraino:
Note that https://github.com/smallstep/certificates/pull/2493 will fix the issue, but a new release would be required to publish the image.

PR #2493, opened 2025-12-04
Update Debian version to trixie

### LINKED ISSUE #2469: Add yubihsm-pkcs11 in step-ca:hsm docker image
## Hello!

## Issue details

First of all, thank for your great work.

I would like to know if it is possible to natively add yubihsm-pkcs11 package in step-ca:hsm docker image. Indeed since Debian Trixie (13) it is available in main repository. That implied to upgrade image from Debian 12 to Debian 13.

## Why is this needed?

It will be more easy to use step-ca:hsm if it ship all dependencies out of the box for YubiHSM.

Of course I understand that your image can not include all dependencies for all HSM over the world, but YubiHSM is a very common product and integration/maintenance will be very easy since it is natively available on Debian.

Best regards.

maraino:
This commit updates the Debian version in the docker images from bookworm to trixie, the latest stable version.

Requires:
- [x] https://github.com/smallstep/cli/pull/1530
- [x] https://github.com/smallstep/step-kms-plugin/pull/287

Fixes #2469, #2480

~@hslatman @tashian: I wasn't sure how to avoid using the shell, without it the environment variables are not expanded. And if I use the old version, that will use a shell, the docker language server complains with:~

~JSON arguments recommended for ENTRYPOINT/CMD to prevent unintended behavior related to OS signals (JSON arguments recommended for CMD to prevent unintended behavior related to OS signals)Docker DX (docker-language-server) [JSONArgsRecommended](https://docs.docker.com/go/dockerfile/rule/json-args-recommended/)~

I've decided to hardcode the password and configuration path, so you can have step-ca running with the PID 1 with a command like this:
```
docker run --rm \
  -e DOCKER_STEPCA_INIT_NAME=localhsot \
  -e DOCKER_STEPCA_INIT_DNS_NAMES=localhost \
  smallstep/step-ca:hsm
```

And k8s or docker allows you to change the CMD easily.

### DIFF
--- a/acme/challenge.go
+++ b/acme/challenge.go
@@ -231,7 +231,7 @@ func tlsAlert(err error) uint8 {
 	if errors.As(err, &opErr) {
 		v := reflect.ValueOf(opErr.Err)
 		if v.Kind() == reflect.Uint8 {
-			return uint8(v.Uint()) //nolint:gosec // handled by checking its type
+			return uint8(v.Uint())
 		}
 	}
 	return 0
--- a/acme/linker.go
+++ b/acme/linker.go
@@ -84,13 +84,13 @@ func GetUnescapedPathSuffix(typ LinkType, provisionerName string, inputs ...stri
 	case NewNonceLinkType, NewAccountLinkType, NewOrderLinkType, NewAuthzLinkType, DirectoryLinkType, KeyChangeLinkType, RevokeCertLinkType:
 		return fmt.Sprintf("/%s/%s", provisionerName, typ)
 	case AccountLinkType, OrderLinkType, AuthzLinkType, CertificateLinkType:
-		return fmt.Sprintf("/%s/%s/%s", provisionerName, typ, inputs[0]) //nolint:gosec // operating on internally defined inputs
+		return fmt.Sprintf("/%s/%s/%s", provisionerName, typ, inputs[0])
 	case ChallengeLinkType:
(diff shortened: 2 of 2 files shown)


hslatman:
@maraino also see https://github.com/smallstep/certificates/issues/2480 and https://github.com/smallstep/certificates/issues/2469.

JoKneeMo:
This hardcoding of the paths was a major change breaking change that should have been called out in the release notes!

The 4 word description of the commit in the a pre-release was not enough.
Furthermore, the release notes have a section for "Changelog", but it's populated with the commits and not the text from CHANGELOG.md.

hslatman:
> This hardcoding of the paths was a major change breaking change that should have been called out in the release notes!

Agree. That's what I mentioned above. 

> The 4 word description of the commit in the a pre-release was not enough. Furthermore, the release notes have a section for "Changelog", but it's populated with the commits and not the text from CHANGELOG.md.

The change affecting you _is_ in the official changelog: https://github.com/smallstep/certificates/blob/master/CHANGELOG.md#changed-1. It doesn't mention explicitly that the paths are now handled differently, but it's not missing fully, and it does mention the changes to the `CMD`.

The one you're referring to on GitHub releases is indeed just the commits. Recently there have been quite a few issues with GitHub Actions, which is what ultimately creates that release for us, so it's possible that plays a role here too.

PR #2511, opened 2026-01-05
feat: unset OnlyContainsUserCerts and OnlyContainsCACerts options during Distribution Point marshaling

savely-krasovsky:
#### Name of feature:

Unset OnlyContainsUserCerts and OnlyContainsCACerts options during Distribution Point marshaling.

💔Thank you!


maraino (commented):
I haven't tried, but my I think that if you set `idpOnlyContainsCACerts` and you revoke a "user" certificate, that will also appear in the CRL.

savely-krasovsky:
If there is no special reason to set `OnlyUser` I think it would make sense to just disable it. Specification says you must set either distribution point or one of those flags on, but in our case DP is already set, so both flags could be unset.

savely-krasovsky:
@maraino what do you think about the change like that? I don't see any particular reason why those params are set. `step-ca` CRL list could contain both CA and user certs.

maraino:
> @maraino what do you think about the change like that? I don't see any particular reason why those params are set. step-ca CRL list could contain both CA and user certs.

That makes more sense to me, we'll include this in our next triage.

filimonic:
CA/Browser forum, Baseline Requirements 2.2.2, 
- [7.2 CRL Profile](https://github.com/cabforum/servercert/blob/main/docs/BR.md#72-crl-profile)
- [7.2.2 CRL and CRL entry extensions](https://github.com/cabforum/servercert/blob/main/docs/BR.md#722-crl-and-crl-entry-extensions)
  - `IssuingDistributionPoint` presense marked as case-dependent, `See 7.2.2.1 CRL Issuing Distribution Point`
- [7.2.2.1 CRL Issuing Distribution Point](https://github.com/cabforum/servercert/blob/main/docs/BR.md#7221-crl-issuing-distribution-point)
  - ⚠️ This extension is NOT RECOMMENDED for full and complete CRLs.

Please see Google's newest [`AE1` subordinate certificate](https://i.pki.goog/ae1.crt) and it's [CRL](https://c.pki.goog/r/r4.crl)

savely-krasovsky:
Can I ask for the current status?

savely-krasovsky:
@maraino any news?

PR #2514, opened 2026-01-06
replace logrus with log/slog

mscso:
#### Name of feature:

Replace unmaintained sirupsen/logrus with stdlib log/slog.

sirupsen/logrus unmaintained.

removes unmaintained dependency risk - also from downstream projects.

💔Thank you!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,6 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.23.2
 	github.com/rs/xid v1.6.0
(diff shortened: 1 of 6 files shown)


hslatman:
Hi @mscso, thank you for opening the PR. While it's indeed a good step to start using `log/slog`, we'll need to adapt more (internal) projects to be able to use it in the CA code base. We have more changes related to this coming up, so we may need to wait for those to land first.

Note that `logrus` isn't fully unmaintained. A new release went out last week: https://github.com/sirupsen/logrus/releases/tag/v1.9.4.

Issue #2509, opened 2025-12-31
[Bug]: cloudCAS GetCertificateAuthority doesn't return IntermediateCertificates, breaking SCEP in RA mode


civicactionsbot:
### Steps to Reproduce

1. Set up GCP Certificate Authority Service with a subordinate (intermediate) CA
2. Configure step-ca in RA mode with cloudCAS:
```json
{
  "authority": {
    "type": "cloudcas",
    "certificateAuthority": "projects/my-project/locations/us-central1/caPools/my-pool/certificateAuthorities/my-intermediate-ca",
    "config": {
      "credentialsFile": "/path/to/gcp-sa-key.json"
    },
    "provisioners": [
      {
        "type": "SCEP",
        "name": "scep",
        "challenge": "secret",
        "minimumPublicKeyLength": 2048,
        "encryptionAlgorithmIdentifier": 2,
        "decrypterCertificate": "-----BEGIN CERTIFICATE-----\n...",
        "decrypterKeyPEM": "-----BEGIN RSA PRIVATE KEY-----\n..."
      }
    ]
  }
}
```
3. Start step-ca

### Your Environment

* OS - macOS (also reproducible on Linux via Helm chart)
* `step-ca` Version - 0.29.0

### Expected Behavior

step-ca should start successfully with SCEP provisioner in cloudCAS RA mode, similar to how VaultCAS works after the fix in PR #1803.

### Actual Behavior

step-ca panics on startup:

```
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/smallstep/certificates/authority.(*Authority).init(0x14000732280)
        /path/to/certificates/authority/authority.go:760 +0x...
```

The panic occurs because `a.intermediateX509Certs` is empty when initializing the SCEP authority.

In `cas/cloudcas/cloudcas.go`, the `GetCertificateAuthority()` function only returns `RootCertificate` but not `IntermediateCertificates`:

```go
func (c *CloudCAS) GetCertificateAuthority(req *apiv1.GetCertificateAuthorityRequest) (*apiv1.GetCertificateAuthorityResponse, error) {
    // ...
    // Last certificate in the chain is the root.
    root, err := parseCertificate(resp.PemCaCertificates[len(resp.PemCaCertificates)-1])
    if err != nil {
        return nil, err
    }

    return &apiv1.GetCertificateAuthorityResponse{
        RootCertificate: root,
        // IntermediateCertificates is NOT populated!
    }, nil
}
```

[...]

PR #2517, opened 2026-01-06
Fix CA startup when configured with SCEP and Google Cloud CAS

hslatman:
Fixes: #2509


hslatman:
@maraino that indeed works. The logic is in line with what we did for Vault now. Users of the endpoint will now possibly get more certificates in the response when using GCP CAS, but it should be in line with what one expects from that endpoint. 

I could follow up with an integration test in a new PR for this behavior? The current tests for the API endpoints are using a minimal mock authority, so it requires some more boilerplate to get an actual authority with a (fake) CAS setup.

PR #2519, opened 2026-01-07
Standalone workflow for publishing to packages.smallstep.com

tashian:
I thought it could be nice to have independent package publishing decoupled from the release process, especially since those steps happen later in the process. That's what this action does. This would let us publish v0.29.0. I hosted the workflow file inside this repo (and not the smallstep/workflows repository) because the bash scripts for package publishing are repo-specific.


hslatman:
@tashian the linter issues were resolved in https://github.com/smallstep/certificates/pull/2523.

PR #2520, opened 2026-01-07
Add changelog entry for smallstep/certificates#2493

maraino:
This commit adds an entry in the changelog related to the changes in smallstep/certificates#2493.


hslatman:
Text looks good. It looks like what is currently unreleased should actually be the most recently release, and what you added in this here should be the actual unreleased, though.

Linter issue is interesting. It looks very familiar to me; might've fixed it already at some point. Maybe the linter was disabled for a while and now enabled again.

PR #2523, opened 2026-01-07
Fix linter errors

maraino:


hslatman:
I've definitely seen these linter issues before. You removed them in https://github.com/smallstep/certificates/commit/5f06a51ebbfdedff08393713173c5ec87db205d1. They were added about a month before that. Some caching going on during the linting?

PR #2533, opened 2026-01-13
Add HTTP transport decorator

maraino:
This commit adds a new option to the CA client that allows the customization of the HTTP transport.

### DIFF
--- a/ca/client.go
+++ b/ca/client.go
@@ -278,7 +278,7 @@ func (o *clientOptions) getTransport(endpoint string) (tr http.RoundTripper, err
 	}
 
 	// Wrap the transport using the decorator function if necessary
-	return decorateTransport(tr, o.transportDecorator), nil
+	return decorateRoundTripper(tr, o.transportDecorator), nil
 }
 
 // WithTransport adds a custom transport to the Client. It will fail if a
@@ -1602,7 +1602,7 @@ func clientError(err error) error {
 	return fmt.Errorf("client request failed: %w", err)
 }
 
-func decorateTransport(tr http.RoundTripper, td TransportDecorator) http.RoundTripper {
+func decorateRoundTripper(tr http.RoundTripper, td TransportDecorator) http.RoundTripper {
 	if td != nil {
 		return td(tr)
 	}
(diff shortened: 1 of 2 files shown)


ca/client.go
```diff
@@ -1583,3 +1601,10 @@ func clientError(err error) error {
 	}
 	return fmt.Errorf("client request failed: %w", err)
 }
+
+func getTransportWithDecorator(tr http.RoundTripper, td TransportDecorator) http.RoundTripper {
+	if td != nil {
```
hslatman: It returns an `http.RoundTripper`, so consider reflecting that in the name. It could become something like `wrapWithDecorator`.

ca/client.go
```diff
@@ -1583,3 +1601,10 @@ func clientError(err error) error {
 	}
 	return fmt.Errorf("client request failed: %w", err)
 }
+
+func getTransportWithDecorator(tr http.RoundTripper, td TransportDecorator) http.RoundTripper {
```
azazeal: Maybe `decorateTransport`?

ca/client.go
```diff
@@ -287,6 +293,16 @@ func WithTransport(tr http.RoundTripper) ClientOption {
 	}
 }
 
+// WithTransportDecorator allows customization of the HTTP transport used by the client.
+// The provided function receives the configured http.RoundTripper and can wrap it
+// with additional functionality.
```
hslatman:
```suggestion
// WithTransportDecorator allows customization of the HTTP transport used by the client.
// The provided function receives the configured [http.RoundTripper] and can wrap it
// with additional functionality.
```

ca/client.go
maraino: Fixed with da56c707 using @azazeal suggestion

ca/client.go
```diff
@@ -562,11 +578,12 @@ func WithProvisionerName(name string) ProvisionerOption {
...
-	endpoint  *url.URL
-	retryFunc RetryFunc
-	timeout   time.Duration
-	opts      []ClientOption
+	client             *uaClient
+	endpoint           *url.URL
+	retryFunc          RetryFunc
+	timeout            time.Duration
+	opts               []ClientOption
+	transportDecorator TransportDecorator
```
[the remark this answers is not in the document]
maraino: We use those options later, perhaps we could refactor that part, but there might be some unexpected implications.

azazeal (approved):

Issue #2535, opened 2026-01-14
[Bug]: by default, CRL file is served with `.DER` extension, expected `.CRL`


filimonic:
### Steps to Reproduce

* Enable CRL
* Download CRL from `/1.0/crl` with something that respects `Content-Disposition: filename`

### Your Environment

* OS - docker image smallstep/step-ca:0.29.0
* `step-ca` Version - 0.29.0

### Expected Behavior

When downloading crl file from step-ca, server should offer client to use `.crl` file extension by default. 

### Actual Behavior

When downloading crl file from step-ca, server offers client to use `crl.der` file name by default (so, `.der` extension). 

### Additional Context

**IANA:**

CRL file in DER format with type `application/pkix-crl` should have `.CRL` extension, as [registered by IANA](https://www.iana.org/assignments/media-types/application/pkix-crl)

**Windows:**

In Windows systems, files with this extension `.der` are associated with `application/x-x509-ca-cert` type. This causes Windows to try opening this file as certificate and showing error `This file is invalid for use as the following: Security Certificate.`

The only extension associated with `application/pkix-crl` is .crl

This line: should be fixed https://github.com/smallstep/certificates/blob/f088be2548472b15571d8813c9579700ac01c5c4/api/crl.go#L43

### Contributing

Vote on this issue by adding a 👍 reaction.
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #2537, opened 2026-01-14
Change file extension suggested for DER format CRL downloads

filimonic:
#### Name of feature:

Change CRL file extension to `.crl` by default for DER format (default)

- Issue #2535 : when opening `.der` files through Windows, Windows default certificate viewer app expects file to be a certificate file, not CRL. This causes error `This file is invalid for use as the following: Security Certificate.`. User should disable _(default)_ hiding of registered extensions, change file extension to .crl (if user knows about it) and open again.

- Bringing to standard.

- CRL file in DER format with type application/pkix-crl should have .CRL extension, as [registered by IANA](https://www.iana.org/assignments/media-types/application/pkix-crl)

- Windows Desktop clients (default and only windows file association for `application/pkix-crl` is `.CRL` ). Any http client that respects `Content-Disposition` response header.

- Any http client that overrides or does not respect `Content-Disposition` response header.

* [IANA `application/pkix-crl`](https://www.iana.org/assignments/media-types/application/pkix-crl)

💔Thank you!

### DIFF
--- a/api/crl.go
+++ b/api/crl.go
@@ -40,7 +40,7 @@ func CRL(w http.ResponseWriter, r *http.Request) {
 		})
 	} else {
 		w.Header().Add("Content-Type", "application/pkix-crl")
-		w.Header().Add("Content-Disposition", "attachment; filename=\"crl.der\"")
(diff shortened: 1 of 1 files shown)


hslatman:
Thank you, @filimonic!

PR #2545, opened 2026-01-22
Migrate to `x509.RevocationListEntry` and add reason codes to CRL

gwe32:
This PR migrates from the deprecated `RevokedCertificates` field to `RevokedCertificateEntries`, which allows including revocation reason codes in the CRL without having to add them via extensions.


dopey:
https://github.com/smallstep/certificates/pull/2601

Issue #2607, opened 2026-03-19
[Bug]: Step CA 0.30.1-hsm fails to start: "connecting to pcsc: access was denied because of a security violation"


jorti:
### Steps to Reproduce

I run Step CA in a container and using a Yubikey.

After upgrading the container from `0.29.0-hsm` to `0.30.1-hsm`, the container fails to start with this error:

```
badger 2026/03/19 07:27:03 INFO: All 1 tables opened in 2ms
badger 2026/03/19 07:27:03 INFO: Replaying file id: 0 at offset: 59458095
badger 2026/03/19 07:27:03 INFO: Replay took: 1.160165ms
connecting to pcsc: access was denied because of a security violation
```

Debugging pcscd, I can see it's failing because of polkit:

```
11953524 [140696989920704] ../src/winscard_msg_srv.c:253:ProcessEventsServer() Common channel packet arrival
00000017 [140696989920704] ../src/winscard_msg_srv.c:264:ProcessEventsServer() ProcessCommonChannelRequest detects: 12
00000002 [140696989920704] ../src/pcscdaemon.c:130:SVCServiceRunLoop() A new context thread creation is requested: 12
00000157 [140696887490240] ../src/auth.c:115:IsClientAuthorized() polkit_authority_get_sync failed: Error initializing authority: Could not connect: No such file or directory
00000007 [140696887490240] ../src/winscard_svc.c:357:ContextThread() Rejected unauthorized PC/SC client
00000008 [140696887490240] ../src/winscard_svc.c:1112:MSGCleanupClient() Thread is stopping: dwClientID=12, threadContext @0x55f9792ec870
00000002 [140696887490240] ../src/winscard_svc.c:1120:MSGCleanupClient() Freeing SCONTEXT @0x55f9792ec870
```

There's no polkit inside a container. Starting pcscd with the option `--disable-polkit` fixes the issue.

### Your Environment

* OS: Fedora CoreOS 43
* `step-ca` Version:
```
Smallstep CA/0.30.1 (linux/amd64)
Release Date: 2026-03-19 01:49 UTC
```

### Expected Behavior

- Container works as it did before.
- pcscd is started with `--disable-polkit` in the container.

### Actual Behavior

Container crashes.

### Contributing

Vote on this issue by adding a 👍 reaction.
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

gtaws:
[...]

PR #2609, opened 2026-03-19
Disable polkit in pcscd

jorti:
There's no use of polkit inside a container and it makes step-ca connections to pcscd fail, so just disable the use of polkit in pcscd.

fixes: smallstep/certificates#2607

#### Name of feature:
Disable use of polkit in pcscd when running in a container.

#2607 

Configurations that were working before are now crashing

Only in containers

Not containers

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -97,7 +97,7 @@ function step_ca_init () {
 }
 
 if [ -f /usr/sbin/pcscd ]; then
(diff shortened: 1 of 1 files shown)


gtaws:
Any progress on this? I can't use the hsm image without it and had to revert to 0.29.0

jorti:
My contribution was greeted with a CLA. I pass to sign anything, so I'll drop this PR.

PR #2618, opened 2026-03-23
Add provisionerName to webhook request for x509 and SSH

Yannik:
#### Name of feature:
Add provisionerName to webhook request for x509 and SSH

The `provisionerName` field in `RequestBody` ([`webhook/types.go:86`](https://github.com/smallstep/certificates/blob/6e8ec614/webhook/types.go#L86)) is only populated for SCEP webhooks ([`authority/provisioner/scep.go:160`](https://github.com/smallstep/certificates/blob/6e8ec614/authority/provisioner/scep.go#L160)). For X509 and SSH webhooks, the field is always empty despite the provisioner being available as a parameter in all four call sites (`callEnrichingWebhooksX509`, `callAuthorizingWebhooksX509`, `callEnrichingWebhooksSSH`, `callAuthorizingWebhooksSSH`). This makes it impossible for webhook servers to implement per-provisioner logic (e.g., enforcing different key policies for different ACME provisioners).

The field already exists in the struct and is documented as part of the webhook request body. Not populating it for X509/SSH breaks the reasonable expectation that all webhook types provide the same base context to webhook servers.

The webhook request body is documented at https://smallstep.com/docs/step-ca/webhooks/ but does not currently mention `provisionerName`. The field exists in the `RequestBody` struct ([`webhook/types.go:86`](https://github.com/smallstep/certificates/blob/6e8ec614/webhook/types.go#L86)) and is already populated for SCEP webhooks, but is undocumented.

All provisioner types that trigger X509 or SSH enriching/authorizing webhooks (ACME, JWK, OIDC, x5c, etc.). SCEP already works and is unchanged.


maraino:
Thanks @Yannik, in PR #2631 I've changed the way to add the provisioner name is added using a new method RequestBodyOption, `webhook.WithProvisionerName(prov)`. And merge that version, it also includes your commits.

PR #2661, opened 2026-04-21
Add additional Wire challenge test cases

SimonThormeyer:
Hello! 👋 
This is intended to add additional test cases for the wire challenges, mainly by just copy-and-pasting existing test cases and tweaking them in the right places. It doesn't touch any code outside of the specific wire tests.

In our downstream usage, we were previously maintaining similar test cases, however, it greatly reduces complexity for us to directly test these cases here. 

Please also refer to the commit sequence to see what is being tested.

Would you be open to review/merge this?


istankovic:
@hslatman could you please take a look at this?

hslatman (approved):

hslatman:
Thank you, @SimonThormeyer!

PR #2662, opened 2026-04-22
acme/api: reject General JSON Serialization JWS bodies per RFC 8555

### LINKED ISSUE #2642: [Bug]: Step-CA accepts non-flattened JWS POST requests
### Steps to Reproduce

From [RFC 8555, Section 6.2](https://datatracker.ietf.org/doc/html/rfc8555#section-6.2):

> A JWS object sent as the body of an ACME request MUST meet the following additional criteria:
>
>   o  The JWS MUST be in the Flattened JSON Serialization [[RFC7515](https://datatracker.ietf.org/doc/html/rfc7515)]

This is not correctly implemented in `step-ca`, which accepts the ["General"](https://datatracker.ietf.org/doc/html/rfc7515#section-7.2.1) JWS format.
This is hard to show with an example, so I give a snippet from acme-tiny that I modified, and did not receive an error (though I should):

```
# Before (Flattened JWS format)
        data = json.dumps({"protected": protected64, "payload": payload64, "signature": _b64(out)})

# After (General JWS format)
        data = json.dumps({"payload": payload64, "signatures": [{"protected": protected64, "signature": _b64(out) }]})
```

Currently, the code uses `jose.ParseJWS` ([link](https://github.com/smallstep/certificates/blob/09be1c12a6426a0987a8bfa3bb6162b5e89fa70b/acme/api/middleware.go#L104)), which eventually calls the go-jose/v3's `ParseSigned` function ([link](https://github.com/smallstep/crypto/blob/76112d40afe9a10e1ce5dc29a475c8368ac002c2/jose/types.go#L285)), which accepts the full serialization format.

I don't believe go-jose provides any way to check this: instead there should be a check on the raw JSON that the `signatures` key does not exist prior to parsing.

### Your Environment

* OS - Linux
* `step-ca` Version - 0.29.0

### Expected Behavior

Step-ca rejects POST requests containing payloads in the General JWS format

### Actual Behavior

Step-ca accepts POST requests containing payloads in the General JWS format

### Additional Context

_No response_

### Contributing

SAY-5:
## Summary

Fixes #2642.

RFC 8555 section 6.2 mandates that ACME JWS bodies MUST use the Flattened JSON Serialization. go-jose/v3's `ParseSigned` accepts both Flattened and General forms, so `step-ca` silently accepted:

```json
{"payload": "...", "signatures": [{"protected": "...", "signature": "..."}]}
```

as if it were a legal RFC 7515 flattened body. As the reporter notes, go-jose doesn't expose a way to restrict this, so the check has to sit at the JSON layer before `jose.ParseJWS` runs.

## Fix

Add a lightweight `rejectGeneralJWS` helper in front of `parseJWS` that JSON-probes the body for a top-level `"signatures"` array (plural, the telltale General-form field). Hit → return `ACME ErrorMalformedType` with a message pointing back to RFC 8555. Bodies that aren't valid JSON fall through to the existing `ParseJWS` path so go-jose's own diagnostics still produce the familiar `compact JWS format must have three parts` error for compact-serialization mistakes.

## Tests

- New sub-test `TestHandler_parseJWS/fail/general-jws` drives a General-form body through `parseJWS` and asserts `400 ErrorMalformedType` with the RFC-8555-pointing message.
- Existing `fail/read-body-error`, `fail/parse-jws-error`, and `ok` cases stay green.

`go build ./acme/...`, `go vet ./acme/...`, `go test ./acme/api -run TestHandler_parseJWS` all green.

### DIFF
--- a/acme/api/middleware.go
+++ b/acme/api/middleware.go
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 	"crypto/rsa"
+	"encoding/json"
 	"errors"
 	"io"
 	"net/http"
@@ -101,6 +102,10 @@ func parseJWS(next nextHTTP) nextHTTP {
 			render.Error(w, r, acme.WrapErrorISE(err, "failed to read request body"))
 			return
 		}
+		if err := rejectGeneralJWS(body); err != nil {
+			render.Error(w, r, acme.WrapError(acme.ErrorMalformedType, err, "failed to parse JWS from request body"))
+			return
+		}
 		jws, err := jose.ParseJWS(string(body))
 		if err != nil {
 			render.Error(w, r, acme.WrapError(acme.ErrorMalformedType, err, "failed to parse JWS from request body"))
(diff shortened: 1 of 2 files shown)


acme/api/middleware.go
```diff
@@ -111,6 +116,28 @@ func parseJWS(next nextHTTP) nextHTTP {
...
 
+// rejectGeneralJWS enforces RFC 8555 section 6.2: ACME JWS requests MUST use
+// the Flattened JSON Serialization. go-jose/v3's ParseSigned accepts both the
+// Flattened and the General serialization, so the flattened-only requirement
+// has to be checked at the JSON layer before handing off. A General-form body
+// carries a top-level "signatures" array (plural); flattened bodies have
+// singular "signature" + "protected" alongside "payload".
+func rejectGeneralJWS(body []byte) error {
+	var peek struct {
+		Signatures json.RawMessage `json:"signatures"`
```
Skyb0rg007, line 129: You should also assert that the message doesn't include the unprotected header

```suggestion
		Signatures json.RawMessage `json:"signatures"`
		Header json.RawMessage `json:"header"`
```

PR #2690, opened 2026-05-21
Ensure empty times are not serialized on orders

cipherboy:
#### Name of bug:

The `omitempty` annotation on Go structs will not elide zero-valued time values if they're not pointers. Setting `omitzero` as well will ensure we do not send these values if they're not.

#### Pain or issue this bug alleviates:

Per RFC 8555, these fields are optional and are requested values by the user; it makes sense that server implementers would reject less-than-useful time values if present. Additionally, RFC 3339 defines date values as being between 0000 and 9999, making it ambiguous outside of the Go ecosystem if 0001-01-01T00:00:00Z would be the zero date. Notably, RFC 3339 Appendix B remarks:

> which may be used to obtain the day of the week for dates on or
> after 0000-03-01

which suggests dates in the year 0000 would work and a reasonable implementation could conclude 0001 is not the zero year.

Greater compatibility of SmallStep's client with existing ACME servers. 

See also: https://datatracker.ietf.org/doc/html/rfc8555#section-7.1.3
See also: https://datatracker.ietf.org/doc/html/rfc3339#appendix-B
See also: https://github.com/openbao/openbao/issues/3147

💔Thank you!


hslatman (approved):

hslatman:
Thank you, @cipherboy! 😄

PR #2715, opened 2026-06-18
Fix name policy panic on empty or short domain

### LINKED ISSUE #2714: [Bug]: name policy panics on a SAN with an empty domain (e.g. an email like "a@")
### Steps to Reproduce

With a name policy that has an email constraint, validate a SAN whose email has an empty domain. A unit test in the `policy` package:

```
func TestEmptyDomainPanic(t *testing.T) {
    e, _ := New(WithPermittedEmailAddresses("@example.com"))
    _ = e.AreSANsAllowed([]string{"a@"})
}
```

This path runs during signing via `IsX509CertificateAllowed` when a CSR/cert carries such an email SAN.

### Your Environment

- certificates master @ `3e9e8bee0cad4a9726f67653ead87c028463163a`
- Go 1.26

### Expected Behavior

The name policy rejects the SAN like any other non-matching name.

### Actual Behavior

An email SAN like `a@` yields an empty domain. net/http recovers the handler panic, so the request 500s rather than crashing the server.

### Additional Context

_No response_

### Contributing

kanywst:
Closes #2714

A SAN with an empty domain (e.g. an email like `a@`) panicked during name-policy validation with a slice index out of range. Guard the empty and single-character domain cases in `matchDomainConstraint`. The existing `matchDomainConstraint` table test gets the two cases, which panic before this change and pass after.

### DIFF
--- a/policy/engine_test.go
+++ b/policy/engine_test.go
@@ -55,6 +55,20 @@ func TestNamePolicyEngine_matchDomainConstraint(t *testing.T) {
 			wantErr:    false,
 		},
 		{
+			name:       "fail/empty-domain",
+			domain:     "",
+			constraint: "host.example.com",
+			want:       false,
+			wantErr:    false,
+		},
+		{
+			name:       "fail/single-asterisk-domain",
+			domain:     "*",
+			constraint: "host.example.com",
+			want:       false,
+			wantErr:    false,
+		},
+		{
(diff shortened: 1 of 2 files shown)


hslatman (approved):

hslatman:
Thank you, @kanywst! 😄

PR #2749, opened 2026-07-22
Add `android-key` attestation format support

hslatman:
This PR builds on #2322.


areed (commented):
Looks good. Should we skip the CRL check when not using the Android roots?

authority/provisioner/provisioner.go
```diff
@@ -281,6 +282,10 @@ type Config struct {
 	// WrapTransport references the function that should wrap any [http.Transport] initialized
 	// down the Config's chain.
 	WrapTransport TransportWrapper
+	// AndroidKeyCRLChecker references an implementation of [androidkey.CRLChecker]
+	// that is responsible for checking revoked Android Key Attestation certificate
+	// serial numbers.
+	AndroidKeyCRLChecker androidkey.CRLChecker
```
areed, line 288: It looks like this is always going to be nil

hslatman: Yes, that's the case for open source. In hosted we'll inject a custom checker so that we can keep a single shared CRL for just the Google roots. 

EDIT: missed a level of passing it around: c4ca1a6a4df86f3a49a74bd459120e85feeb5bf4.

areed (approved):

jbpin:
❤️

PR #2795, opened 2026-09-11
Added AWS mx-central-1 RSA key

Alshadex:
Originally added new AWS mx-central-1 region RSA Identity Document key. Decided to also add remaining latest new regions.

- mx-central-1 (Mexico)
- ap-southeast-5 (Malaysia)
- ap-southeast-6 (New Zealand)
- ap-southeast-7 (Thailand)
- ap-east-2 (Taipei)
- eusc-de-east-1 (AWS European Sovereign Cloud)


sanchit3110:
@Alshadex - there might be more regions that are missing https://docs.aws.amazon.com/global-infrastructure/latest/regions/doc-history.html 
Basically from the year 2025 and 2024

Issue #2800, opened 2026-09-17
[Bug]: SCEP response signing reuses the configured decrypter key, causing signature verification failures with FIPS-approved-only clients


tarokon:
### Steps to Reproduce

1. Configure a SCEP provisioner with a custom decrypter certificate/key so that decryption uses a dedicated RSA key pair instead of the CA's default signer key:
```
{
  "type": "SCEP",
  "name": "est-core-scep",
  "decrypterCertificate": "<base64 PEM>",
  "decrypterKeyPEM": "<base64 PEM>",
  "encryptionAlgorithmIdentifier": 2
}
```

2. Start step-ca with this configuration and confirm the SCEP endpoint is reachable:
```
curl -i "http://localhost/scep/est-core-scep?operation=GetCACaps"
```

3. Submit a SCEP `PKIOperation` request (`PKCSReq`) from a client whose CMS/PKCS#7 implementation runs in a FIPS-approved-only mode (for example, Bouncy Castle FIPS with `org.bouncycastle.fips.approved_only=true`), where the client:
    - Encrypts the request to the configured [decrypterCertificate](https://github.com/smallstep/certificates/blob/master/scep/authority.go#L80C3-L80C23).
    - Verifies the signature on the returned `CertRep` using the CA certificate chain obtained from `GetCACert`.

4. Observe step-ca's log for the request — it shows `status=200` and a successfully issued certificate:
```
method=POST path="/scep/est-core-scep/pkiclient.exe?operation=PKIOperation"
status=200 subject=Aruba-Gateway-01 issuer="EstScepCA Intermediate CA"
```

5. Observe the SCEP client attempting to verify the `CertRep` signature.

### Your Environment

- step-ca version: custom build from `master` (commit-equivalent to the current scep.go implementation using **DecrypterKeyPEM**/**DecrypterKeyURI** for both decryption and signing)

- OS: Ubuntu Linux (host running step-ca as a native binary)

- CA intermediate: RSA 3072-bit

- SCEP decrypter: separate self-signed RSA 3072-bit certificate/key (`CN=est-core-scep-decrypter`), configured via **decrypterCertificate** + `decrypterKeyPEM`

- SCEP client: Java 17 SCEP implementation backed by Bouncy Castle FIPS (`BC-FIPS`), running with `-Dorg.bouncycastle.fips.approved_only=true`

- Client CMS verification library: Bouncy Castle **CMSSignedData / SignerInformation.verify(...)**

### Expected Behavior

[...]

PR #2801, opened 2026-09-17
Fixes #2800 [SCEP] Drop redundant signer creation from SCEP decrypter key

tarokon:
#### Name of feature:
Not a new feature — removes redundant signer creation from the SCEP provisioner.

Previously, `SCEP.Init()` built both a `crypto.Decrypter` and a separate `crypto.Signer` from the configured `DecrypterKeyPEM`/`DecrypterKeyURI `(and set `signerCertificate `to the decrypter certificate). This signer/certificate pair was never actually needed in the normal path, since authority.go's `selectSigner()` already falls back to the CA's own default signer (`a.signerCertificate`, `a.defaultSigner`) whenever a provisioner doesn't supply one via `GetSigner()`. Building an extra signer from the decrypter key on every `Init()` was wasted work (an unnecessary KMS/key operation) and added confusing surface area suggesting SCEP responses were signed with the decrypter key rather than the CA's intermediate.

Reduces unnecessary key-manager calls during provisioner initialization and simplifies the SCEP provisioner's responsibility to just decryption, making the signing path unambiguous (CA's default signer is always used for signing SCEP responses unless a future distinct-signer option is added).

N/A — no user-facing behavior or configuration changes; the signer/signerCertificate/GetSigner() still exist and work as before (returning nil, nil when unset, in which case authority.go falls back to the default signer).

All environments using the SCEP provisioner with decrypterKeyPEM or decrypterKey (URI) configured — no change in supported platforms/workflows.

💔Thank you!


hslatman:
This looks like too big of a hammer to support the FIPS-specific requirement to me. We depend on the current configuration, and it has been successfully used on many platforms so far. Ripping the RSA signer out is thus quite risky without extensive multi-platform testing.

My TODOs that were removed in this PR hint at the proper solution: it needs a separate signer certificate and key.

Going to close this with additional remarks in #2800.

tarokon:
Hi @hslatman ,

Understood, and I agree with your concerns. Let me put together another PR that preserves the current backward-compatible behavior by default, while adding configuration to support separate signing and decryption key pairs.

PR #2803, opened 2026-09-18
Return badCSR when a certificate template rejects an ACME order

Gauss23:
### Problem

A certificate template can reject a request with `fail "..."`, for example
to refuse RSA keys shorter than 3072 bits. For that case the authority
already returns a 400 with the template's message (`authority/tls.go`,
`x509util.TemplateError`). `Order.Finalize` in the ACME layer wraps every
signing error other than webhook errors with `WrapErrorISE`, so the client
only sees:

```
500 urn:ietf:params:acme:error:serverInternal
The server experienced an internal error
```

The message that tells the user what to fix is only in the CA log.

### Change

If `SignWithContext` returns an error that contains an
`*x509util.TemplateError`, `Finalize` returns
`NewDetailedError(ErrorBadCSRType, "%s", msg)`, next to the existing
special case for webhook errors. The client then sees, for example:

```
error finalizing order: The CSR is unacceptable: RSA key too short
```

### RFC

RFC 8555, section 7.4: if the CA is unwilling to issue a certificate for the
submitted CSR, "the problem document returned by the server SHOULD use
error code 'badCSR' and describe specific reasons the CSR was rejected in
its 'detail' field".

### Tests

- `TestOrder_Finalize`: new case `fail/template-error`, where the mock
  signer wraps a `TemplateError` the same way `authority.Sign` does. Without
  the change it fails (`serverInternal`/500 instead of `badCSR`/400).
- `go test ./acme/... ./authority/...` passes with Go 1.26.8 and 1.27.1.


hslatman:
Hey @Gauss23,

It is correct that a template error can be due to data from the CSR. However, many template errors are actually due to a faulty template configured on the CA. In that case the error is thus not due to the ACME client, and returning "bad CSR" would be surprising. We'll have to look into options to discern these cases.

Gauss23:
Hey @hslatman ,

thanks for the comment and you're right. This could be "fixed" with an error message stating that the error is on the CA side and not the client's fault. The user may then be able to contact the administrator of this Step-CA instance and can provide valuable error message instead of a generic one.

